PR #8, opened 2023-03-28
all: Refactoring code to separate resource policy from rest

jukkar:
Refactoring code so that resource policy specific files are placed to pkg/resmgr/ directory. All the generic files are left to pkg/ directory.


jukkar:
> The only one I'm wondering about is `pkg/apis/resmgr`

I just missed that, so this PR was more or less wip. I will move it.

klihub (approved):
LGTM.

PR #17, opened 2023-04-03
cache, topology: fix hint generation bugs.

klihub:
When inserting containers into the cache, make sure host and container paths for mounts are passed in the correct order to the hint generator function.

For each injected device, expect a corresponding cgroup access rule and use it to determine if the device is in read-only mode.

If there is a sysRoot given, use it when looking for a sysfs device based on path, or type/major/minor combo.

### DIFF
--- a/pkg/topology/topology.go
+++ b/pkg/topology/topology.go
@@ -213,15 +213,47 @@ func FindSysFsDevice(dev string) (string, error) {
 		}
 	}
 
-	major := unix.Major(rdev)
-	minor := unix.Minor(rdev)
+	major := int64(unix.Major(rdev))
+	minor := int64(unix.Minor(rdev))
 	if major == 0 {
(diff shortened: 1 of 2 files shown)


pkg/topology/topology.go
```diff
@@ -213,15 +213,47 @@ func FindSysFsDevice(dev string) (string, error) {
 		}
 	}
 
-	major := unix.Major(rdev)
-	minor := unix.Minor(rdev)
+	major := int64(unix.Major(rdev))
+	minor := int64(unix.Minor(rdev))
```
kad, line 217: isn't it uint8 or uint16 in datatypes?

klihub: No. IIRC `uint32` in `x/sys/unix` and `int64` in the OCI Spec.

jukkar (approved):
LGTM

PR #22, opened 2023-04-13
Add initial ghcr support

jukkar:


jukkar:
Closing as the ghcr push already works so this is no longer needed.

Issue #24, opened 2023-04-14
make golangci-lint fails


mythi:
...with a bunch of linter errors (`golangci-lint has version v1.50.1`)

fmuyassarov:
Perhaps it would be also good to add a github action to run golangci-lint as part of presubmit?

klihub:
> Perhaps it would be also good to add a github action to run golangci-lint as part of presubmit?

Yes, once we have fixed this. It is a known issue and the reason why we haven't enabled running golangci-lint as part of our github actions yet.

fmuyassarov:
I have submitted a fix for all the issues currently reported by the linter. Maybe we can enable it once https://github.com/containers/nri-plugins/pull/25 gets merged.

klihub:
All known issues should be fixed in latest `main/HEAD` and in `v0.8.0`, golangci-lint enabled in CI.

PR #20, opened 2023-04-05
fixes: rework cache, remove direct CRI and K8s dependencies.

klihub:
This PR aims to eliminate direct dependency on CRI data types and any packages directly imported from k8s.io/kubernetes. The PR eliminates a number of obsolete leftovers from the cache. It also reworks the implementation of cached pods and containers to wrap corresponding NRI data types. Finally, the PR
gets rid of a few constants and milli-CPUs conversion functions and replaces them with locally defined equivalents to eliminate the need to import anything directly from k8s.io/kubernetes/pkg/*.

- cache: remove obsolete CRI-based insertion/refresh
- cache: remove unused/obsolete container functions
- cache, policies: remove obsolete pod/container CacheID (and update e2e tests accordingly)
- cache: remove annotated PodResourceRequirements (no support for webhook)
- cache: rework Pod to wrap nri.PodSandbox
- cache: rework Container to wrap nri.Container (and update e2e tests accordingly)
- cache: rework cache, pod, container and cache utils tests
- all: switch to using k8s.io/utils/cpuset (instead of kubernetes/pkg/kubelet/cm/cpuset)
- cache: stop importing kubernetes/kubelet packages

The changes in this PR are rather split to more smaller self-contained logical units instead of fewer but larger ones, for easier reviewing.

PR #25, opened 2023-04-17
golangci: fix golangci-lint reported linting issues

fmuyassarov:
This commit fixes go code style according to golangci-lint best practices and makes
`golangci-lint` Makefile target to pass. Fixes: https://github.com/containers/nri-plugins/issues/24


fmuyassarov:
cc @mythi

klihub:
In any case, let's hold this back until #20 is merged. That one is fairly intrusive in terms of which files it needs to touch.

fmuyassarov:
> @fmuyassarov shall we mark this PR as non-draft, get it reviewed and merged?

I even forgot about this patch, lol. Yes, let's do it. I will try to fix other issues as well. I think it will be ready latest by tomorrow.

PR #27, opened 2023-04-24
fixes: update container state after resource allocation in CreateContainer.

klihub:
Update container state after resource allocation. At that point the container has been created from our point of view. Any subsequent changes to (any resources of) the container must use a ContainerUpdate instead of a ContainerAdjustment. Updating the state correctly will force subsequent changes to be collected as updates.


fmuyassarov (approved):
/lgtm

jukkar (approved):

PR #30, opened 2023-04-24
test/statistics-analysis: add test setup for viewing container latencies and NRI plugin resource usage

VillePihlava:
The current setup has been used with containerd. Instructions for using and setting up the test can be found in the README.md file.

### DIFF
--- a/test/statistics-analysis/manifests/sleep-deployment.yaml
+++ b/test/statistics-analysis/manifests/sleep-deployment.yaml
@@ -1,0 +1,30 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: sleep
+  labels:
+    app: sleep
+spec:
+  replicas: ${NUMBER_OF_REPLICAS}
+  selector:
+    matchLabels:
+      app: sleep
+  template:
+    metadata:
+      labels:
+        app: sleep
+    spec:
+      containers:
+      - name: sleep
+        image: quay.io/crio/busybox:latest
+        imagePullPolicy: IfNotPresent
+        command:
+          - sh
+          - -c
+          - sleep 604800
+        resources:
+          requests:
+            cpu: 10m
+            memory: 20M
+          limits:
+            cpu: 10m
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ coverage.txt
 
 *.stamp
 test/e2e/**/output
+
+test/statistics-analysis/output
--- a/test/statistics-analysis/README.md
+++ b/test/statistics-analysis/README.md
@@ -1,0 +1,82 @@
+# Statistics analysis tool
+
+This file contains instructions on using the tool.
+
+## How to use
+
+0. In order to save container runtime logs, add yourself to systemd-journal
+   group and make sure you are in that group when running the script (use "id" command)
+
+1. Install [helm](https://helm.sh/) for using Prometheus chart.
+   * In Fedora: "dnf install helm"
+
+2. Install Python plotting libraries.
+```console
+pip3 install -r requirements.txt
+```
+
+### Running the scripts together
+
+3. Run the script, for example:
+
+```console
+template=~/nri-plugins/build/images/nri-resource-policy-template-deployment.yaml topology_aware=~/nri-plugins/build/images/nri-resource-policy-topology-aware-deployment.yaml balloons=~/nri-plugins/build/images/nri-resource-policy-balloons-deployment.yaml ./scripts/run-tests.sh
+```
+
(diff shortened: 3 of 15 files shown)


VillePihlava:
@jukkar @klihub

benchmarking/.gitignore
```diff
@@ -0,0 +1,2 @@
+/output
+Pipfile.lock
```
jukkar: I think we could add these to main .gitignore file, no need to add a new file here.

benchmarking/get-prometheus-timeseries-data.py
```diff
@@ -0,0 +1,113 @@
...
+        sys.exit(1)
+
+    if args.start is not None and args.end is not None:
+        print(handleQueryOutput(args.url, args.csv, queries, labels, args.start, args.end))
+    else:
+        currentTime = time.time()
+        print(handleQueryOutput(args.url, args.csv, queries, labels, currentTime - args.duration, currentTime))
+
+if __name__ == "__main__":
+    main()
```
jukkar: missing \n at the end of this file

benchmarking/prometheus-values.yaml
```diff
@@ -0,0 +1,17 @@
...
+  enabled: false
+server:
+  global:
+    scrape_interval: 15s
+    evaluation_interval: 15s
+  persistentVolume:
+    enabled: false
+  service:
+    type: NodePort
+    nodePort: 30000
```
jukkar: \n missing

benchmarking/README.md
```diff
@@ -8,6 +8,13 @@ This is work in progress
```
jukkar: In commit "test/statistics-analysis: Fix typo and readme file", the commit message and subject now look a bit weird as it is combined from two different commits.
Could you change this to
```
test/statistics-analysis: Fix topology_aware name in scripts

Fix the typo in run_tests.sh, the topology aware variable
was incorrectly written.
Fix the name of the topology-aware jaeger file, it was
topology-aware but file name is currently topology_aware

Add python dependency list to readme file.
```

benchmarking/manifests/jaeger-deployment.yaml
```diff
@@ -25,6 +25,7 @@ spec:
         - containerPort: 4317
```
jukkar: Please fix the commit message of `test/statistics-analysis: only pull stress-ng if not present.`
it just looks weird as it is a combination of multiple commits but without the commit message fixed.

benchmarking/README.md
```diff
@@ -4,6 +4,9 @@ This is work in progress
```
jukkar: I am not very happy with the squashing of commits, they are no longer containing logical changes but it looks like a too big task to go back. Anyway, minimum you need to do is to sanitize the commit messages of the squashed commits, for example the `test/statistics-analysis: Save NRI resource policy and runtime logs` needs to be fixed.

jukkar (changes requested):
Would it be possible to have the directory change in the first commit instead of last? If too big task, then please split the last commit to two, first one adding the colors etc, and then the last one doing the directory change.

benchmarking/.gitignore
VillePihlava: Done, I also removed the Pipfile.lock because now only a requirements.txt file is used.

benchmarking/get-prometheus-timeseries-data.py
VillePihlava: Fixed

benchmarking/prometheus-values.yaml
VillePihlava: Fixed

benchmarking/README.md
VillePihlava: I changed the commit history by a lot. Now the commits add separate pieces of functionality.

test/statistics-analysis/manifests/sleep-deployment.yaml
```diff
@@ -0,0 +1,30 @@
...
+        command:
+          - sh
+          - -c
+          - sleep 604800
+        resources:
+          requests:
+            cpu: 10m
+            memory: 20M
+          limits:
+            cpu: 10m
```
jukkar, line 30: \n missing

jukkar (approved):
Looks good, minor nit about missing \n in one yaml file

test/statistics-analysis/manifests/sleep-deployment.yaml
VillePihlava: Should now be fixed

klihub (approved):

PR #31, opened 2023-04-26
policy: Cleanup container allocations at startup

jukkar:
When policy is started, it reads the saved container allocations from cache file. This can cause problems if we change the resource allocation policy as the new policy would use the previous resource allocations and not generate allocations based on its own policy. Solve this by forcing re-allocation of resources at startup.

### DIFF
--- a/pkg/resmgr/nri.go
+++ b/pkg/resmgr/nri.go
@@ -115,32 +115,49 @@ func (p *nriPlugin) Configure(cfg, runtime, version string) (stub.EventMask, err
 func (p *nriPlugin) syncWithNRI(pods []*api.PodSandbox, containers []*api.Container) ([]cache.Container, []cache.Container, error) {
 	m := p.resmgr
 
-	m.Info("synchronizing cache state with NRI/CRI runtime...")
+	allocated := []cache.Container{}
+	released := []cache.Container{}
 
-	add, del := []cache.Container{}, []cache.Container{}
+	m.Info("synchronizing cache state with NRI runtime...")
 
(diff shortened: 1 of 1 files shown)


pkg/resmgr/nri.go
```diff
@@ -112,37 +112,6 @@ func (p *nriPlugin) Configure(cfg, runtime, version string) (stub.EventMask, err
...
 }
 
-func (p *nriPlugin) syncWithNRI(pods []*api.PodSandbox, containers []*api.Container) ([]cache.Container, []cache.Container, error) {
-	m := p.resmgr
-
-	m.Info("synchronizing cache state with NRI/CRI runtime...")
-
-	add, del := []cache.Container{}, []cache.Container{}
-
-	_, _, deleted := m.cache.RefreshPods(pods)
```
klihub, line 122: We can't omit calling this as this is the one which makes that new pods are properly injected into the cache and stale ones are removed.

pkg/resmgr/nri.go
```diff
@@ -112,37 +112,6 @@ func (p *nriPlugin) Configure(cfg, runtime, version string) (stub.EventMask, err
...
-	for _, c := range deleted {
-		m.Info("discovered stale container %s...", c.GetID())
-		del = append(del, c)
-	}
-
-	added, deleted := m.cache.RefreshContainers(containers)
```
klihub, line 128: We can't omit calling this as this is the one which makes sure the new containers are properly injected to the cache and the stale ones are removed.

klihub (approved):

PR #37, opened 2023-05-03
benchmarks: check deployment files prior to tests/use.

klihub:
Check the existence of deployment files before running any tests. If a deployment file is simply set to "1" or "yes", try picking up the corresponding generated file from its stock location, assuming we're running out of a cloned repository.

### DIFF
--- a/test/statistics-analysis/scripts/run-tests.sh
+++ b/test/statistics-analysis/scripts/run-tests.sh
@@ -47,6 +47,30 @@ get_pod_name() {
     echo $pod
 }
 
+resolve_deployment() {
+    local name=$1
+    local deployment="${!name}"
+
+    if [ "$name" == "baseline" ]; then
+	return 0
+    fi
+    if [ -z "$deployment" ] || [ -f "$deployment" ]; then
+	return 0
+    fi
+
+    case $deployment in
+	1|yes)
+	    deployment="$BASE_DIR/../../build/images/nri-resource-policy-${name//_/-}-deployment.yaml"
+	    if [ -f "$deployment" ]; then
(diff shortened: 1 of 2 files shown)


jukkar (commented):
Looks good, minor nit about commit msg. The previous commits related to this feature are now named "test/statistics-analysis: ...." perhaps we could follow suit here so the commits are easier to find later (if needed).

test/statistics-analysis/scripts/run-tests.sh
```diff
@@ -47,6 +47,34 @@ get_pod_name() {
...
+    if [ "$name" == "baseline" ]; then
+	return 0
+    fi
+    if [ -z "$deployment" ]; then
+	return 0
+    fi
+
+    if [ -f "$deployment" ]; then
+	return 0
+    fi
```
fmuyassarov, line 63: nit, can we perhaps merge these two operators?
```suggestion
    if  [ -z "$deployment" ] || [ -f "$deployment" ]; then
	return 0
    fi
```

klihub: Updated.

klihub:
> Looks good, minor nit about commit msg. The previous commits related to this feature are now named "test/statistics-analysis: ...." perhaps we could follow suit here so the commits are easier to find later (if needed).

Fair enough. I updated the commit message. Also added another commit which verifies that known dependencies are present  (python3, nc, envsubst) and tests that the plotter can import its python3 dependencies.

fmuyassarov (approved):
LGTM

jukkar (approved):

PR #40, opened 2023-05-15
instrumentation, resmgr: instrument top-level NRI requests with traces.

klihub:
This patch series reworks instrumentation, with the main focus on tracing. In particular it
- reworks tracing support to use opentelemetry instead of opencensus
- simplifies internal instrumentation service
- drops unused obsolete gRPC trace views
- instruments top-level NRI request handlers with traces

The prometheus exporter interface in opentelemetry is not stable yet. Therefore we're stuck with the opencensus implementation for the time being, only reworking it a bit to be structurally more in line with tracing.

Note that the NRI request spans will not be properly nested under the corresponding runtime/CRI request spans. To fix that, we'll need to update the NRI stub interface to pass its context on to the plugin's request handlers and modify ttRPC itself to pass span IDs over the connection.


jukkar (approved):
Looks great :+1:

PR #42, opened 2023-05-16
.github: trigger e2e-test on a self-hosted runner for PRs.

klihub:
For PRs against main run our end-to-end tests on a self- hosted runner.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -4,8 +4,8 @@ name: NRI Resource Policy CI
 
 # Controls when the workflow will run
 on:
-#  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     name: "e2e tests"
     runs-on: >-
       ${{ (
-        github.event_name == 'push' ||
+        github.event_name == 'push' || github.event_name == 'pull_request' ||
```
fmuyassarov, line 20: this seems to be unnecessary since you already have `on.pull_request`
```
on:
  pull_request:
    branches: [ "main" ]
```

PR #43, opened 2023-05-16
e2e: Add developer friendly output to self hosted runner

jukkar:
Prepend each line printed by e2e tests by current time so that it is easier to see how long the various operations take.


klihub (approved):
LGTM.

PR #6, opened 2023-03-23
Resource policy documentation

jukkar:
Documentation from cri-rm updated for nri resource policy.

PR #45, opened 2023-05-17
helm: add helm chart for plugins installations

fmuyassarov:
This commit adds a helm chart to install balloons and topology-aware NRI plugins.

To install the chart into local k8s cluster without current default settings, simply running below is enough:
```sh
# install topology aware plugin
helm install topology-aware deployment/helm/resource-management-policies/topology-aware/

# install balloons aware plugin
helm install balloons deployment/helm/resource-management-policies/balloons
```
For testing and just only generating the final YAML without sending it to k8s
```sh
#  generate  topology aware plugin deployment with  dry run
helm install topology-aware deployment/helm/resource-management-policies/topology-aware/ --dry-run > deployment.yaml

# generate balloons plugin deployment with dry run
helm install balloons deployment/helm/resource-management-policies/balloons --dry-run > deployment.yaml
```

The images currently the Helm chart defaulting to are:
[ghcr.io/containers/nri-plugins/nri-resource-policy-topology-aware:unstable](ghcr.io/containers/nri-plugins/nri-resource-policy-topology-aware:unstable)
[ghcr.io/containers/nri-plugins/nri-resource-policy-balloons:unstable](ghcr.io/containers/nri-plugins/nri-resource-policy-balloons:unstable)

### DIFF
--- a/deployment/helm/resource-management-policies/balloons/Chart.yaml
+++ b/deployment/helm/resource-management-policies/balloons/Chart.yaml
@@ -1,0 +1,11 @@
+apiVersion: v2
+appVersion: main
+description: |
+  The balloons NRI resource policy plugin implements workload placement into
+  “balloons” that are disjoint CPU pools.
+name: nri-resource-policy-balloons
+sources:
+ - https://github.com/containers/nri-plugins
+home: https://github.com/containers/nri-plugins
+type: application
+version: 0.0.0
--- a/deployment/helm/resource-management-policies/balloons/templates/_helpers.tpl
+++ b/deployment/helm/resource-management-policies/balloons/templates/_helpers.tpl
@@ -1,0 +1,8 @@
+{{/*
+Common labels
+*/}}
+{{- define "balloons-plugin.labels" -}}
+app: nri-resource-policy-balloons
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end -}}
--- a/deployment/helm/resource-management-policies/balloons/templates/clusterrole.yaml
+++ b/deployment/helm/resource-management-policies/balloons/templates/clusterrole.yaml
@@ -1,0 +1,24 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nri-resource-policy-balloons
+  labels:
+    {{- include "balloons-plugin.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - watch
+- apiGroups:
+  - topology.node.k8s.io
+  resources:
+  - noderesourcetopologies
+  verbs:
+  - create
+  - get
+  - list
+  - update
+  - delete
--- a/deployment/helm/resource-management-policies/balloons/templates/clusterrolebinding.yaml
+++ b/deployment/helm/resource-management-policies/balloons/templates/clusterrolebinding.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: nri-resource-policy-balloons
+  labels:
+    {{- include "balloons-plugin.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: nri-resource-policy-balloons
+subjects:
+- kind: ServiceAccount
+  name: nri-resource-policy-balloons
+  namespace: {{ .Release.Namespace }}
--- a/deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
+++ b/deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nri-resource-policy-balloons-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "balloons-plugin.labels" . | nindent 4 }}
+data:
+  policy: |+
+    {{- toYaml .Values.config | nindent 4 }}
--- a/deployment/helm/resource-management-policies/balloons/values.yaml
+++ b/deployment/helm/resource-management-policies/balloons/values.yaml
@@ -1,0 +1,19 @@
+# Default values for nri-plugins.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+---
+image:
+  name: ghcr.io/containers/nri-plugins/nri-resource-policy-balloons
+  # tag, if defined will use the given image tag, otherwise Chart.AppVersion will be used
+  tag: unstable
+  pullPolicy: Always
+
+config:
+  ReservedResources:
+    cpu: 750m
+
+hostPort: 8891
+
+resources:
+  cpu: 500m
+  memory: 512Mi
--- a/deployment/helm/resource-management-policies/topology-aware/values.yaml
+++ b/deployment/helm/resource-management-policies/topology-aware/values.yaml
@@ -1,0 +1,19 @@
+# Default values for nri-plugins.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+---
+image:
+  name: ghcr.io/containers/nri-plugins/nri-resource-policy-topology-aware
+  # tag, if defined will use the given image tag, otherwise Chart.AppVersion will be used
+  tag: unstable
+  pullPolicy: Always
+
+config:
+  ReservedResources:
+    cpu: 750m
+
+hostPort: 8891
+
+resources:
+  cpu: 500m
+  memory: 512Mi
--- a/deployment/helm/resource-management-policies/balloons/crds/noderesourcetopology.yaml
+++ b/deployment/helm/resource-management-policies/balloons/crds/noderesourcetopology.yaml
@@ -1,0 +1,270 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870
+    controller-gen.kubebuilder.io/version: v0.11.2
+  creationTimestamp: null
+  name: noderesourcetopologies.topology.node.k8s.io
+spec:
+  group: topology.node.k8s.io
+  names:
+    kind: NodeResourceTopology
+    listKind: NodeResourceTopologyList
+    plural: noderesourcetopologies
+    shortNames:
+    - node-res-topo
+    singular: noderesourcetopology
+  scope: Cluster
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: NodeResourceTopology describes node resources and their topology.
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          topologyPolicies:
+            items:
+              type: string
+            type: array
+          zones:
+            description: ZoneList contains an array of Zone objects.
+            items:
+              description: Zone represents a resource topology zone, e.g. socket,
+                node, die or core.
+              properties:
+                attributes:
+                  description: AttributeList contains an array of AttributeInfo objects.
+                  items:
+                    description: AttributeInfo contains one attribute of a Zone.
+                    properties:
+                      name:
+                        type: string
+                      value:
+                        type: string
+                    required:
+                    - name
+                    - value
+                    type: object
+                  type: array
+                costs:
+                  description: CostList contains an array of CostInfo objects.
+                  items:
+                    description: CostInfo describes the cost (or distance) between
+                      two Zones.
+                    properties:
+                      name:
+                        type: string
+                      value:
+                        format: int64
+                        type: integer
+                    required:
+                    - name
+                    - value
+                    type: object
+                  type: array
+                name:
+                  type: string
+                parent:
+                  type: string
+                resources:
+                  description: ResourceInfoList contains an array of ResourceInfo
+                    objects.
+                  items:
+                    description: ResourceInfo contains information about one resource
+                      type.
+                    properties:
+                      allocatable:
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        description: Allocatable quantity of the resource, corresponding
+                          to allocatable in node status, i.e. total amount of this
+                          resource available to be used by pods.
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
+                      available:
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        description: Available is the amount of this resource currently
+                          available for new (to be scheduled) pods, i.e. Allocatable
+                          minus the resources reserved by currently running pods.
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
+                      capacity:
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        description: Capacity of the resource, corresponding to capacity
+                          in node status, i.e. total amount of this resource that
+                          the node has.
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
+                      name:
+                        description: Name of the resource.
+                        type: string
+                    required:
+                    - allocatable
+                    - available
+                    - capacity
+                    - name
+                    type: object
+                  type: array
+                type:
+                  type: string
+              required:
+              - name
+              - type
+              type: object
+            type: array
+        required:
+        - topologyPolicies
+        - zones
+        type: object
+    served: true
+    storage: false
+  - name: v1alpha2
+    schema:
+      openAPIV3Schema:
+        description: NodeResourceTopology describes node resources and their topology.
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          attributes:
+            description: AttributeList contains an array of AttributeInfo objects.
+            items:
+              description: AttributeInfo contains one attribute of a Zone.
+              properties:
+                name:
+                  type: string
+                value:
+                  type: string
+              required:
+              - name
+              - value
+              type: object
+            type: array
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          topologyPolicies:
+            description: 'DEPRECATED (to be removed in v1beta1): use top level attributes
+              if needed'
+            items:
+              type: string
+            type: array
+          zones:
+            description: ZoneList contains an array of Zone objects.
+            items:
+              description: Zone represents a resource topology zone, e.g. socket,
+                node, die or core.
+              properties:
+                attributes:
+                  description: AttributeList contains an array of AttributeInfo objects.
+                  items:
+                    description: AttributeInfo contains one attribute of a Zone.
+                    properties:
+                      name:
+                        type: string
+                      value:
+                        type: string
+                    required:
+                    - name
+                    - value
+                    type: object
+                  type: array
+                costs:
+                  description: CostList contains an array of CostInfo objects.
+                  items:
+                    description: CostInfo describes the cost (or distance) between
+                      two Zones.
+                    properties:
+                      name:
+                        type: string
+                      value:
+                        format: int64
+                        type: integer
+                    required:
+                    - name
+                    - value
+                    type: object
+                  type: array
+                name:
+                  type: string
+                parent:
+                  type: string
+                resources:
+                  description: ResourceInfoList contains an array of ResourceInfo
+                    objects.
+                  items:
+                    description: ResourceInfo contains information about one resource
+                      type.
+                    properties:
+                      allocatable:
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        description: Allocatable quantity of the resource, corresponding
+                          to allocatable in node status, i.e. total amount of this
+                          resource available to be used by pods.
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
+                      available:
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        description: Available is the amount of this resource currently
+                          available for new (to be scheduled) pods, i.e. Allocatable
+                          minus the resources reserved by currently running pods.
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
+                      capacity:
+                        anyOf:
+                        - type: integer
+                        - type: string
+                        description: Capacity of the resource, corresponding to capacity
+                          in node status, i.e. total amount of this resource that
+                          the node has.
+                        pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                        x-kubernetes-int-or-string: true
+                      name:
+                        description: Name of the resource.
+                        type: string
+                    required:
+                    - allocatable
+                    - available
+                    - capacity
+                    - name
+                    type: object
+                  type: array
+                type:
+                  type: string
+              required:
+              - name
+              - type
+              type: object
+            type: array
+        required:
+        - zones
+        type: object
+    served: true
+    storage: true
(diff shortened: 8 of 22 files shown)


fmuyassarov:
ping @klihub @jukkar 
This is ready but there is only one thing I wanted to clarify and that's why I opened the PR as draft.
The container image is tagged as `devel` in some manifests and the one that is published on the GitHub is `unstable`.
I'm not sure how we want to the images to be tagged during templating.

fmuyassarov:
Another thing, would you like me to add 
1. documentation/instructions how to install the plugins with kustomize
2. documentation/instructions how to install the plugins with helm
3. GitHub action based Helm linter

klihub:
> Another thing, would you like me to add
> 
> 1. documentation/instructions how to install the plugins with kustomize

This would be nice to have as a separate PR.

> 2. documentation/instructions how to install the plugins with helm

This could be part of this PR.

> 3. GitHub action based Helm linter

Can you elaborate what does that mean ?

fmuyassarov:
> > 3. GitHub action based Helm linter
> 
> Can you elaborate what does that mean ?

Helm client has built in linting command to catch possible issues with charts and there is a github action that basically runs the lint command on github pull request. So, I was wondering, if would need something like that. I don't have strong opinion here, because on one hand it is good to lint and catch possible bugs as early as possible (before merging and then realizing it) but on the other hand I don't expect to see many PRs against helm charts in the future. So, it doesn't really make sense to enable the GitHub action for every PR. GitHub action can be limited though to be executed only when charts directory is modified. So, I'm not sure to be honest.

fmuyassarov:
> We need to update the documentation and add information about Helm after the docs pr #6 is merged.

I have half cooked up PR for kustomize way of installing too. Will send once your PR is merged.

deployment/helm/nri-plugins/topology_aware_values.yaml
```diff
@@ -0,0 +1,43 @@
+# Default values for nri-plugins.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+---
+image:
+  name: ghcr.io/containers/nri-plugins/nri-resource-policy-topology-aware
+  # tag, if defined will use the given image tag, otherwise Chart.AppVersion will be used
+  tag: unstable
```
fmuyassarov, line 8: By default the image that Helm chart will be setting on the final deployment is
`ghcr.io/containers/nri-plugins/nri-resource-policy-topology-aware:unstable`

fmuyassarov:
ping @klihub @jukkar @marquiz for review.

deployment/helm/nri-plugins/Chart.yaml
marquiz, line 1: Just a question that should we already now prepare for other plugins as well,  using a subdir like `deployment/helm/resource-policy` or smth?

deployment/helm/nri-plugins/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+appVersion: main
+description: |
+  NRI resource policy is a NRI plugin that will apply hardware-aware resource allocation
+  policies to the containers running in the system.
+name: nri-plugins
```
marquiz, line 6: ´nri-resource-policy` instead? Just thinking that we prolly don't put all future plugins under this chart

deployment/helm/nri-plugins/balloons_values.yaml
```diff
@@ -0,0 +1,43 @@
+# Default values for nri-plugins.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+---
+image:
+  name: ghcr.io/containers/nri-plugins/nri-resource-policy-balloons
+  # tag, if defined will use the given image tag, otherwise Chart.appVersion will be used
+  tag: unstable
+  pullPolicy: IfNotPresent
```
marquiz, line 9: I think this should be always as `unstable` is a "rolling" tag

deployment/helm/nri-plugins/balloons_values.yaml
```diff
@@ -0,0 +1,43 @@
...
+configMap:
+  data:
+    ReservedResources:
+      cpu: 750m
+
+spec:
+  containers:
+    ports:
+    - containerPort: 8891
+      hostPort: 8891
```
marquiz, line 24: I wouldn't "steal" this port from the host. Isn't containerPort enough?

deployment/helm/nri-plugins/templates/clusterrole.yaml
marquiz, line 1: Oh Monday, stupid question: what did we need the access to Nodes again for? @klihub @askervin @jukkar ?

deployment/helm/nri-plugins/templates/daemonset.yaml
```diff
@@ -0,0 +1,66 @@
...
+  name: nri-resource-policy
+  namespace: {{ .Values.namespace }}
+spec:
+  selector:
+    matchLabels:
+    {{- include "nri-plugins.labels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+      {{- include "nri-plugins.labels" . | nindent 7 }}
```
marquiz, line 15: nit:
```suggestion
      {{- include "nri-plugins.labels" . | nindent 8 }}
```

deployment/helm/nri-plugins/templates/noderesourcetopology.yaml
marquiz, line 1: I think CRDs should be put under `deployment/helm/nri-plugins/crds/`. Then Helm does a better job managing them (support e.g. `--skip-crds`)

deployment/helm/nri-plugins/templates/role.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - watch
+- apiGroups:
+  - topology.node.k8s.io
+  resources:
+  - noderesourcetopologies
```
marquiz, line 17: Hmm, wouldn't we actually need ClusterRole for managing cluster-scoped objects?

marquiz (commented):
Thsnk @fmuyassarov. some quick comments below. Are you planning to submit documentation as a separate PR?

fmuyassarov:
> Thsnk @fmuyassarov. some quick comments below. Are you planning to submit documentation as a separate PR?

yes, but can add here as well.

deployment/helm/nri-plugins/templates/noderesourcetopology.yaml
fmuyassarov: sure, makes sense to seperate

marquiz:
> yes, but can add here as well.

With this initial work, I'm fine with whatever you (or others think is preferred)

deployment/helm/nri-plugins/templates/clusterrole.yaml
klihub: Config group labeling ?

deployment/helm/nri-plugins/balloons_values.yaml
fmuyassarov: I'm not sure, but again, this is based on the kustomize https://github.com/containers/nri-plugins/blob/main/deployment/base/nri-resource-policy/daemonset.yaml#L31 and the other deployment methods we already have. To be frank, I'm not sure if it is super necessary to expose the container port to the host. Perhaps it is needed for debugging (or instrumentation) but perhaps it is okay not to expose the port to the host by default and let the users decide if they wish.

WDYT @klihub @jukkar @marquiz @askervin

fmuyassarov: @klihub I just checked removing the `hostPort` in my OpenShift based setup (because I still had it running), so not exposing host port didn't break the way TA plugin assigns resources to containers.

deployment/helm/nri-plugins/templates/role.yaml
fmuyassarov: Fixed in https://github.com/containers/nri-plugins/pull/59

deployment/helm/nri-plugins/templates/clusterrole.yaml
marquiz: Ah yes, that's at least one

marquiz: Consider this as resolved

fmuyassarov:
@marquiz thanks for the reviews. Addressed comments.

fmuyassarov:
> > yes, but can add here as well.
> 
> With this initial work, I'm fine with whatever you (or others think is preferred)

Added in https://github.com/containers/nri-plugins/pull/62

deployment/helm/resource-policy/templates/daemonset.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    {{- include "nri-plugins.labels" . | nindent 4 }}
+  name: nri-resource-policy
+  namespace: {{ .Values.namespace }}
```
marquiz, line 7: Hmm, I think we should respect the helm `--namespace` option

```suggestion
  namespace: {{ .Release.Namespace }}
```

Ditto other templates and drop `namespace` from values.yaml(s)

marquiz: Or, actually we probably wouldn't (or shouldn't) need to supply the `namespace:` field in the chart at all. Helm will install stuff in the correct namespace automatically

marquiz: ...deleted comment about omitting the `namespace:` field, I think it's cleaner to have it there explicitly

deployment/helm/resource-policy/topology_aware_values.yaml
```diff
@@ -0,0 +1,42 @@
...
+  data:
+    ReservedResources:
+      cpu: 750m
+
+spec:
+  containers:
+    ports:
+    - containerPort: 8891
+      protocol: TCP
+      name: metrics
```
marquiz, line 25: Now taking a second look at this, I'd simplify the values a lot and move all the spec.containers stuff just directly to the daemonset template. These nested lists and args etc are very awkward for the user to set from the command line, think like `helm install --set spec.containers.ports[0].containerPort=9876`.

I'd prefere a lot "flatter" structure, e.g. with top-level key like `metricsPort` or `metrics.port` directly.

Keep it simple and add stuff when required.

WDYT?

deployment/helm/resource-policy/topology_aware_values.yaml
```diff
@@ -0,0 +1,42 @@
...
+  pullPolicy: Always
+container:
+  name: nri-resource-policy-topology-aware
+
+namespace: kube-system
+
+configMap:
+  data:
+    ReservedResources:
+      cpu: 750m
```
marquiz, line 18: Can we have simply

```suggestion
config:
  ReservedResources:
    cpu: 750m
```

No need to replicate the podspec/template structure here. WDYT?

deployment/helm/resource-policy/topology_aware_values.yaml
```diff
@@ -0,0 +1,42 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+---
+image:
+  name: ghcr.io/containers/nri-plugins/nri-resource-policy-topology-aware
+  # tag, if defined will use the given image tag, otherwise Chart.AppVersion will be used
+  tag: unstable
+  pullPolicy: Always
+container:
+  name: nri-resource-policy-topology-aware
```
marquiz, line 11: Any reason to have this?  I'd suggest just to drop

```suggestion
```

marquiz:
@fmuyassarov as I commented [here](https://github.com/containers/nri-plugins/pull/62#issuecomment-1589188432) I think we should have separate charts for the different policies and manage those as subcharts. When releasing the chart(s) in a helm repo we cannot rely on people relying on some policy-specific values file on their local machine. We need to have a chart that can be deployed with simple `helm install` and I don't know how to achieve that unless having separate (sub-)charts for each policy. WDYT?

fmuyassarov:
> @fmuyassarov as I commented [here](https://github.com/containers/nri-plugins/pull/62#issuecomment-1589188432) I think we should have separate charts for the different policies and manage those as subcharts. When releasing the chart(s) in a helm repo we cannot rely on people relying on some policy-specific values file on their local machine. We need to have a chart that can be deployed with simple `helm install` and I don't know how to achieve that unless having separate (sub-)charts for each policy. WDYT?

Yes, one way is to separate the plugins as stand alone charts instead of seperate values file.

From the previous [comment](https://github.com/containers/nri-plugins/pull/62#issuecomment-1589188432)
> Hmm, I think we should have a main "values.yaml" for the chart, e.g. how else are we going to release the chart?

I'm not sure I understand this. If a user installs based on the main values.yaml what is that user going to get as a result? I mean what plugin? I think it would be correct to actually have separate charts per plugin as you suggested above, although most of the things (templates) are going to be duplicated. But that will give is stand alone chart which has it is own lifecycle.

marquiz:
> I think it would be correct to actually have separate charts per plugin as you suggested above, although most of the things (templates) are going to be duplicated. But that will give is stand alone chart which has it is own lifecycle.

Exactly (my comment was probably a bit cryptic). What I tried to say is that the chart can basically have only one values.yaml. You cannot choose between two different values.yaml in a chart when doing an install.

With subcharts you shouldn't need to duplicate the templates, right? Have common values there plus all the templates. Then just override required stuff in the subchart. Wouldn't that work and eliminate most of the duplication?

deployment/helm/resource-policy/topology_aware_values.yaml
fmuyassarov: we name containers differently for each plugin. Since we have a single set of templates to be updated by each plugin, we need it here.

marquiz: Why would you need to rename containers differently in different policy deployments. cant you just name it like `policy` or smth?

fmuyassarov: yes, it doesn't make sense indeed. Will remove it.

fmuyassarov:
@marquiz @klihub @jukkar 

Unfortunately, I was unable to succeed or I chose not to proceed with the Library chart, and here's why:

It has already become quite complicated, and I can't imagine how challenging it will be to maintain in the future. For instance, overriding the image name in the DaemonSet of the shared library alone requires at least three steps for a single parameterization.

1. Define a named function which encapsulated the DaemonSet definition

    ```tpl
    {{- define "resource-management-library.configmap.tpl" -}}
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: nri-resource-policy-config.default
      namespace: default
    data: {}
    {{- end -}}
    {{- define "resource-management-library.configmap" -}}
    {{- include "resource-management-library.util.merge" (append . "resource-management-library.configmap.tpl") -}}
    {{- end -}}

    ```

2. Define another named function in another .tpl file to merge image value coming from the Balloons Helm chart into the one written in step (1)
    ```tpl
    {{- /*
    resource-management-library.util.merge will merge two YAML templates and output the result.
    This takes an array of three values:
    - the top context
    - the template name of the overrides (destination)
    - the template name of the base (source)
    */}}
    {{- define "resource-management-library.util.merge" -}}
    {{- $top := first . -}}
    {{- $overrides := fromYaml (include (index . 1) $top) | default (dict ) -}}
    {{- $tpl := fromYaml (include (index . 2) $top) | default (dict ) -}}
    {{- toYaml (merge $overrides $tpl) -}}
    {{- end -}}

    ```

3. Have include entry (or in other words simply call the named function) from within the Helm chart
    ```tpl
    {{- include "resource-management-library.configmap" (list . "balloons.configmap") -}}
    {{- define "balloons.configmap" -}}
    data:
      myvalue: "Hello World"
    {{- end -}}
    ```

As you can see, this involves simply overriding the configMap data portion from the Balloons Helm chart. We would need to do the same for all other fields that we want the user to be able to override, such as ports, protocols, image versions, and so on.

However, what's even more problematic is that I'm currently unsure how users can perform these overrides. In the example above, the Balloons Helm chart is hard-coding the configMap data as "Hello World" within the file, and I don't know how to enable users to easily provide custom data from the terminal during chart installation. Essentially, the values.yaml file becomes useless in this kind of setup. Perhaps there is a way, but I have been unable to find it.

As such, I don't want to complicate it further and instead have separate charts with their own templates, even if they result in duplication. Sorry for switching back and forth between different approaches. It's just that none of them seem to be perfect, or there is not an easy way to override common boilerplate as there is, for example, in Kustomize.

If you're interested in seeing what this means in practice, you can take a look at my branch [helm-monster](https://github.com/fmuyassarov/nri-plugins/tree/helm-monster) branch, where I have kept changes for overriding comnfigMap and also DaemonSet if I recall it right.

fmuyassarov:
and sorry for delaying the release. I just really didn't want to have duplication ....

deployment/helm/resource-policy/topology_aware_values.yaml
fmuyassarov: good point. Fixed

deployment/helm/resource-policy/topology_aware_values.yaml
fmuyassarov: good point. Fixed

klihub (approved):
If you @fmuyassarov and @marquiz consider this good enough to start with, then I'm also fine with it. I don't know enough about helm charts to be able to judge myself.

klihub:
> > If you @fmuyassarov and @marquiz consider this good enough to start with, then I'm also fine with it. I don't know enough about helm charts to be able to judge myself.
> 
> There are a couple of improvements I would like to see in Helm charts. Like those we discussed over the call today. But it doesn't have to be done here but instead as a follow up PR. So, I think this is fine to merge it, but I would like @marquiz to have his words,

@marquiz ?

jukkar (approved):
LGTM

deployment/helm/resource-management-policies/balloons/templates/clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nri-resource-policy
```
marquiz, line 4: Distinct naming. Plus, I think the Helm-specific labels should be applied to every single object

```suggestion
  name: nri-resource-policy-balloons
  labels:
    {{- include "nri-plugins.labels" . | nindent 4 }}
```

... i.e. ditto all other yamls (both the naming and the labels)

deployment/helm/resource-management-policies/topology-aware/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+image:
+  name: ghcr.io/containers/nri-plugins/nri-resource-policy-topology-aware
+  # tag, if defined will use the given image tag, otherwise Chart.AppVersion will be used
+  tag: unstable
+  pullPolicy: Always
+
+namespace: kube-system
+
+config:
+  Active: topology-aware
```
marquiz, line 14: Is this setting really needed? @klihub @jukkar ?

marquiz (commented):
Some comments, mostly about naming

deployment/helm/resource-management-policies/balloons/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+appVersion: main
+description: |
+  The balloons NRI resource policy plugin implements workload placement into
+  “balloons” that are disjoint CPU pools.
+name: nri-resource-policy
```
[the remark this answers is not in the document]
fmuyassarov: ah missed that, thanks for the catch

deployment/helm/resource-management-policies/balloons/templates/clusterrole.yaml
fmuyassarov: done

deployment/helm/resource-management-policies/balloons/templates/_helpers.tpl
```diff
@@ -0,0 +1,8 @@
+{{/*
+Common labels
+*/}}
+{{- define "nri-plugins.labels" -}}
+app: nri-resource-policy
```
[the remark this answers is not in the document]
fmuyassarov: done

deployment/helm/resource-management-policies/topology-aware/values.yaml
fmuyassarov: I have seen us using it in some places and not using in other. Probably can work without that...
But Krisztian and Jukka are the best to confirm.

fmuyassarov:
@marquiz comments are addressed now.

deployment/helm/resource-management-policies/balloons/templates/clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nri-resource-policy-balloons
```
marquiz, line 4: As I commented earlier, helm labels need to be added

deployment/helm/resource-management-policies/balloons/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: nri-resource-policy-balloons
```
marquiz, line 4: Labels are missing

Ditto all all other objects...

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nri-resource-policy-config.default
```
marquiz, line 4: Also here, the name needs to be distinctive

```suggestion
  name: nri-resource-policy-balloons-config.default
```

plus the labels...

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
```diff
@@ -0,0 +1,8 @@
...
+  namespace: {{ .Values.namespace }}
```
marquiz, line 5: See https://github.com/containers/nri-plugins/pull/45#discussion_r1227963129

```suggestion
  namespace: {{ .Release.namespace }}
```

Applies to all templates

deployment/helm/resource-management-policies/topology-aware/values.yaml
marquiz: This just doesn't make any sense and is very confusing to the user as the image should determine what policy to use.

deployment/helm/resource-management-policies/balloons/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+namespace: kube-system
+
+config:
+  Active: balloons
+  ReservedResources:
+    cpu: 750m
+
+metrics:
+  port: 8891
+  protocol: TCP
```
marquiz, line 20: Is this something that is relevant for the user to change? Drop?

deployment/helm/resource-management-policies/topology-aware/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  ReservedResources:
+    cpu: 750m
+
+metrics:
+  port: 8891
+  protocol: TCP
+  hostPort: 8891
+
+quota:
```
marquiz, line 23: Was there discussion about this, why is it called quota? I myself was immediately thinking about ResourceQuota

fmuyassarov: no specific reason. I will rename it

deployment/helm/resource-management-policies/balloons/values.yaml
fmuyassarov: maybe not protocol and container port, but at least host port is relevant in case 8891 port on the host is utilized by something else already and the user wants to expose the metrics on the host

fmuyassarov: I removed those fields and kept only the hostport, because I think that might be desirable by the user to be configured

fmuyassarov:
@marquiz thanks for another round of review. PTAL once more. Thanks

fmuyassarov:
> @marquiz thanks for another round of review. PTAL once more. Thanks

also added docs update accordingly here: https://github.com/containers/nri-plugins/pull/74

jukkar:
FYI, just noticed that upstream documentation generation is broken. I get this error
```
Warning, treated as error:
/docs/docs/resource-policy/installation.md:10:'myst' reference target not found: #Helm
```

fmuyassarov:
> FYI, just noticed that upstream documentation generation is broken. I get this error
> 
> ```
> Warning, treated as error:
> /docs/docs/resource-policy/installation.md:10:'myst' reference target not found: #Helm
> ```
> 
> Also #Manual ref is not found.
> 
> ```
> /docs/docs/resource-policy/installation.md:11:'myst' reference target not found: #Manual
> ```
> 
> and TODO
> 
> ```
> /docs/docs/resource-policy/installation.md:40:'myst' reference target not found: TODO
> ```
> 
> and "terminal"
> 
> ```
> /docs/docs/resource-policy/installation.md:48:Pygments lexer name 'terminal' is not known
> ```
> 
> I suggest using "console" instead which works for me.

thanks for the catch. Let's not do it here, because this PR is not about doc. I see that @marquiz has already submitted a patch in https://github.com/containers/nri-plugins/pull/75

deployment/helm/resource-management-policies/balloons/values.yaml
marquiz: Why would we want hostport? Then that port is reserved and e.g. another plugin cannot be deployed. Containerport should be enough, right?

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nri-resource-policy-balloons-config
```
marquiz, line 4: I think we want the helm labels here too

deployment/helm/resource-management-policies/balloons/values.yaml
marquiz: Or, actually 😏 The HostPort setting could be THE way to achieve pod anti-affinity, if the daemonset controller takes that into account. I.e. one resource policy reserves that port and another one using the same port cannot be scheduled on that node. I don't like it necessarily but it could work

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
fmuyassarov: Fixed, thanks

deployment/helm/resource-management-policies/balloons/values.yaml
fmuyassarov: > Why would we want hostport? Then that port is reserved and e.g. another plugin cannot be deployed. Containerport should be enough, right?

each plugin can be exposed to a different hostPort. Why user might need is for exposing metrics fro example (Prometheus) to the host.

fmuyassarov: > Or, actually smirk The HostPort setting could be THE way to achieve pod anti-affinity, if the daemonset controller takes that into account. I.e. one resource policy reserves that port and another one using the same port cannot be scheduled on that node. I don't like it necessarily but it could work

it might work, but when user has given some custom hostPort, then it will be missed.

marquiz (approved):
Thanks @fmuyassarov for working on this 👍 I didn't spot anything obvious anymore so LGTM from my side.

Has anyone actually verified (i.e. tried in practice that they work) these?

fmuyassarov:
> Thanks @fmuyassarov for working on this +1 I didn't spot anything obvious anymore so LGTM from my side.
> 
> Has anyone actually verified (i.e. tried in practice that they work) these?

I have not installed it but I tested generation of Manifests and since they are the same as we have in kustomize, or in e2e, I assume this should work too as long as other prerequisites are met.

jukkar:
I could try this before we merge this, just a sec...

jukkar:
Couple of notes:

* Could we have pointer to the site where helm can be downloaded/installed. This would help the user to install helm if it is missing from the system
* The daemonset name is now `nri-resource-policy-topology-aware` so the pod name will be like this `nri-resource-policy-topology-aware-ws78f`. The daemonset name was `nri-resource-policy`earlier which created pod name like `nri-resource-policy-ws78f`. This is perhaps not a big issue as we do not use helm to install the plugins in our e2e tests. Those tests expect pod name to be without the plugin type and will fail with this naming schema. As I said, perhaps this is not an issue, but just mentioning it.

jukkar:
Please rebase to main and tweak installation.md to add the helm location (in a separate commit).

fmuyassarov:
> Couple of notes:
> 
>     * Could we have pointer to the site where helm can be downloaded/installed. This would help the user to install helm if it is missing from the system
> 

I have already added the documentation. We don't have a place to where we can upload the charts because we don't even have public images yet. Once we have images available and the initial release, then we can find a place for it in somewhere remotely from where it can be directly installed. 

>     * The daemonset name is now `nri-resource-policy-topology-aware` so the pod name will be like this `nri-resource-policy-topology-aware-ws78f`. The daemonset name was `nri-resource-policy`earlier which created pod name like `nri-resource-policy-ws78f`. This is perhaps not a big issue as we do not use helm to install the plugins in our e2e tests. Those tests expect pod name to be without the plugin type and will fail with this naming schema. As I said, perhaps this is not an issue, but just mentioning it.

Yes, e2e changes will be on a follow up PR.

PR #47, opened 2023-05-17
resmgr, instrumentation: tag instrumentation traces with policy name.

klihub:
Set up an extra `resource-manager.policy = $policy` attribute in our instrumentation telemetry process/Resource.

### DIFF
--- a/pkg/instrumentation/instrumentation.go
+++ b/pkg/instrumentation/instrumentation.go
@@ -28,7 +28,7 @@ import (
 
 const (
 	// ServiceName is our service name in external tracing and metrics services.
-	ServiceName = "NRI-Resource-Plugin"
+	ServiceName = "nri-resource-policy"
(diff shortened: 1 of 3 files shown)


jukkar (commented):
It is not clear from commit message why this is needed, could you elaborate more in commit msg.

klihub:
> It is not clear from commit message why this is needed, could you elaborate more in commit msg.

I update the description in the commit message to
```
    Set up an extra 'resource-manager.policy' attribute for the
    tracing telemetry 'Process' resource. This attribute is part
    of the same collection of data where host name, process ID,
    and our recently added git version and build tags are.
    
    With this addition, our injected tags combined should allow
    one to uniquely identify both the git tree and the policy
    implementation used to record the traces.
```

jukkar (approved):

PR #51, opened 2023-05-24
Makefile: better plugin binary dependencies.

klihub:
Make plugin binaries depend on the go sources of their own main packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ $(BIN_PATH)/%: .static.%.$(STATIC)
 $(BIN_PATH)/nri-resource-policy-topology-aware: \
     $(shell for f in cmd/topology-aware/*.go; do echo $$f; done; \
             for dir in $(shell $(GO_DEPS) ./cmd/topology-aware/... | \
-                          grep '/nri-plugins/' | \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -165,23 +165,23 @@ $(BIN_PATH)/%: .static.%.$(STATIC)
 $(BIN_PATH)/nri-resource-policy-topology-aware: \
     $(shell for f in cmd/topology-aware/*.go; do echo $$f; done; \
             for dir in $(shell $(GO_DEPS) ./cmd/topology-aware/... | \
-                          grep '/nri-plugins/' | \
+                          grep -E '(/nri-plugins/)|(cmd/topology-aware/)' | \
```
jukkar, line 168: Is this really needed? I mean the dependency list already uses cmd/topology-aware in previous line. I checked the output with and without this change and both are identical.

klihub: $(GO_DEPS) does not list 'main', so we need to inject the files belonging to main (each plugin's main) manually to the generate dep list. The first echo should do that. So we generate the list with multiple (two) commands but then grep-filter what we want from them with a single grep. Or at least that was the idea sans any screw-up I potentially did...

jukkar (approved):

PR #54, opened 2023-05-31
cache: Check allowed and denied mounts/devices for hints

jukkar:
Check if user has specified any allowed or denied mount or device paths when generating hints. The code first checks the deny path list and then the allowed one.

The annotations could look like this:
```
  annotations:
    allow.topologyhints.resource-policy.nri.io/pod: |+
      type: prefix
      paths:
        - /foo/bar/whitelisted-path1
        - /xy-zy/another-whitelisted-path1
    allow.topologyhints.resource-policy.nri.io/container.pod0c0: |+
      type: glob
      paths:
        - /whitelisted-path*2
        - /xy-zy/another-whitelisted-path2
    deny.topologyhints.resource-policy.nri.io: |+
      type: prefix
      paths:
        - /foo/bar/blacklisted-path3
    deny.topologyhints.resource-policy.nri.io/pod: |+
      type: glob
      paths:
        - /blacklisted-path*4
    deny.topologyhints.resource-policy.nri.io/container.pod0c1: |+
      type: prefix
      paths:
        - /foo/bar/blacklisted-path5
        - /xy-zy/another-blacklisted-path5
```


jukkar:
If this looks ok, I can start to work on unit tests.

klihub:
This is not something for this PR, but a more generic comment related to our 'effective annotations concept', that came to mind as I was looking at the provided sample annotations:

I think this is one of those cases where letting a container inherit pod-level settings in addition to container-level settings could make sense. That could be done in several ways.
1. As a one-off thing, specific to these annotation, for instance by enclosing `PathList` in another top-level structure and adding there an 'InheritPod bool` field. Then the user can set it to `true` to explicitly indicate inheritability.
2. At the infra-level, by adding something like `GetEffectiveInheritableAnnotations(c cache.Container, name string) []string` (or always returning a `[2]string`), which would dig out both pod and container-specific annotations and return both. In this case, inheritable annotations would be a more generic concept, but inheritance would be an implicit semantic property of each annotation, not something that is explicitly indicated.
3. Change the current `GetEffectiveAnnotations()` to `GetEffectiveAnnotations(c cache.Container, name string) [2]string` which would then always assume inheritability at the cache/container/pod, and it would be left to the caller whether it only uses the last annotation returned (no inheritance), or both if there are two (annotation with inheritance).  From the PodSpec user point of view this is identical to 2. Iheritability is an implicit property of the annotation, and not something which is explicitly indicated.
4. A variation of 2/3 with more user-visibility and control would be to extend the current effective annotation notation to allow indicating wheter pod-level annotation is inherited to all containers. For instance instead of `$KEY/pod`, and `$KEY/container.$CTR`, we could use `$KEY/{default,pod}`, `$KEY/container.$CTR`, and `$KEY/inherited`, with the last indicating an inherited annotation.

pkg/resmgr/cache/container.go
```diff
@@ -31,8 +32,60 @@ import (
...
+)
+
+type PathList struct {
+	Type  MatchType `yaml:"type"`
+	Paths []string  `yaml:"paths"`
+}
+
+func (t *MatchType) UnmarshalJSON(data []byte) error {
+	switch string(data) {
+	case "\"prefix\"":
```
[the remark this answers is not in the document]
jukkar: Added a test case for checking this (missing prefix)

docs/resource-policy/policy/topology-aware.md
```diff
@@ -430,3 +430,46 @@ metadata:
...
+    deny.topologyhints.resource-policy.nri.io/pod: |+
+      type: glob
+      paths:
+        - /blacklisted-path*4
+    deny.topologyhints.resource-policy.nri.io/container.pod0c1: |+
+      type: prefix
+      paths:
+        - /foo/bar/blacklisted-path5
+        - /xy-zy/another-blacklisted-path5
+```
```
klihub, line 475: @jukkar I think we'd need one or more examples about to how to blacklist devices. Maybe one with a particular device and another about blacklisting all block devices. I'm fine with adding those in another PR.

klihub (approved):
LGTM with a small nit about missing device examples.

docs/resource-policy/policy/topology-aware.md
jukkar: I can try to come up with an example here.

PR #55, opened 2023-05-31
e2e: fix test os distro defaulting

fmuyassarov:
Currently `DEFAULT_DISTRO` can not be tuned via environment variable from terminal because no matter what is set from terminal gets overrided by the same variable in run.sh script which gets executed before we reach variable export here.

Example: `DEFAULT_DISTRO=generic/ubuntu2204 ./run_tests.sh `won't create a ubuntu vm but instead will use fedora.

### DIFF
--- a/test/e2e/run_tests.sh
+++ b/test/e2e/run_tests.sh
@@ -3,7 +3,7 @@
 TESTS_DIR="$1"
 RUN_SH="${0%/*}/run.sh"
 
-DEFAULT_DISTRO="generic/fedora37"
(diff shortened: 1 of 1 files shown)


jukkar (approved):

klihub (approved):

jukkar:
After refreshing my memory and looking the code with fresher eyes, the original code was actually correct. The DEFAULT_DISTRO is exactly that, a default one so there is no need to override that in command line.
For selecting the distro, one should use the `distro` env variable, so like this
```
distro=generic/ubuntu2204 ./run_tests.sh
```
It does not harm to have it like what is in the commit so no need to revert the PR.

PR #59, opened 2023-06-12
rbac: control noderesourcetopologies permissions from ClusterRole

fmuyassarov:
NodeReourceTopologies CRD is cluster-scoped resource. As such, RBAC rules related to NodeReourceTopologies should be controlled from ClusterRole object instead of Role object which is more for namespaced objects.


marquiz (approved):
/lgtm

jukkar (approved):

PR #60, opened 2023-06-13
E2E test controller

jukkar:
Simple controller for e2e testing the controller code.
PR contains also hook calls.

Still missing the actual e2e tests, so this is still WIP.


jukkar:
Fixed the post allocation hooks so that PreCreate hook is called properly.
The e2e script was also fixed to catch the PreCreate event.

pkg/resmgr/flags.go
```diff
@@ -72,7 +73,7 @@ func init() {
 		"Minimum interval between two container rebalancing attempts. Use 'disable' for disabling.")
 	flag.StringVar(&opt.StateDir, "state-dir", "/var/lib/nri-resource-policy",
 		"Permanent storage directory path for the resource manager to store its state in.")
-
+	flag.BoolVar(&opt.E2ETestHTTPQuery, "enable-e2e-test-http-query", false, "Allow query via HTTP when doing e2e testing.")
```
klihub, line 76: nitpick: I think this command line option name could be more descriptive. How about something like:
```
    flag.BoolVar(&opt.enableTestAPIs, "enable-test-apis", false, "Allow enabling various test APIs (currently only 'e2e-test' test controller).")
```

klihub (commented):
I think the split to commits is a bit weird now. At least I find it weird that starting up the controller infra is in the e2e test controller creation PR. I think that should go together with the commit that adds callouts to the hooks. So maybe something like this would be a more logical split:
- enable controller (initiate control, call StartStopControllers(), add callout to hooks)
- add e2e test controller
- add test cases

klihub (approved):
LGTM.

PR #62, opened 2023-06-13
docs: add plugin installation document

fmuyassarov:
Add plugin installation document.

### DIFF
--- a/docs/resource-policy/installation.md
+++ b/docs/resource-policy/installation.md
@@ -1,1 +1,140 @@
 # Installation
+
+This repository hosts a collection of plugins of various types, one of which is the resource
+policy plugins. In this example, we will demonstrate the installation process for the topology-aware
+plugin, which falls under the resource policy type. The installation methods outlined
+here can be applied to any other plugin hosted in this repository, regardless of its type.
+
+Currently, there are two installation methods available.
+
+1. [Helm](#Helm)
+2. [Manual](#Manual)
+
+Regardless of the chosen installation method, the NRI plugin installation includes the
+following components: DaemonSet, ConfigMap, CustomResourceDefinition, and RBAC-related objects.
+
+## Prerequisites
+
+- Container runtime:
+    - containerD:
+        - At least [containerd 1.7.0](https://github.com/containerd/containerd/releases/tag/v1.7.0) release version to use the NRI feature
+        - Enable NRI feature by following [these](TODO link) detailed instructions.
+    - CRI-O
+        - At least [v1.26.0](https://github.com/cri-o/cri-o/releases/tag/v1.26.0) release version to use the NRI feature
+        - Enable NRI feature by following [these](TODO link) detailed instructions.
+- Kubernetes 1.24+
+- Helm 3.0.0+
+
+## Installing the Helm Chart
+
+1. Clone the project to your local machine
+    ```sh
+    git clone https://github.com/containers/nri-plugins.git
+    ```
+
+1. Navigate to the project directory
+    ```sh
+    cd nri-plugins
+    ```
+
+1. Install the plugin using Helm. Replace release name with the desired name for your Helm release. In this example, we named it as topology-aware. The default values for topology-aware resource policy plugin are stored in [values.yaml](TODO) file. If you wish to provide custom values to the Helm chart, refer to the [table](#Helm-parameters) below, which describes the available parameters that can be modified before installation.
+
+    ```sh
+    helm install topology-aware deployment/helm/resource-management-policies/topology-aware/
+    ```
+
+1. Verify the status of the daemonset to ensure that the plugin is running successfully
+
+    ```terminal
+    kubectl get daemonset -n kube-system nri-resource-policy
+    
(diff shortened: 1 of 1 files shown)


fmuyassarov:
cc @marquiz

fmuyassarov:
I would like to hear your opinion about two things: 
1. I added default helm chart values for the TA plugin, shall I add it for the balloons too? But then, what if there is another plugin added in the future, do we keep adding for every plugin?
2. Do we need to give some examples which shows Helm installation with custom values?

marquiz:
> * I added default helm chart values for the TA plugin, shall I add it for the balloons too? But then, what if there is another plugin added in the future, do we keep adding for every plugin?

Hmm, I think we should have a main "values.yaml" for the chart, e.g. how else are we going to release the chart? 🧐 Maybe the two policies should actually be managed with subcharts. Anyway, I think it would be good to specify defaults for both (or all) policies where they differ. Common defaults could be a separate table. Thoughts?

> * Do we need to give some examples which shows Helm installation with custom values?

We could have a simple example involving `helm --set` for example.

fmuyassarov:
> Hmm, I think we should have a main "values.yaml" for the chart, e.g. how else are we going to release the chart? monocle_face Maybe the two policies should actually be managed with subcharts. Anyway, I think it would be good to specify defaults for both (or all) policies where they differ. Common defaults could be a separate table. Thoughts?
> 
Just replied on the Helm addition PR
> > * Do we need to give some examples which shows Helm installation with custom values?
> 
> We could have a simple example involving `helm --set` for example.

+1

Missed your replies, for some reason for this project GitHub sends notifications to my intel email address.

docs/resource-policy/installation.md
```diff
@@ -1 +1,264 @@
...
+    - Enable NRI feature by following [these](TODO link) detailed instructions.
+- CRI-O
+    - At least [v1.26.0](https://github.com/cri-o/cri-o/releases/tag/v1.26.0) release version to use the NRI feature
+    - Enable NRI feature by following [these](TODO link) detailed instructions.
+
+## Helm
+
+1. Clone the project to your local machine
+    ```sh
+    git clone https://github.com/containers/nri-plugins.git
```
jukkar, line 32: Can the helm chart installation done without this step i.e., can we point directly to a file in github?

docs/resource-policy/installation.md
```diff
@@ -1 +1,264 @@
...
+
+That's it! You have now installed the topology-aware NRI resource policy plugin using Helm.
+
+#### Helm default values
+
+This table provides an overview of the default values for the Topology-aware plugin Helm chart.
+
+<table>
+<tr>
+<td>
```
jukkar, line 63: Instead of embedding html code to this file, could you use the markdown table formatting here?

docs/resource-policy/installation.md
marquiz: I'm afraid it cannot be done without a packed helm chart. I don't think we want to include packed chart (.tgz) in the main branch so this will become available with the first release when the helm repo within gh-pages gets populated.

What we could do is to extend the update-gh-pages script to always pack the chart from the tip of the main branch addi it to the gh-pages branch (and update the helm repo index). But I think that's subject of a separate PR...

docs/resource-policy/installation.md
fmuyassarov: Fixed

docs/resource-policy/installation.md
fmuyassarov: Yes, I think we need at least one release and then update the doc again.

fmuyassarov:
cc @klihub @marquiz @jukkar 

The documentation PR is also updated according to the latest changes in the Helm PR.

docs/resource-policy/installation.md
```diff
@@ -1 +1,140 @@
...
+| `quota.memory`     | 512Mi                                                                                                                                      | Memory qouta for the Pod                             |
+| `metrics.port`     | 8891                                                                                                                                       | metrics port on the container                        |
+| `metrics.hostPort` | 8891                                                                                                                                       | metrics port to expose on the host                   |
+| `config`           | <pre><code>Active: topology-aware</code><br><code>ReservedResources:</code><br><code>  cpu: 750m</code></pre>                                                                                                                                                                                                                                                           | plugin configuration data                            |
+
+## Manual installation
```
jukkar, line 100: Do we need to have instructions using git clone + make + kubectl apply, or is it too low level and for developers only?

jukkar (approved):
LGTM

docs/resource-policy/installation.md
fmuyassarov: or do you mean to say have a Makefile target to run kubectl command?

jukkar: No, I did not mean that. I was just wondering that if user does not have helm or kustomize, what he/she could do then. But that is probably something only for developers and we do not need to think about that here.

fmuyassarov: if no helm or kustommize, they can just simply apply the deployment with kubectl

PR #64, opened 2023-06-14
Implement handling of update container requests.

klihub:
This patch set implements handling requests to update container resources. This can happen if the cluster has the InPlacePodVerticalScaling feature gate set to true.

The patch series 
- updates our NRI dependencies to the latest HEAD
- updates our stub plugin interfaces with the necessary related changes (context passed from stub to plugin)
- adds functions for storing and retrieving container resource updates in the cache
- hooks the policy into NRI request processing for UpdateContainers
- implements reallocation of resources for the topology-aware policy


jukkar (approved):
LGTM

PR #71, opened 2023-06-20
Memtierd NRI plugin

luukasmakila:
Memtierd NRI plugin code and docs.

Summary of changes:
- Added the cmd/memtierd directory
- Added main.go containing the code for the plugin
- Added README.md and creating-your-own-memtierd-configurations.md to the docs/memtierd directory
- Added a Dockerfile to building your own version of the plugin
- Added cmd/memtierd/templates containing an example config and deployment that show how the plugin can be used
- In the main.go only the StartContainer(), StopContainer(), getFullCgroupPath(), editMemtierdConfig() and startMemtierd() functions have been modified and the rest are from the NRI plugin template.

Open questions:
- Can the go decencies be handled better? I see the other plugins don't have go.mod and go.sum with their main.go code.
- Should the memtierd docs be moved to some other place? Right now I just added a memtierd directory under the docs directory.


jukkar:
> * Should the memtierd docs be moved to some other place? Right now I just added a memtierd directory under the docs directory.

I think this is fine, the resource policy plugin doc files are in `docs/resource-policy` so it makes sense to place memtierd plugin documentation to `docs/memtierd`

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+		fields := strings.Fields(line)
+		if fields[0] == "cgroup2" {
+			cgroupMountPoint = fields[1]
+			break
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		log.Fatalf("failed to read /proc/mounts: %v", err)
+	}
+	findCmd := exec.Command("find", cgroupMountPoint, "-type", "d", "-wholename", fullPath)
```
jukkar, line 277: I think we should only use golang provided functionality here and not fire external program for this. Could we for example use `filepath.WalkDir` here?

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/containerd/nri/pkg/api"
+	"github.com/containerd/nri/pkg/stub"
+)
+
+type config struct {
+	CfgParam1 string `json:"cfgParam1"`
+}
```
klihub, line 41: This can/should be removed.

jukkar (commented):
Looks reasonable. Please fix the commit msg, the subject could be something like (use your own words if this does not look good)
 `memtierd: Add NRI plugin for memory tiering`

And please do not leave commit body empty. Put here for example description what this new plugin does etc. so that someone looking this in the future can figure out what this commit is about (without looking the actual diff).

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+	if config == "" {
+		return 0, nil
+	}
+
+	err := yaml.Unmarshal([]byte(config), &cfg)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse configuration: %w", err)
+	}
+
+	log.Info("Got configuration data %+v...", cfg)
```
klihub, line 80: This should probably also be removed, if the plugin does not expect to get pre-launched by the runtime (by getting symlinked to /opt/nri/plugins/$idx-$plugin) and possibly get configuration data (stored in /etc/nri/conf.d).

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+func (p *plugin) StopPodSandbox(pod *api.PodSandbox) error {
+	log.Infof("Stopped pod %s/%s...", pod.GetNamespace(), pod.GetName())
+	return nil
+}
+
+func (p *plugin) RemovePodSandbox(pod *api.PodSandbox) error {
+	log.Infof("Removed pod %s/%s...", pod.GetNamespace(), pod.GetName())
+	return nil
+}
```
klihub, line 107: These can all be removed if the plugin does not need to process those events.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+	adjustment := &api.ContainerAdjustment{}
+	updates := []*api.ContainerUpdate{}
+	return adjustment, updates, nil
+}
+
+func (p *plugin) PostCreateContainer(pod *api.PodSandbox, ctr *api.Container) error {
+	log.Infof("Created container %s/%s/%s...", pod.GetNamespace(), pod.GetName(), ctr.GetName())
+	return nil
+}
```
klihub, line 132: These can be removed if the plugin does not process those events.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+	updates := []*api.ContainerUpdate{}
+
+	return updates, nil
+}
+
+func (p *plugin) PostUpdateContainer(pod *api.PodSandbox, ctr *api.Container) error {
+	log.Infof("Updated container %s/%s/%s...", pod.GetNamespace(), pod.GetName(), ctr.GetName())
+	return nil
+}
```
klihub, line 201: These can be removed if the plugin does not process those events.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+		fmt.Println(err)
+	}
+
+	return []*api.ContainerUpdate{}, nil
+}
+
+func (p *plugin) RemoveContainer(pod *api.PodSandbox, ctr *api.Container) error {
+	log.Infof("Removed container %s/%s/%s...", pod.GetNamespace(), pod.GetName(), ctr.GetName())
+	return nil
+}
```
klihub, line 239: This can be removed if the plugin does not process this event.

cmd/memtierd/main.go
klihub, line 277: I think this is probably not a good idea...

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+	outputFile, err := os.OpenFile("/host"+outputFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		fmt.Printf("Failed to open output file: %v\n", err)
+	}
+
+	// Create the command and write its output to the output file
+	cmd := exec.Command("memtierd", "-c", "", "-config", configFilePath)
+	cmd.Stdout = outputFile
+	cmd.Stderr = outputFile
```
klihub, line 351: Maybe it's okay to start like this. But in the long run wouldn't it be better to refactor memtierd so that it offers an easily consumable interface for importing and running it as part of other binaries (typically on a dedicated goroutine) ?

We did something similar a while ago just to eliminate most of the duplicated boilerplate code in our various plugins which set up resource management. I think it was commit b62f866e04f0d7ef81e8f2ad9485f6ce740a83cb.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,397 @@
...
+	log.Infof("Stopped pod %s/%s...", pod.GetNamespace(), pod.GetName())
+	return nil
+}
+
+func (p *plugin) RemovePodSandbox(pod *api.PodSandbox) error {
+	log.Infof("Removed pod %s/%s...", pod.GetNamespace(), pod.GetName())
+	return nil
+}
+
+func (p *plugin) CreateContainer(pod *api.PodSandbox, ctr *api.Container) (*api.ContainerAdjustment, []*api.ContainerUpdate, error) {
```
klihub, line 109: You will need to add a `context.Context` first argument here and all the remaining other NRI event handlers (StartContainer, StopContainer).

It is necessary since commit 5e0c7288c9d2543f2348e270a83b12c96c5b3111 which was merged recently. Otherwise the stub will refuse to instantiate itself for your plugin.

cmd/memtierd/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN GO111MODULE=on go get -u github.com/containerd/nri/pkg/stub
+RUN GO111MODULE=on go get -u github.com/sirupsen/logrus
+RUN GO111MODULE=on go get -u gopkg.in/yaml.v2
+
+# Build the plugin binary
+RUN go build -gcflags "all=-N -l"
+
+# Make the plugin directory & add the plugin there
+RUN sudo mkdir -p /opt/nri/plugins/
+RUN mv memtierd-nri /opt/nri/plugins/10-memtier-nri
```
klihub, line 26: This only makes sense on the host side to make the plugin get automatically started by the runtime (whenever the runtime starts with NRI enabled). Otherwise in an image it is better to just put the plugin in /bin.

docs/memtierd/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Prerequisities
+- NRI enabled on your container runtime
+
+## Making your own image
+
+The Memtierd binary needs to built first. To do so, follow the instructions found [here](https://github.com/intel/memtierd/tree/main/cmd/memtierd).
+
+```console
+# Move the memtierd binary to the memtierd directory, so that it will be involved in the image
```
klihub, line 13: typo: s/involved/included

docs/memtierd/README.md
```diff
@@ -0,0 +1,77 @@
+# Memtierd NRI plugin
+
+An easy way to start using [Memtierd](https://github.com/intel/memtierd/tree/main) as a memory manager in your Kubernetes cluster.
+
+## Prerequisities
+- NRI enabled on your container runtime
+
+## Making your own image
+
+The Memtierd binary needs to built first. To do so, follow the instructions found [here](https://github.com/intel/memtierd/tree/main/cmd/memtierd).
```
klihub, line 10: I think you should do a `go install github.com/intel/memtierd/cmd/memtierd@latest` from the dockerfile, then copy the binary from the golang installed path to /bin... if memtierd can be compiled with a simple 'go build/go install'. If not, then cloning the repo in the Dockerfile then building with make should work.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,331 @@
...
+	log.Infof("Stopped container %s/%s/%s...", pod.GetNamespace(), pod.GetName(), ctr.GetName())
+
+	//
+	// This is the container (post-)stop request handler. You can update any
+	// of the remaining running containers. Take a look at the functions in
+	// pkg/api/update.go to see the available controls.
+	//
+
+	podName := pod.GetName()
+	dirPath := fmt.Sprintf("/tmp/memtierd/%s", podName)
```
jukkar, line 126: This certainly works, but could be replaced with
`dirPath := fmt.Sprintf("%s/memtierd/%s", os.TempDir(), podName)`
 to get the tmp directory as then user is able to override it if needed.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,331 @@
...
+
+func editMemtierdConfig(fullCgroupPath string, podName string, containerName string, template string) (string, string, string) {
+	templatePath := fmt.Sprintf("/templates/%s", template)
+	yamlFile, err := ioutil.ReadFile(templatePath)
+	if err != nil {
+		log.Fatalf("Error reading YAML file: %v\n", err)
+	}
+
+	// Create pod directory if it doesn't exist
+	podDirectory := fmt.Sprintf("/tmp/memtierd/%s", podName)
```
jukkar, line 230: And here
`podDirectory := fmt.Sprintf("%s/memtierd/%s", os.TempDir(), podName)`

cmd/memtierd/main.go
```diff
@@ -0,0 +1,331 @@
...
+	if err := os.MkdirAll("/host"+podDirectory, 0755); err != nil {
```
jukkar, line 231: The `/host` part must not be hard coded but should be given as a parameter to this program.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,331 @@
...
+
+	outputFilePath := fmt.Sprintf("%s/memtierd.%s.output", podDirectory, containerName)
+
+	var memtierdConfig MemtierdConfig
+	err = yaml.Unmarshal(yamlFile, &memtierdConfig)
+	if err != nil {
+		log.Fatalf("Error unmarshaling YAML: %v\n", err)
+	}
+
+	//fullCgroupPathString := string(fullCgroupPath)
```
jukkar, line 243: Please remove commented out code, they just confuse future reader

cmd/memtierd/main.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	memtierdConfig.Policy.Config = policyConfigFieldString
+
+	out, err := yaml.Marshal(&memtierdConfig)
+	if err != nil {
+		log.Fatalf("Error marshaling YAML: %v\n", err)
+	}
+
+	configFilePath := fmt.Sprintf("/host"+podDirectory+"/%s.yaml", containerName)
```
jukkar, line 264: Also `/host` here needs to be configurable.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	log.Infof("YAML file successfully modified.")
+
+	return podDirectory, outputFilePath, configFilePath
+}
+
+func startMemtierd(podName string, containerName string, podDirectory string, outputFilePath string, configFilePath string) {
+	log.Infof("Starting Memtierd")
+
+	outputFile, err := os.OpenFile("/host"+outputFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
```
jukkar, line 277: and here too

jukkar:
Please do not put merge commits in the PR. Not sure what happened with the /host handling as I am not seeing it in the diff. My earlier comment about it was that we should have a command line option that allows user to set it so similar way as how resource plugins do it. You need to also add the relevant host command line option to deployment file in this case.
So the idea is that user can set the host mount path and then we use that in the plugin (if it is set).

luukasmakila:
> Please do not put merge commits in the PR. Not sure what happened with the /host handling as I am not seeing it in the diff. My earlier comment about it was that we should have a command line option that allows user to set it so similar way as how resource plugins do it. You need to also add the relevant host command line option to deployment file in this case.
> So the idea is that user can set the host mount path and then we use that in the plugin (if it is set).

For some reason its giving me a conflict on the docs/index.md but im not sure why that is happening I just added a new line to the file.

About the host path. In the templates/pod-memtierd.yaml we are actually mapping that path into the container so I just removed the host path stuff it still works. I had not realised that earlier.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func (p *plugin) StopContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) ([]*api.ContainerUpdate, error) {
+	log.Infof("Stopped container %s/%s/%s...", pod.GetNamespace(), pod.GetName(), ctr.GetName())
+
+	//
+	// This is the container (post-)stop request handler. You can update any
+	// of the remaining running containers. Take a look at the functions in
+	// pkg/api/update.go to see the available controls.
+	//
```
klihub, line 131: I would remove this comment (which originates from the NRI template plugin), or replace it with a concise summary of what this plugin does when a container is stopped.

cmd/memtierd/main.go
```diff
@@ -0,0 +1,339 @@
...
+	os.Exit(0)
+}
+
+func getFullCgroupPath(ctr *api.Container) string {
+	cgroupPath := ctr.Linux.CgroupsPath
+
+	split := strings.Split(cgroupPath, ":")
+
+	partOne := split[0]
+	partTwo := fmt.Sprintf("%s-%s.scope", split[1], split[2])
```
klihub, line 168: What happens is kubelet/the runtime is configured to run with the cgroupfs cgroup driver instead of systemd ? Is that caught and reported here or somewhere else ?

marquiz:
Superseded by #117

PR #75, opened 2023-06-21
Fix building of docs

marquiz:
Fix problems in the installation doc.

Also add CI steps to verify that we don't break docs in the future


fmuyassarov (approved):

jukkar (approved):
LGTM

PR #76, opened 2023-06-21
docs: fix references to github repo

marquiz:
This relies on the myst parser's (new in v0.19) feature of customizing
external URL schemes with the "myst_url_schemes" configuration option.
We specify new URL schemes "blob" and "tree" which are supposed to be
used for references to blobs (files) and trees (directories) in the git
repo.

This also drops sphinx event hooks ("doctree-resolved" and
"missing-reference"), simplifying the conf.py considerably. They were
useless anyway as they were totally broken and didn't actually fix any
references to repo files.

The downside of this is that the affected references are not shown when
browsing .md files directly from github (as github dosn't handle our
custom 'tree:' and 'blob:' url schemes. However, it should be preferable
to have the cross-references working in the official documentation.

This PR also updates myst-parser to the latest version and pins the versions of
sphinx-rtd-theme and sphinx-markdown-tables for predictable results.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,16 +32,6 @@ master_doc = 'docs/index'
 #
 # This section determines the behavior of links to local items in .md files.
 #
-#  if useGitHubURL == True:
-#
-#     links to local files and directories will be turned into github URLs
-#     using either the baseBranch defined here or using the commit SHA.
-#
-#  if useGitHubURL == False:
-#
-#     local files will be moved to the website directory structure when built
-#     local directories will still be links to github URLs
-#
 #  if built with GitHub workflows:
 #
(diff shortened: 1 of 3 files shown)


klihub:
> The downside of this is that the affected references are not shown when browsing .md files directly from github (as github dosn't handle our custom 'tree:' and 'blob:' url schemes. However, it should be preferable to have the cross-references working in the official documentation.

The best would be to get both working: keep the links working in the .md files and get them working in the generated documentation as well which I believe will require some python post-genertion fixup logic in the end. Why didn't the fixup in the event hooks work ?

jukkar (approved):

klihub (approved):

PR #77, opened 2023-06-21
github/workflows: fix broken yaml

marquiz:
Fix yaml syntax that got broken in
19bbdd5438fd30bf09c7e33930a5877b9f4b4418.

In addition there's another (required) patch to simplify the workflow file.

### DIFF
--- a/.github/workflows/project-checks.yaml
+++ b/.github/workflows/project-checks.yaml
@@ -38,18 +38,18 @@ jobs:
           make verify
         working-directory: src/github.com/containerd/nri-plugin-collection
 
-    - name: Build docs
-      run: |
-        make site-build
-        make clean-html
+      - name: Build docs
+        run: |
+          make site-build
+          make clean-html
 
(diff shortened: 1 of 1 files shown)


klihub:
@marquiz I mean I screwed up my earlier review of your PR where I was wondering about the same but I was hasty and approved it. Then it went in, although I think it should never have. We have another workflow already in place which is dealing exclusively with publishing documentation. It is this [publish-docs.yaml]( https://github.com/containers/nri-plugins/blob/main/.github/workflows/publish-docs.yml). I think it would be better to stick with that (it is basically the same as you added) and make fixes there.

marquiz:
> Isn't this conflicting with publish-docs.yaml ?

No, that one is about actually publishing the docs. This one just verifies that the docs are buildable and that the gh-pages update script doesn't fail.

marquiz:
ping @klihub @jukkar the CI is broken until we merge this one

klihub (approved):

PR #78, opened 2023-06-22
docs: flatten the hierarchy

marquiz:
Drop one level of sub-hierarchy, making "Policies" a top-level item in the "Resource Policy Plugins" section. Makes the documentation easier to follow (and easier to find the detailed docs of the actual policies/plugins).

### DIFF
--- a/docs/resource-policy/configuration.rst
+++ b/docs/resource-policy/configuration.rst
@@ -1,7 +1,0 @@
-Configuration
-#############
-.. toctree::
-   :maxdepth: 1
-
-   policy/index.rst
-   node-agent.md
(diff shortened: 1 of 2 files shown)


klihub (approved):
I'm not sure which one is better, so I'm fine with this.

marquiz:
I just find the current very deep hierarchy really odd and cumbersome

Resource Policy Plugins -> Configuration -> Policy -> Topology-Aware

Especially when it's not just about configuration of the policy but a detailed documentation of all aspects of the policy

klihub:
Merged as part of stacked PR #80.

PR #86, opened 2023-06-26
instrumentation: use a more appropriate propagator.

klihub:
Use a correct propagator for ttRPC (and gRPC) span propagation. Without this, ttRPC/gRPC trace spans information is not be properly propagated over the IPC.


jukkar (commented):
I have some difficulties understand what this does and why we need it.  Could this be opened a bit more in the commit message?

jukkar (approved):

PR #91, opened 2023-06-30
config: Fix configuration files for sample-configmap

changzhi1990:
It tries to read the "nri-resource-policy.cfg" file from the configmap. But the configmap misses the "nri-resource-policy.cfg" key.

Otherwise, the pod can't start with the "no such file or directory" error.

The detailed error message is:
```
F0629 07:03:01.197013       1 log.go:486] failed to create resource manager: resource-manager: failed to load fallback configuration /etc/nri-resource-policy/nri-resource-policy.cfg: config error: failed to apply configuration from file: config error: failed to read file "/etc/nri-resource-policy/nri-resource-policy.cfg": open /etc/nri-resource-policy/nri-resource-policy.cfg: no such file or directory
```

### DIFF
--- a/deployment/overlays/balloons/sample-configmap.yaml
+++ b/deployment/overlays/balloons/sample-configmap.yaml
@@ -3,38 +3,53 @@ kind: ConfigMap
 metadata:
   name: nri-resource-policy-config.default
 data:
-  policy: |+
-    ReservedResources:
-      cpu: 750m
-  #  balloons:
-  #    PinCPU: true
-  #    PinMemory: true
-  #    IdleCPUClass: lowpower
-  #    BalloonTypes:
-  #      - Name: "quad"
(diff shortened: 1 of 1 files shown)


klihub:
@fmuyassarov Can you take a look at the overlay/kustomize/etc. setup we have ?

The quoted error message from the logs comes from the fallback config not being found. That config us normally a ConfigMap, but one which is mounted in the filesystem as a file at a fallback config location which is then passed to the plugin on the command line. Based on a quick glance at the commit in this PR, it does not look like doing the right thing. Could you try digging into what is the root cause of fallback config file not being found error ?

fmuyassarov:
Hi @changzhi1990 can you give a bit more info how are you deploying the plugins into the k8s cluster and preferably share the deployment you are using ?

PR #96, opened 2023-07-18
Drop unmantained deployment files

marquiz:
Drop the redundant and unmaintained *yaml.in deployment templates and the related sed wizardry from the makefile. We now use kustomize, instead.


fmuyassarov:
In general I'm fine but I recall that we had a discussion about the removal with @jukkar  and @klihub , but I don't remember all the details. If I'm not mistaken, they wanted to keep it so that developers could easily deploy and test their changes.

marquiz:
Yeah, let's wait for them , this is not urgent. IMHO it would be just best to drop this undocumented/unmaintained deployment method. To me it doesn't even look handy or useful either. In quick one-node testing the developer could just as easily (or easier) run the nri-plugin directly from command line. In multi-node scenario the user in practice needs to make the images available in some registry. We should make developers to use kustomize or helm so these would get some test coverage from there as well

Makefile
```diff
@@ -297,27 +297,7 @@ verify-fmt:
...
-	sed -e "s|IMAGE_PLACEHOLDER|$${NRI_IMAGE_REPOTAG}|g" \
-            -e 's|^\(\s*\)tolerations:$$|\1tolerations:\n\1  - {"key": "cmk", "operator": "Equal", "value": "true", "effect": "NoSchedule"}|g' \
-            -e 's/imagePullPolicy: Always/imagePullPolicy: Never/g' \
-            < "`pwd`/cmd/$${img}/$${tag}-deployment.yaml.in" \
-            > "$(IMAGE_PATH)/$${tag}-deployment.yaml"; \
-	sed -e "s|IMAGE_PLACEHOLDER|$${NRI_IMAGE_REPOTAG}|g" \
-            -e 's|^\(\s*\)tolerations:$$|\1tolerations:\n\1  - {"key": "cmk", "operator": "Equal", "value": "true", "effect": "NoSchedule"}|g' \
-            -e 's/imagePullPolicy: Always/imagePullPolicy: Never/g' \
-            < "`pwd`/test/e2e/files/$${tag}-deployment.yaml.in" \
-            > "$(IMAGE_PATH)/$${tag}-deployment-e2e.yaml"
```
jukkar, line 320: e2e tests do not work if we remove these lines, we could remove the above sed that copies the deployment file from cmd directory

jukkar: Alternatively, we could remove all of these but then have the e2e deployment file set somewhere in the ansible playbook. Perhaps it is easier for this PR just to change the Makefile so that the e2e script still work and do the e2e changes later.

klihub:
Closing this for now. Let's revisit this once we've have the new CR-based configuration in place, and potentially also some of the reworked e2e test bits from CRI Resource Manager lifter over.

PR #97, opened 2023-07-19
Patch containerd to enable NRI before running NRI plugin

fmuyassarov:
This commit adds an init container to enable NRI in containerd configuration file. Users installing plugin via Helm have an option to opt in for using that init container to enable the NRI before running plugin. By default, this option is disabled.

### DIFF
--- a/deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
+++ b/deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
@@ -1,0 +1,31 @@
+{{- if .Values.nri.enable }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: patch-containerd-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "balloons-plugin.labels" . | nindent 4 }}
+data:
+  patch_nri.sh: |
+   #!/bin/bash
+
+   config_file="/home/config.toml" 
+   cp /etc/containerd/config.toml "$config_file"
+
+   # Check if the disable field is set to false under [plugins."io.containerd.nri.v1.nri"]
+   status=$(awk -F ' = ' '/^\[plugins\."io\.containerd\.nri\.v1\.nri"\]/{flag=1} flag && /disable/{print $2; exit}' "$config_file")
+   if [ "$status" = "false" ]; then
+      echo "NRI is already enabled..."
+   elif [ "$status" = "true" ]; then
+      # Enable the NRI plugin in containerd
+      sed '/\[plugins.\"io.containerd.nri.v1.nri/,/^$/s/disable = true/disable = false/' -i "$config_file"
+      cp -f "$config_file" /etc/containerd/config.toml
+      echo "Restarting containerd...."
+      nsenter -t 1 -m systemctl restart containerd
+      echo "NRI has now been enabled..."
+   else
+      echo "NRI plugin entry doesn't exist in the continerd config file."
+      echo "Doing nothing..."
+   fi
+{{- end }}
--- a/deployment/helm/resource-management-policies/topology-aware/templates/configmap_emable_nri.yaml
+++ b/deployment/helm/resource-management-policies/topology-aware/templates/configmap_emable_nri.yaml
@@ -1,0 +1,31 @@
+{{- if .Values.nri.enable }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: patch-containerd-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "topology-aware-plugin.labels" . | nindent 4 }}
+data:
+  patch_nri.sh: |
+   #!/bin/bash
+
+   config_file="/home/config.toml" 
+   cp /etc/containerd/config.toml "$config_file"
+
+   # Check if the disable field is set to false under [plugins."io.containerd.nri.v1.nri"]
+   status=$(awk -F ' = ' '/^\[plugins\."io\.containerd\.nri\.v1\.nri"\]/{flag=1} flag && /disable/{print $2; exit}' "$config_file")
+   if [ "$status" = "false" ]; then
+      echo "NRI is already enabled..."
+   elif [ "$status" = "true" ]; then
+      # Enable the NRI plugin in containerd
+      sed '/\[plugins.\"io.containerd.nri.v1.nri/,/^$/s/disable = true/disable = false/' -i "$config_file"
+      cp -f "$config_file" /etc/containerd/config.toml
+      echo "Restarting containerd...."
+      nsenter -t 1 -m systemctl restart containerd
+      echo "NRI has now been enabled..."
+   else
+      echo "NRI plugin entry doesn't exist in the continerd config file."
+      echo "Doing nothing..."
+   fi
+{{- end }}
--- a/deployment/helm/resource-management-policies/balloons/templates/daemonset.yaml
+++ b/deployment/helm/resource-management-policies/balloons/templates/daemonset.yaml
@@ -17,6 +17,20 @@ spec:
       serviceAccount: nri-resource-policy-balloons
       nodeSelector:
         kubernetes.io/os: "linux"
+      {{- if .Values.nri.enable }}
+      hostPID: true
+      initContainers:
+      - name: patch-containerd
+        image: ubuntu:22.04
+        command: ["/tmp/patch_nri.sh"]
+        volumeMounts:
+        - name: patch-containerd-config
+          mountPath: /tmp
+        - name: containerd-config
+          mountPath: /etc/containerd/config.toml
+        securityContext:
+          privileged: true
+      {{- end }}
       containers:
         - name: nri-resource-policy-balloons
           args:
@@ -78,3 +92,13 @@ spec:
         hostPath:
           path: /var/run/nri
           type: Directory
+      {{- if .Values.nri.enable }}
+      - name: patch-containerd-config
+        configMap:
+          name: patch-containerd-config
+          defaultMode: 0777
+      - name: containerd-config
+        hostPath:
+          path: /etc/containerd/config.toml
+          type: File
+      {{- end }}
--- a/deployment/helm/resource-management-policies/balloons/values.yaml
+++ b/deployment/helm/resource-management-policies/balloons/values.yaml
@@ -17,3 +17,6 @@ hostPort: 8891
 resources:
   cpu: 500m
   memory: 512Mi
+
+nri:
+  enable: false
--- a/deployment/helm/resource-management-policies/topology-aware/templates/daemonset.yaml
+++ b/deployment/helm/resource-management-policies/topology-aware/templates/daemonset.yaml
@@ -17,6 +17,20 @@ spec:
       serviceAccount: nri-resource-policy-topology-aware
       nodeSelector:
         kubernetes.io/os: "linux"
+      {{- if .Values.nri.enable }}
+      hostPID: true
+      initContainers:
+      - name: patch-containerd
+        image: ubuntu:22.04
+        command: ["/tmp/patch_nri.sh"]
+        volumeMounts:
+        - name: patch-containerd-config
+          mountPath: /tmp
+        - name: containerd-config
+          mountPath: /etc/containerd/config.toml
+        securityContext:
+          privileged: true
+      {{- end }}
       containers:
         - name: nri-resource-policy-topology-aware
           args:
@@ -78,3 +92,13 @@ spec:
         hostPath:
           path: /var/run/nri
           type: Directory
+      {{- if .Values.nri.enable }}
+      - name: patch-containerd-config
+        configMap:
+          name: patch-containerd-config
+          defaultMode: 0777
+      - name: containerd-config
+        hostPath:
+          path: /etc/containerd/config.toml
+          type: File
+      {{- end }}
--- a/deployment/helm/resource-management-policies/topology-aware/values.yaml
+++ b/deployment/helm/resource-management-policies/topology-aware/values.yaml
@@ -17,3 +17,6 @@ hostPort: 8891
 resources:
   cpu: 500m
   memory: 512Mi
+
+nri:
+  enable: false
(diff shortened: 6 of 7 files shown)


deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
```diff
@@ -0,0 +1,36 @@
...
+     # Check if the nri plugin is already enabled
+     if grep -q "disable = false" "$config_file"; then
+       echo "NRI is already enabled..."
+     else
+       # Enable the NRI plugin
+       sed -i "s/disable = true/$switch/" "$config_file"
+       echo "NRI has been enabled..."
+       cp -f "$config_file" /etc/containerd/config.toml
+       echo "Restarting containerd...."
+       nsenter -t 1 -m bash
```
kad, line 30: do you really need bash here or just execute systemctl in ns?

deployment/helm/resource-management-policies/topology-aware/templates/configmap_emable_nri.yaml
```diff
@@ -0,0 +1,35 @@
...
+     # Check if the nri plugin is already enabled
+     if grep -q "disable = false" "$config_file"; then
+       echo "NRI plugin is already enabled..."
+     else
+       # Enable the NRI plugin
+       sed -i "s/disable = true/$switch/" "$config_file"
+       echo "NRI plugin has been enabled..."
+       cp -f "$config_file" /etc/containerd/config.toml
+       echo "Restarting containerd...."
+       systemctl restart containerd
```
kad, line 30: you have difference in scripts here and in balloons directory. is that intentional?

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
fmuyassarov: only systemctl isn't enough, I also need something to read the file and modify it is necessary

fmuyassarov:
@kad image is now pinned to [ubuntu:22.04](https://hub.docker.com/layers/library/ubuntu/22.04/images/sha256-965fbcae990b0467ed5657caceaec165018ef44a4d2d46c7cdea80a9dff0d1ea?context=explore).

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
```diff
@@ -0,0 +1,36 @@
...
+     # Check if the nri is already enabled
+     if grep -q "disable = false" "$config_file"; then
+       echo "NRI is already enabled..."
+     else
+       # Enable the NRI plugin
+       sed -i "s/disable = true/$switch/" "$config_file"
+       cp -f "$config_file" /etc/containerd/config.toml
+       echo "Restarting containerd...."
+       nsenter -t 1 -m bash
+       systemctl restart containerd
```
klihub, line 30: Does this really work as intended ?

I wonder about that `nsenter` usage. AFAICT, here it starts `bash` in the (mount) namespace of init/pid 1. Then once that bash exits, the original patch_nri.sh script continues in its original namespaces and runs `systemctl restart containerd`. So what does that `nsenter` accomplish ?

fmuyassarov: > Does this really work as intended ?
> 
Yes
> I wonder about that `nsenter` usage. AFAICT, here it starts `bash` in the (mount) namespace of init/pid 1. Then once that bash exits, the original patch_nri.sh script continues in its original namespaces and runs `systemctl restart containerd`. So what does that `nsenter` accomplish ?

We don't exit the bash of the pid1 until we run `systemctl restart containerd`. We enter the pid1 in `nsenter -t 1 -m bash` and we don't exit but instead we just kill the init container as a whole. So, `nsenter` is entering the mount namespace of the init process (PID 1) and launching a bash within that namespace.  For having the access to the host pid, I added  `hostPID: true.`

klihub: I think that nsenter-started bash exits immediately (when running without a controlling tty) and `systemctl restart containerd` is run in the same namespace as the main script.

Anyway, I wonder what `nsenter -t 1 -m` would anyway change from the namespaces point of view. Namespaces are inherited across forks. If a process is not actively changing its namespaces, it should be running in the same namespace as its parent process, so transitively as PID 1/its init process.

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
```diff
@@ -0,0 +1,36 @@
...
+   config_file="/home/config.toml"
+   plugin="[plugins.\"io.containerd.nri.v1.nri\"]"
+   switch="disable = false"
+   
+   cp /etc/containerd/config.toml "$config_file"
+
+   # Check if the nri entry exists in the config file
+   if grep -q "$plugin" "$config_file"; then
+     # Check if the nri is already enabled
+     if grep -q "disable = false" "$config_file"; then
```
marquiz, line 22: This looks really haphazard - some other plugin config might have a similarly nameed (`disable`) setting. Remembet that it could be a 3rd party plugin outside containerd's codebase. I'd strongly suggest to use sed here and only edit lines between ` [plugins."io.containerd.nri.v1.nri"]` and the next config setction (line starting with a bracket `[`)

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if .Values.nri.enable }}
+apiVersion: v1
+kind: ConfigMap
```
marquiz, line 3: Do we really need this configmap business, the script is not that super complex. I think it would be simpler and safer (no way to affect what get's executed in privileged mode in host pid namespace by just mangling a configmap).

I'd suggest to write the script in the init container spec, smth like
```yaml
    command:
      - "/bin/bassh"
      - "-c"
    args:
      - |
        echo Patching containerd config
        sed '/\[plugins.\"io.containerd.nri.v1.nri/,/^$/s/disable = false/disable = true/' -i /etc/containerd/config.toml
        systemctl restart containerd
```

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
marquiz: Yeah, the nsenter bash looks like nop to me, too

fmuyassarov: It seems that relying on `nsenter -t 1 -m` wasn't sufficient to execute the systemctl... command from within a script, even though it functions well when manually executed in the current terminal session. I have now used `nsenter -t 1 -m systemctl restart containerd` instead, thanks to the help of @klihub .

Here is the short POC:

[POC](https://asciinema.org/a/sTsXlRdPA8mgkyxB9KNVzyggt)

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
fmuyassarov: Thanks, replaced by sed now.

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
fmuyassarov: I'm personally not fan of adding multi line commands into pod spec especially when the command is long like with `sed` here. It makes it hard to read the DaemonSet file.

marquiz: I fail to see why the configmap is any more readable. What about the security concerns?

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
```diff
@@ -0,0 +1,31 @@
...
+   elif [ "$status" = "true" ]; then
+      # Enable the NRI plugin in containerd
+      sed '/\[plugins.\"io.containerd.nri.v1.nri/,/^$/s/disable = true/disable = false/' -i "$config_file"
+      cp -f "$config_file" /etc/containerd/config.toml
+      echo "Restarting containerd...."
+      nsenter -t 1 -m systemctl restart containerd
+      echo "NRI has now been enabled..."
+   else
+      echo "NRI plugin entry doesn't exist in the continerd config file."
+      echo "Doing nothing..."
```
marquiz, line 29: Shouldn't we add the nri plugin config in this case? The containerd config might be a stub, only containing settings that differ from the defaults.

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
```diff
@@ -0,0 +1,31 @@
...
+kind: ConfigMap
+metadata:
+  name: patch-containerd-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "balloons-plugin.labels" . | nindent 4 }}
+data:
+  patch_nri.sh: |
+   #!/bin/bash
+
```
marquiz, line 12: I think we should bail out on any error
```suggestion
set -e
```

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
```diff
@@ -0,0 +1,31 @@
...
+   config_file="/home/config.toml" 
+   cp /etc/containerd/config.toml "$config_file"
+
+   # Check if the disable field is set to false under [plugins."io.containerd.nri.v1.nri"]
+   status=$(awk -F ' = ' '/^\[plugins\."io\.containerd\.nri\.v1\.nri"\]/{flag=1} flag && /disable/{print $2; exit}' "$config_file")
```
marquiz, line 17: Wouldn't it be simpler to just grep if `[plugins."io.containerd.nri.v1.nri"]` exists, do sed unconditioinally and use e.g. diff to see if anything was changed? WDYT?

Plus, in either of these cases there's still the corner case that the`[plugins."io.containerd.nri.v1.nri"]` section is present but the the `disable` setting hasn't been specified explicitly. Do we need to cover this(?) 🧐

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
marquiz: I mean you should be able to put the script in the pod spec formatted exactly like in the configmap

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
fmuyassarov: I was not sure. If `[plugins."io.containerd.nri.v1.nri"]` is not there, we can of course add the following defaults.
WDYT @klihub ?

```toml
    # Enable NRI support in containerd.
    disable = false
    # Allow connections from externally launched NRI plugins.
    disable_connections = false
    # plugin_config_path is the directory to search for plugin-specific configuration.
    plugin_config_path = "/etc/nri/conf.d"
    # plugin_path is the directory to search for plugins to launch on startup.
    plugin_path = "/opt/nri/plugins"
    # plugin_registration_timeout is the timeout for a plugin to register after connection.
    plugin_registration_timeout = "5s"
    # plugin_requst_timeout is the timeout for a plugin to handle an event/request.
    plugin_request_timeout = "2s"
    # socket_path is the path of the NRI socket to create for plugins to connect to.
    socket_path = "/var/run/nri/nri.sock"
```

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
fmuyassarov: > Wouldn't it be simpler to just grep if `[plugins."io.containerd.nri.v1.nri"]` exists, do sed unconditioinally and use e.g. diff to see if anything was changed? WDYT?
> 
It can be simpler, but it also means extra/unnecessary restart of the containerd.

> Plus, in either of these cases there's still the corner case that the`[plugins."io.containerd.nri.v1.nri"]` section is present but the the `disable` setting hasn't been specified explicitly. Do we need to cover this(?) monocle_face

Is that a valid config? I mean, does containerd accept a config file without disable field set to either true/false?

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
fmuyassarov: > I fail to see why the configmap is any more readable. What about the security concerns?

what kind of security concerns do you mean?

fmuyassarov: > I mean you should be able to put the script in the pod spec formatted exactly like in the configmap

Yes I know. I just didn't like having multi line, long command in the spec of the pod. 

I tend to isolate things into their own entity, instead of having one big YAML. 
But anyways, I don't mean to say it is incorrect, it is just my personal preference and I'm totally fine to do it in the spec.

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
marquiz: > So, diff would be used I guess to check if restart is needed or not right? Would not it be better instead of having diff just to check if we actually need to touch the file or not? Actually for me these two options (1. what I have, 2. what you suggest) don't really make a much difference. One is doing diff before touching the file the other is after touching the file.

Sure, if the diff is empty don't update the file. I was thinking about (in pseudo'ish bash cod)

```
sed ... /etc/containerd/config.toml > config.tmp
if ! diff -q /etc/containerd/config.toml config.tmp; then 
  mv config.tmp /etc/containerd/config.toml
  systemctl restart containerd
fi
```

> > Plus, in either of these cases there's still the corner case that the`[plugins."io.containerd.nri.v1.nri"]` section is present but the the `disable` setting hasn't been specified explicitly. Do we need to cover this(?) monocle_face
> 
> Is that a valid config? I mean, does containerd accept a config file without disable field set to either true/false?

Yeah, it is. The config may be an empty file or just one line if at the extrema.

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
marquiz: OR just put `disable = false` as that's the only one needed, and not pretend that we necessarily know all the nri plugin config options 

Of course this only makes sense if the containerd version that is running supports nri...

deployment/helm/resource-management-policies/balloons/templates/configmap_enable_nri.yaml
marquiz: > what kind of security concerns do you mean?

The visibility and assurance about what acutally gets run in the privileged host-pid-ns container. This whole business of mounting "some" configmap into the container and running a shell script from there

fmuyassarov:
closing this in favor of https://github.com/containers/nri-plugins/pull/107.

Issue #99, opened 2023-07-26
Issues with balloons helm installation


askervin:
Make sure that helm installation of the balloons policy works, applies proper CPU frequencies (use kHz instead of MHz in sample CPU frequencies), and implement a e2e test for it.

changzhi1990:
Use this command to install the helm charts

```
helm install balloon .  --namespace kube-system
```

changzhi1990:
I think we can add some documents to clarify this.

askervin:
Closing as helm is used now as a default method for deploying policies in e2e tests and documentation has been updated. Feel free to reopen or file new issue if there are any problems with helm installation.

PR #100, opened 2023-07-27
Fix balloons helm installation and add doc

changzhi1990:
1. Add more document about balloons helm installation guide.
2. Remove the ".default" suffix from the daemonset and configmap yaml files.

### DIFF
--- a/deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
+++ b/deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
@@ -1,10 +1,15 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: nri-resource-policy-config.default
+  name: nri-resource-policy-config
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "balloons-plugin.labels" . | nindent 4 }}
 data:
-  policy: |+
-    {{- toYaml .Values.config | nindent 4 }}
+  nri-resource-policy.cfg: |+
+    policy:
+      {{- toYaml .Values.config | nindent 6 }}
+    instrumentation:
+      {{- toYaml .Values.instrumentation | nindent 6 }}
+    cpu:
+      {{- toYaml .Values.cpu | nindent 6 }}
--- a/docs/resource-policy/installation.md
+++ b/docs/resource-policy/installation.md
@@ -139,6 +139,14 @@ For the manual installation we will be using templating tool to generate Kuberne
 
 That's it! You have now installed the topology-aware NRI resource policy plugin using kutomize.
 
+Also, you can use the following **helm** command to install the NRI resource policy plugin in the **nri-plugins/deployment/helm/resource-management-policies/balloons** directory.
+
+```
+helm install balloons . --namespace kube-system
+```
+
+*Note: please specify the kube-system namespace.*
+
 ## Manual uninstallation
 
 To uninstall plugin manifests you can run the following command:
@@ -149,3 +157,5 @@ kustomize build deployment/overlays/topology-aware/ | kubectl delete -f -
 
 Note: this removes DaemonSet, ConfigMap, CustomResourceDefinition, and RBAC-related objects associated
 with the chart.
+
+Also, you can use the **helm uninstall balloons --namespace kube-system** to uninstall the NRI resource policy plugin.
--- a/deployment/helm/resource-management-policies/balloons/templates/daemonset.yaml
+++ b/deployment/helm/resource-management-policies/balloons/templates/daemonset.yaml
@@ -73,7 +73,7 @@ spec:
           path: /var/run/nri-resource-policy
       - name: resource-policyconfig
         configMap:
-          name: nri-resource-policy-config.default
+          name: nri-resource-policy-config
       - name: nrisockets
         hostPath:
           path: /var/run/nri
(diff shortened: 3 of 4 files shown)


changzhi1990:
Related https://github.com/containers/nri-plugins/issues/99

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
```diff
@@ -1,10 +1,15 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: nri-resource-policy-config.default
+  name: nri-resource-policy-config
```
fmuyassarov, line 4: Why is this being removed? 
Watcher is expecting to have .default, see here https://github.com/containers/nri-plugins/blob/66846cb397314376bc8269ce0549387c89bcc595/pkg/resmgr/agent/watcher.go#L206

docs/resource-policy/installation.md
```diff
@@ -139,6 +139,14 @@ For the manual installation we will be using templating tool to generate Kuberne
...
 That's it! You have now installed the topology-aware NRI resource policy plugin using kutomize.
 
+Also, you can use the following **helm** command to install the NRI resource policy plugin in the **nri-plugins/deployment/helm/resource-management-policies/balloons** directory.
+
+```
+helm install balloons . --namespace kube-system
+```
+
+*Note: please specify the kube-system namespace.*
+
```
fmuyassarov, line 149: I would instead put the `cd/pushd` into the directory as another step instead of having two helm install commands which are doing the same thing but from different directory.

docs/resource-policy/installation.md
```diff
@@ -149,3 +157,5 @@ kustomize build deployment/overlays/topology-aware/ | kubectl delete -f -
 
 Note: this removes DaemonSet, ConfigMap, CustomResourceDefinition, and RBAC-related objects associated
 with the chart.
+
+Also, you can use the **helm uninstall balloons --namespace kube-system** to uninstall the NRI resource policy plugin.
```
fmuyassarov, line 161: same comment here.

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
changzhi1990: Hi, I encounter this error in the nri pod if I don't remove the .default suffix. 
```
E0727 13:32:22.818135       1 resource-manager.go:478] configuration rejected: config error: module main: no child corresponding to data nri-resource-policy.cfg
```

I don't why the error can be gone if I remove the ".default" suffix.

docs/resource-policy/installation.md
changzhi1990: I think that we should provide the helm install and uninstall commands because people are familiar with them.

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
fmuyassarov: Can you give a bit more context about how are you installing, which manifests you are using?

docs/resource-policy/installation.md
changzhi1990: Thanks for your link. I notice that there is a helm install command for the topology-aware policy. Why the balloon policy doesn't have any helm install command like topology-aware policy?

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
changzhi1990: image
Here is the detailed error message in the nri pod.

Also, I use the latest nri code and the helm command is 

```
sdp@b49691a74bec:~/zhi/own/nri-plugins/deployment/helm/resource-management-policies/balloons/balloons$ helm install balloon . --namespace kube-system
```

docs/resource-policy/installation.md
fmuyassarov: Well the only thing that changes between plugins when installing via Helm is the directory and it is mentioned at the top of the document that we are showcasing TA as an example which can be applied to any other plugin hosted on this repo.

I don't think that it is a good idea to repeat the same commands for each plugin, especially when the number of the plugins
in the future. We expect that every plugin Helm charts follow the same structure in terms of file structure, thus ensure that same installation command can be applied to all plugins with minor changes of the directory and the chart name potentially.

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
fmuyassarov: It is strange...but I will see if I can reproduce it and dig a bit.

docs/resource-policy/installation.md
fmuyassarov: But perhaps we should make it a bit more clear that the same commands can be re-used for other plugins installations with minor changes?

deployment/helm/resource-management-policies/balloons/templates/configmap.yaml
marquiz: I guess the problem is that we're using the same configmap in two conflicting ways:
1. mount the configmap inside the container to use for --fallback-config
2. use the same config map for dynamic configuration, as a K8s object directly watched by the daemon (and not mounted inside the container)

The only quick fix I can think of would be to have two separate configmaps, one for the fallback config and one for the dynamic configuration. Going forward we should try to get rid of the fallback config

PR #102, opened 2023-08-07
github: Disable e2e-test workflow

jukkar:
Currently there is no self hosted runner that could run the e2e-test workflow so disable it for now.


klihub (approved):
LGTM.

PR #106, opened 2023-08-22
sample-configs: drop rdt from the sample configuration

marquiz:
RDT configuration is not supported.


fmuyassarov:
Should we also clean up other RDT related pieces in the code as well on the same PR?

marquiz:
> Should we also clean up other RDT related pieces in the code as well on the same PR?

No, I don't think so. We can still set the RDT class. We just don't support configuring the resctrl fs

klihub (approved):

PR #107, opened 2023-08-24
Add utility to patch containerd config for enabling nri

fmuyassarov:
This commit adds an init container to enable NRI in containerd configuration file. Users installing plugin via Helm have an option to opt in for using that init container to enable the NRI before running plugin. By default, this option is disabled.


fmuyassarov:
ping @kad @klihub @marquiz

deployment/helm/resource-management-policies/balloons/templates/daemonset.yaml
```diff
@@ -17,6 +17,17 @@ spec:
       serviceAccount: nri-resource-policy-balloons
       nodeSelector:
         kubernetes.io/os: "linux"
+      {{- if .Values.nri.enable }}
```
klihub, line 20: nit: Wouldn't calling this something like `[.Values.]patchContainerdConfig` be more apt ? The extra action behind this switch is containerd-specific.

pkg/utils/config/main.go
```diff
@@ -0,0 +1,128 @@
...
+func unmarshalTOML(data []byte) (map[string]interface{}, error) {
+	// Unmarshal TOML content into a map
+	var tomlMap map[string]interface{}
+	if err := tomlv2.Unmarshal(data, &tomlMap); err != nil {
+		return nil, err
+	}
+	return tomlMap, nil
+}
+
+// go-tomml v2 doesn't automatically indents content of tables by default.
```
klihub, line 74: typo: s/indents/indent/

pkg/utils/config/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := writeToFile(tomlFilePath, buf.String()); err != nil {
+		log.Fatalf("Error writing to file: %v", err)
+	}
+
+	if err := runNsenter(); err != nil {
+		log.Fatalf("Failed to restard the containerd: %v", err)
+	}
+}
+
+func runNsenter() error {
```
klihub, line 48: nit: why not to call this `restartContainerd()` ?

pkg/utils/config/main.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	defer f.Close()
+
+	w := bufio.NewWriter(f)
+	_, err = w.WriteString(content)
+	if err != nil {
+		return err
+	}
+	return w.Flush()
+
```
klihub, line 99: nit: extra empty line

pkg/utils/config/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := runNsenter(); err != nil {
+		log.Fatalf("Failed to restard the containerd: %v", err)
+	}
+}
+
+func runNsenter() error {
+	pid := "1"
+	command := "nsenter"
+	args := []string{"-t", pid, "-m", "systemctl", "restart", "containerd"}
```
klihub, line 51: nit: why the separate indirection via a variable for pid 1 ?

pkg/utils/config/go.mod
```diff
@@ -0,0 +1,5 @@
+module nri/nri-plugins/pkg/utils/config
+
+go 1.19
+
+require github.com/pelletier/go-toml/v2 v2.0.9
```
klihub, line 5: nit: containerd itself is using go-toml v1 (v1.9.5). Wouldn't it be safer to use the same version.

deployment/helm/resource-management-policies/balloons/templates/daemonset.yaml
fmuyassarov: done

pkg/utils/config/main.go
fmuyassarov: done

pkg/utils/config/go.mod
fmuyassarov: I'm not sure. I'm not sure if using v2 can cause us some issues when modifying the config. At least, when I tried to modify it with v2 and pass that config to containerd it was fine. And I hope containerd is would also switch to v2.

fmuyassarov:
@klihub PTAL, addressed your comments. Thanks.

pkg/utils/config/main.go
```diff
@@ -0,0 +1,129 @@
...
+	if err := restartContainerd(); err != nil {
+		log.Fatalf("Failed to restard the containerd: %v", err)
+	}
+}
+
+// restartContainerd uses nsenter to access the host's mount and PID namespaces.
+// This enables us to restart the host's containerd from within the init container
+// after making necessary configuration modifications.
+func restartContainerd() error {
+	command := "nsenter"
```
klihub, line 52: I still feel that we'd need a bigger/fatter comment/warning here/somewhere to clarify that this binary must be run in the host PID namespace (and consequently with `hostPID: true` when running as a Pod/daemonset) for any of this to work.

kad: we can integrate check that /proc/1/cmdline is present and have expected value (e.g. contains "systemd")

pkg/utils/config/go.mod
```diff
@@ -0,0 +1,5 @@
+module nri/nri-plugins/pkg/utils/config
+
+go 1.19
```
kad, line 3: small nit: with go 1.19 already out of support, should we require minimum 1.20?

fmuyassarov: I used 1.19 because the entire project is on 1.19. If you don't mind, I will switch to 1.20, which is my own preference too.

pkg/utils/config/main.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+		} else {
+			plugins[nriPluginKey] = map[string]interface{}{
+				disableKey: false,
+			}
+			log.Println("NRI plugin section not found, adding it to enable NRI...")
+		}
+	} else {
+		fmt.Println("Top level plugins section not found, adding it to enable NRI...")
+		config = make(map[string]interface{})
```
klihub, line 121: Isn't this resetting the configuration completely instead of just adding a "plugins" section with the NRI plugin enabled ?

pkg/utils/config/main.go
fmuyassarov: I improved the comment here on the func() and also added a note in the documentation for the helm chart field.

pkg/utils/config/main.go
```diff
@@ -0,0 +1,131 @@
...
+	tomlMap = updateNRIplugin(tomlMap)
+
+	buf, err := addIndentation(tomlMap)
+	if err != nil {
+		log.Fatalf("Error adding indentation: %v", err)
+	}
+
+	if err := writeToFile(tomlFilePath, buf.String()); err != nil {
+		log.Fatalf("Error writing to file: %v", err)
+	}
```
klihub, line 41: Maybe it's just me but I'd
- combine reading and decoding to a single `readConfig(file string) (map[string]interface{}, error)` function (probably using just toml.NewDecoder().Decode()`
- have an `enableNRIPlugin(map [string]interface{}) error function (your `updateNRIplugin()` with fixes)
- combine encoding with indentation and file writing into a single `writeConfig(file string, config map[string]interface{}) error` which would just do a `toml.NewEncoder(f).SetIndentTables(true).Encode(config)`

pkg/utils/config/go.mod
klihub: Let's do that in a separate PR.

pkg/utils/config/main.go
fmuyassarov: I don't think so. Let me check and confirm.

fmuyassarov: I tried with some data unrelated to plugins as in the example and in that case yes. You are right. I will fix it.
```
[debug]
  address = "/run/containerd/debug.sock"
  uid = 0
  gid = 0
  level = "info"

[metrics]
  address = ""
  grpc_histogram = false

[cgroup]
  path = ""

```

pkg/utils/config/go.mod
klihub: Okay, fair enough.

pkg/utils/config/main.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if err := restartContainerd(); err != nil {
+		log.Fatalf("Failed to restard the containerd: %v", err)
```
klihub, line 44: typo: s/restard/restart/

pkg/utils/config/main.go
klihub: And this was merely just me thinking out loud. These changes are not required for merging this PR.

fmuyassarov:
@klihub @marquiz @kad I updated the PR by removing the usage of nsenter and did some other clean ups. PTAL.

pkg/utils/config-manager/main.go
```diff
@@ -0,0 +1,119 @@
...
+	var buf bytes.Buffer
+	enc := tomlv2.NewEncoder(&buf)
+	enc.SetIndentTables(true)
+	if err := enc.Encode(config); err != nil {
+		return fmt.Errorf("error encoding file: %w", err)
+	}
+
+	f, err := os.Create(file)
+	if err != nil {
+		log.Fatalf("error truncating the file: %q: %w", file, err)
```
klihub, line 53: Why the different handling here then on the other error branches ?

fmuyassarov: ah I missed to update this one, let me fix it quickly. Thanks for the catch

fmuyassarov: fixed.

deployment/helm/resource-management-policies/balloons/templates/daemonset.yaml
```diff
@@ -78,3 +91,13 @@ spec:
...
           type: Directory
+      {{- if .Values.nri.patchContainerdConfig }}
+      - name: containerd-config
+        hostPath:
+          path: /etc/containerd/config.toml
+          type: File
+      - name: dbus-sockets
+        hostPath:
+          path: /var/run/dbus
+          type: Directory
```
marquiz, line 102: nit: why not mount the file that we really need? This would verify its existence (and type) on the host
```suggestion
          path: /run/dbus/system_bus_socket
          type: Socket
```

klihub:
One last nitpick about `writeConfig()`. And sorry for dropping these comments so annoyingly piecemeal.

Ideally we should leave the original file intact if something goes wrong. Also, ideally atomically overwriting the content (creating a new file, writing the content to it, then renaming the file) is often better.  Could be done something like this...

```golang
func writeConfig(file string, config map[string]interface{}) error {
    var (
        tmp  = file + ".tmp"
        perm = os.FileMode(0o644)
    )

    fi, err := os.Stat(file)
    if err == nil {
        perm = fi.Mode().Perm()
    }

    f, err = os.OpenFile(tmp, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, perm)
    if err != nil {
        return fmt.Errorf("failed to create %s: %w", tmp, err)
    }
    defer func() {
        if f != nil {
            f.Close()
        }
    }()

    err = tomlv2.NewEncoder(f).SetIndentTables(true).Encode(config)
    if err != nil {
        return fmt.Errorf("failed to encode/write toml data to %s: %w", tmp, err)
    }

    f.Close()
    f = nil

    err = os.Rename(tmp, file)
    if err != nil {
        return fmt.Errorf("failed to rename %s to %s: %w", tmp, file, err)
    }

    return nil
}
```

Instead of the extra `Stat()` before writing, you could alternatively modify `readConfig()` to return both the content and the permissions. If you'd do reading and decoding at the same time (`f := Open(file); defer f.Close(); err := tomlv2.NewDecoder(f).Decode(&tomlMap); fi := f.Stat(); perm := fi.Mode().Perm()`), then it would be easy to dig out with a `Stat()` there.

deployment/helm/resource-management-policies/topology-aware/values.yaml
```diff
@@ -18,3 +18,11 @@ hostPort: 8891
 resources:
   cpu: 500m
   memory: 512Mi
+
+nri:
+  patchContainerdConfig: false
+
+init:
```
marquiz, line 25: nit: plain `init` looks somehow ambiguous to me. How about `initContainerImage` or `initImage` for example (we would then have `image` and `initContainerImage` fields )?

```suggestion
initContainerImage:
```

pkg/utils/config-manager/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Build Stage
+FROM golang:1.19.0 AS build
```
marquiz, line 2: I'd suggerst to use the same golang version as for other images

```suggestion
FROM golang:${GO_VERSION}-buster as build
```

marquiz (commented):
Thanks @fmuyassarov for the effort. A few comments below, nothing big

pkg/utils/config-manager/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Build Stage
```
klihub, line 1: These should all be in `cmd/config-manager` and not anywhere under `pkg` (it has a lot of non-golang cruft and the golang package is main).

deployment/helm/resource-management-policies/topology-aware/values.yaml
klihub: +1 for initContainerImage.

fmuyassarov: replaced to be `initContainerImage`

pkg/utils/config-manager/Dockerfile
fmuyassarov: done

fmuyassarov:
Yes, that was the reason indeed.

@kad we have agreed with @klihub that current state is good enough to merge the PR and I have tested it Azure AKS to check the functionality and sent you the recording yesterday.

Perhaps we can take this in if there are no more comments left to address?

cmd/config-manager/go.mod
```diff
@@ -0,0 +1,10 @@
+module nri/nri-plugins/cmd/config-manager
```
klihub, line 1: Wrong module path IMO.

klihub (changes requested):
I think you have the module path wrong.

cmd/config-manager/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Build Stage
+ARG GO_VERSION=1.19
+
+FROM golang:${GO_VERSION}-buster as build
```
klihub, line 4: Please use golang:${GO_VERSION}-bullseye for building like the rest of the plugins/binaries. That'll give us up to date golang toolchain.

klihub:
@fmuyassarov If you make `cmd/config-manager` a package of its own, you'll also need to stick a separate `LICENSE` file there for `go-licenses` to find. It can't extract licenses from go source code AFAICT.

fmuyassarov:
> I think you have the module path wrong.

That's right, fixed.

marquiz:
> @fmuyassarov If you make `cmd/config-manager` a package of its own, you'll also need to stick a separate `LICENSE` file there for `go-licenses` to find. It can't extract licenses from go source code AFAICT.

I don't see much reason to make the config-mananger a separate module. IMHO just so much easier (to maintain) and cleaner to have everythin under the same top-level module. We can merge this PR whichever way but there might be a follow-up PR dropping the separate module...

fmuyassarov:
> > @fmuyassarov If you make `cmd/config-manager` a package of its own, you'll also need to stick a separate `LICENSE` file there for `go-licenses` to find. It can't extract licenses from go source code AFAICT.
> 
> I don't see much reason to make the config-mananger a separate module. IMHO just so much easier (to maintain) and cleaner to have everythin under the same top-level module. We can merge this PR whichever way but there might be a follow-up PR dropping the separate module...

Let's fix it now.

mythi:
FYI, we had a similar use-case where restarting containerd from a container was needed. The dbus approach turned out to be problematic and we eventually moved to use host systemctl instead: https://github.com/confidential-containers/operator/pull/242

marquiz:
> FYI, we had a similar use-case where restarting containerd from a container was needed. The dbus approach turned out to be problematic and we eventually moved to use host systemctl instead: [confidential-containers/operator#242](https://github.com/confidential-containers/operator/pull/242)

But you were using `systemctl` from the container. We're using dbus directly (afaiu, with the go-systemd) so I think we should have small(er) risk for api incompatibility. I'd stick with what we have until proven not working

mythi:
Makes sense to keep what's there today, just shared what we did. We ran into nodes where dbus was not available.

fmuyassarov:
Thanks @mythi for sharing. Initially I was using nsenter (via host systemctl) similar to the patch you shared but then we decided to rely on dbus as more reliable approach. But we will see of course in the future if some people run into issues with that.

PR #109, opened 2023-09-01
build: band-aid build system to allow building binaries other than resource policy plugins.

klihub:
This PR aims to relax some of the most restrictive implicit assumptions about the types of binaries we build, and with fairly minimal changes to the Makefile. As a result, we now can have
- resource policy plugins, located in cmd/plugins, binaries named nri-resource-policy-*cmd/plugins
- other types of plugins, located in cmd/plugins, binaries named nri-*
- other binaries, located in cmd, binaries named something else than nri-*

The PR has one extra commit to remove obviously unused cruft from the `Makefile`, but otherwise it is fairly minimal. All other relevant changes are limited to the `Makefile` and to updating the import path of the plugins. All other file changes are moving plugins around (from `cmd` to `cmd/plugins`).


Makefile
```diff
@@ -143,18 +150,40 @@ clean-plugins:
...
 
+$(BIN_PATH)/nri-resource-policy-%: .static.%.$(STATIC)
+	$(Q)echo "Building $(STATIC_TYPE)$@ (version $(BUILD_VERSION), build $(BUILD_BUILDID))..."; \
+	src="./cmd/plugins/$(patsubst nri-resource-policy-%,%,$(notdir $@))"; \
+	mkdir -p $(BIN_PATH); \
+	cd "$$src" && $(GO_BUILD) $(BUILD_TAGS) $(LDFLAGS) $(GCFLAGS) -o $@
+
+$(BIN_PATH)/nri-%: .static.%.$(STATIC)
+	$(Q)echo "Building $(STATIC_TYPE)$@ (version $(BUILD_VERSION), build $(BUILD_BUILDID))..."; \
+	src="./cmd/plugins/$(patsubst nri-%,%,./cmd/plugins/$(notdir $@))"; \
```
askervin, line 178:
```
src="./cmd/plugins/$(patsubst nri-%,%,./cmd/plugins/$(notdir $@))";
->
src="./cmd/plugins/$(patsubst nri-%,%,$(notdir $@))";
```

askervin (commented):
Minor fix needed for building non-resource-manager NRI plugins.

Makefile
klihub: Thx. Fixed.

askervin (approved):
Thanks @klihub for this work! I'm good to go with these changes.

At some point I'd like to allow overriding `IMAGE_VERSION` (just replace `:=` with `?=` in order to tag images with `e2e` or whatever...), but I'd vote for merging this as is already now and tune little things later.

klihub:
> Thanks @klihub for this work! I'm good to go with these changes.
> 
> At some point I'd like to allow overriding `IMAGE_VERSION` (just replace `:=` with `?=` in order to tag images with `e2e` or whatever...), but I'd vote for merging this as is already now and tune little things later.

I don't think that using the `?=` conditional assignment operator would be necessary for that. I think a simply expanded variable assignment can be overridden from the command line as usual:

```diff
klitkey1-mobl1 nri-plugins $ git diff 
diff --git a/Makefile b/Makefile
index b77240b..a7035b5 100644
--- a/Makefile
+++ b/Makefile
@@ -429,3 +429,6 @@ site-serve: .$(DOCKER_SITE_BUILDER_IMAGE).image.stamp
        touch $@
 
 docs: site-build
+
+show:
+       $(Q)echo "$(VAR)=$($(VAR))"
klitkey1-mobl1 nri-plugins $ make show VAR=IMAGE_VERSION
IMAGE_VERSION=v0.1.0-30-g1da8430-dirty
klitkey1-mobl1 nri-plugins $ make IMAGE_VERSION=my-image-version show VAR=IMAGE_VERSION
IMAGE_VERSION=my-image-version

```

Makefile
```diff
@@ -143,18 +152,40 @@ clean-plugins:
...
 # plugins build targets
 #
 
+$(BIN_PATH)/nri-resource-policy-%: .static.%.$(STATIC)
+	$(Q)echo "Building $(STATIC_TYPE)$@ (version $(BUILD_VERSION), build $(BUILD_BUILDID))..."; \
+	src="./cmd/plugins/$(patsubst nri-resource-policy-%,%,$(notdir $@))"; \
+	mkdir -p $(BIN_PATH); \
+	cd "$$src" && $(GO_BUILD) $(BUILD_TAGS) $(LDFLAGS) $(GCFLAGS) -o $@
+
+$(BIN_PATH)/nri-%: .static.%.$(STATIC)
```
marquiz, line 178: What is this rule for? Should/could we rename `config-manager` to `nri-config-manager` and modify this target to build that (and possibly drop the target below)?

marquiz:
> > At some point I'd like to allow overriding `IMAGE_VERSION` (just replace `:=` with `?=` in order to tag images with `e2e` or whatever...), but I'd vote for merging this as is already now and tune little things later.
> 
> @askervin I don't think that using the `?=` conditional assignment operator would be necessary for that. I think a simply expanded variable assignment can be overridden from the command line as usual:

I think one practical difference is that with `?=` you can also specify variables from the environment, not only as make variable overrides, i.e. both `IMAGE_VERSION=foo make images` and `make images IMAGE_VERSION=foo` would work

Makefile
klihub: No. According to the current (documented) conventions, only plugins should be named `nri-%`. Everything else is some other helper/nri-related binary.

klihub:
> > > At some point I'd like to allow overriding `IMAGE_VERSION` (just replace `:=` with `?=` in order to tag images with `e2e` or whatever...), but I'd vote for merging this as is already now and tune little things later.
> > 
> > 
> > @askervin I don't think that using the `?=` conditional assignment operator would be necessary for that. I think a simply expanded variable assignment can be overridden from the command line as usual:
> 
> I think one practical difference is that with `?=` you can also specify variables from the environment, not only as make variable overrides, i.e. both `IMAGE_VERSION=foo make images` and `make images IMAGE_VERSION=foo` would work

That's a good point. I was not aware of that. If we want to have the ability to inherit some of the variables from the environment, we'll need to decide which ones make sense then consistently update the assignment of those to use the `?=`-operator.

marquiz (approved):
OK, LGTM

PR #114, opened 2023-09-19
github: use the git checkout action to fetch all refs

marquiz:
Drop the separate step for fetching gh-pages branch.

Also update the checkout action to the latest version.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -15,10 +15,9 @@ jobs:
   update-gh-pages:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-
-    - name: Fetch gh-pages
-      run: git fetch --no-tags --prune --depth=1 origin refs/heads/gh-pages:refs/heads/gh-pages
+    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -12,7 +12,9 @@ jobs:
   update-gh-pages:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
fmuyassarov, line 17:
```suggestion
    - uses: actions/checkout@v4
      with:
        fetch-depth: 0
        ref: gh-pages
```
WDYT? To be safe that we are not the right branch?

marquiz: The script we run comes from the ref that was pushed. It does not exist in gh-pages

fmuyassarov: ah right.

fmuyassarov:
LGTM

kad (approved):
LGTM

marquiz:
Rebased

PR #118, opened 2023-09-28
helm: add charts for memtierd and memory-qos

fmuyassarov:
This commit adds two new Helm charts for installing memory-qos(https://github.com/containers/nri-plugins/pull/115) and memtierd(https://github.com/containers/nri-plugins/pull/117) NRI plugins via Helm.

### DIFF
--- a/deployment/helm/memory-qos/Chart.yaml
+++ b/deployment/helm/memory-qos/Chart.yaml
@@ -1,0 +1,11 @@
+apiVersion: v2
+appVersion: "main"
+description: |
+  The memory-qos NRI plugin adds two methods for controlling cgroups v2
+  memory.* parameters: QoS class and direct memory annotations.
+name: nri-memory-qos
+sources:
+ - https://github.com/containers/nri-plugins
+home: https://github.com/containers/nri-plugins
+type: application
+version: v0.0.0
--- a/deployment/helm/memory-qos/templates/_helpers.tpl
+++ b/deployment/helm/memory-qos/templates/_helpers.tpl
@@ -1,0 +1,16 @@
+{{/*
+Common labels
+*/}}
+{{- define "memory-qos.labels" -}}
+helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{ include "memory-qos.selectorLabels" . }}
+{{- end -}}
+
+{{/*
+Selector labels
+*/}}
+{{- define "memory-qos.selectorLabels" -}}
+app.kubernetes.io/name: nri-memory-qos
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end -}}
--- a/deployment/helm/memtierd/Chart.yaml
+++ b/deployment/helm/memtierd/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v2
+appVersion: "main"
+description: |
+  The memtierd NRI plugin enables managing workloads with Memtierd in Kubernetes.
+name: nri-memtierd
+sources:
+ - https://github.com/containers/nri-plugins
+home: https://github.com/containers/nri-plugins
+type: application
+version: v0.0.0
--- a/deployment/helm/memtierd/templates/daemonset.yaml
+++ b/deployment/helm/memtierd/templates/daemonset.yaml
@@ -1,0 +1,77 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    {{- include "memtierd.labels" . | nindent 4 }}
+  name: nri-memtierd
+  namespace: {{ .Release.Namespace }}
+spec:
+  selector:
+    matchLabels:
+    {{- include "memtierd.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+      {{- include "memtierd.labels" . | nindent 8 }}
+    spec:
+      nodeSelector:
+        kubernetes.io/os: "linux"
+      hostPID: true
+      containers:
+        - name: nri-memtierd
+          command:
+            - nri-memtierd
+            - --idx
+            - "45"
+            - --config
+            - /etc/nri/memtierd/config.yaml
+            {{- if .Values.outputDir }}
+            - --run-dir
+            - /run-dir
+            {{- end }}
+            - --cgroups-dir
+            - /sys/fs/cgroup
+            - -v
+          image: {{ .Values.image.name }}:{{ .Values.image.tag | default .Chart.AppVersion }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          resources:
+            requests:
+              cpu: {{ .Values.resources.cpu }}
+              memory: {{ .Values.resources.memory }}
+          securityContext:
+            privileged: true
+          volumeMounts:
+          - name: memtierd-config-vol
+            mountPath: /etc/nri/memtierd
+          - name: nrisockets
+            mountPath: /var/run/nri
+          - name: host-bitmap
+            mountPath: /sys/kernel/mm/page_idle/bitmap
+          - name: cgroups-vol
+            mountPath: /sys/fs/cgroup
+          {{- if .Values.outputDir }}
+          # Set outputDir parameter to access memtierd.output files from host
+          - name: run-dir-vol
+            mountPath: /run-dir
+          {{- end }}
+      volumes:
+      - name: memtierd-config-vol
+        configMap:
+          name: nri-memtierd-config.default
+      - name: nrisockets
+        hostPath:
+          path: /var/run/nri
+          type: Directory
+      - name: host-bitmap
+        hostPath:
+          path: /sys/kernel/mm/page_idle/bitmap
+      - name: cgroups-vol
+        hostPath:
+          path: /sys/fs/cgroup
+          type: Directory
+      {{- if .Values.outputDir }}
+      - name: run-dir-vol
+        hostPath:
+          path: {{ .Values.outputDir }}
+          type: DirectoryOrCreate
+      {{- end }}
--- a/docs/resource-policy/installation.md
+++ b/docs/resource-policy/installation.md
@@ -103,6 +103,27 @@ along with the default values, for the Topology-aware and Balloons plugins Helm 
 | `hostPort`         | 8891                                                                                                                          | metrics port to expose on the host                   |
 | `config`           | <pre><code>ReservedResources:</code><br><code>  cpu: 750m</code></pre>                                                        | plugin configuration data                            |
 
+#### Memtierd
+
+| Name               | Default                                                                                                                       | Description                                          |
+| ------------------ | ----------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- |
+| `image.name`       | [ghcr.io/containers/nri-plugins/nri-memtierd](ghcr.io/containers/nri-plugins/nri-memtierd)                                    | container image name                                 |
+| `image.tag`        | unstable                                                                                                                      | container image tag                                  |
+| `image.pullPolicy` | Always                                                                                                                        | image pull policy                                    |
+| `resources.cpu`    | 250m                                                                                                                          | cpu resources for the Pod                            |
+| `resources.memory` | 100Mi                                                                                                                         | memory qouta for the                                 |
+| `outputDir`        | empty string                                                                                                                  | host directory for memtierd.output files             |
+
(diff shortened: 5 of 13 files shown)


docs/resource-policy/installation.md
```diff
@@ -103,6 +103,26 @@ along with the default values, for the Topology-aware and Balloons plugins Helm
 | `hostPort`         | 8891                                                                                                                          | metrics port to expose on the host                   |
 | `config`           | <pre><code>ReservedResources:</code><br><code>  cpu: 750m</code></pre>                                                        | plugin configuration data                            |
 
+#### Memtierd
+
+| Name               | Default                                                                                                                       | Description                                          |
+| ------------------ | ----------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- |
+| `image.name`       | [ghcr.io/containers/nri-plugins/nri-resource-policy-memtierd](ghcr.io/containers/nri-plugins/nri-resource-policy-balloons)    | container image name                                 |
```
kad, line 110: it's not really resource policy.

docs/resource-policy/installation.md
```diff
@@ -103,6 +103,26 @@ along with the default values, for the Topology-aware and Balloons plugins Helm
...
+| `resources.cpu`    | 250m                                                                                                                          | cpu resources for the Pod                            |
+| `resources.memory` | 100Mi                                                                                                                         | memory qouta for the                                 |
+
+#### Memory-qos
+
+| Name               | Default                                                                                                                       | Description                                          |
+| ------------------ | ----------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- |
+| `image.name`       | [ghcr.io/containers/nri-plugins/nri-resource-policy-memory-qos](ghcr.io/containers/nri-plugins/nri-resource-policy-balloons)  | container image name                                 |
```
kad, line 120: same here. it is nri memory management plugin, but not resource policy.

docs/resource-policy/installation.md
```diff
@@ -103,6 +103,26 @@ along with the default values, for the Topology-aware and Balloons plugins Helm
...
+| `image.tag`        | unstable                                                                                                                      | container image tag                                  |
+| `image.pullPolicy` | Always                                                                                                                        | image pull policy                                    |
+| `resources.cpu`    | 10m                                                                                                                          | cpu resources for the Pod                            |
```
kad, line 123: would it be enough to have it as besteffort container?

docs/resource-policy/installation.md
fmuyassarov: Aha good to know. I will fix it here and in other locations.

docs/resource-policy/installation.md
fmuyassarov: Probably @askervin can give better answer :)

askervin: `nri-memory-qos` could indeed be a besteffort container, because it is very-very lightweight task that it is handling.

`nri-memtierd` is another story. Setting resources.limits works infact as a convenient way to make sure that memtierd processes would not consume too much CPU from the node. On the other hand, deciding what would be a proper portion of CPU that should be sufficient, depends heavily on user needs and other workloads (most essentially how much there is memory to scan). Today I would vote for _limiting_ CPU usage to something like 250m or 500m, but not necessarily _requesting_ any CPU, so that deploying the plugin would not be prevented due to exhausted reserved CPU resources on a node. @kad , what do you think?

docs/resource-policy/installation.md
askervin: Coo, replacing every "`resource-policy-`" with empty string should do the trick.

docs/resource-policy/installation.md
kad: my point was for "memtier-qos" container, which is just very simple, modifies only cgroupsv2 parameters. and thus best effort: neither cpu nor memory request.

for full memtierd, yes some minimal requirements good to have plus some way to limit burstability.

BTW, remove references to "balloons" in the URLs in markdown.

docs/resource-policy/installation.md
fmuyassarov: @kad TA and Balloons are classified as nri-resource-management. 
Are you suggesting to have another group(and directory) called something like nri-memory-management?

fmuyassarov:
@kad @askervin PR is now updated. But I have not added https://github.com/containers/nri-plugins/pull/121 changes, which I think we need. But since it is under active reviews, perhaps we can do it once the patch has landed.

Both plugins are now outside the resource management dir, and the corresponding images are called
`ghcr.io/containers/nri-plugins/nri-memtierd` & `ghcr.io/containers/nri-plugins/nri-memory-qos`. 

Same with the labels and DaemonSet object naming, `nri-memtierd` & `nri-memory-qos`.

askervin:
Cool, thanks @fmuyassarov! This works fine already. The last wishes from me:
- Would you add the [track-working-set-size](https://github.com/askervin/nri-plugins/blob/5RA_memtierd_host_root/deployment/overlays/memtierd/sample-configmap.yaml#L31) class to deployments/helm/memtierd/templates/configmap.yaml
- Is it possible to make [this](https://github.com/askervin/nri-plugins/blob/5RA_memtierd_host_root/deployment/overlays/memtierd/daemonset.yaml#L46) volume mount a configurable option in helm installation?

fmuyassarov:
> Cool, thanks @fmuyassarov! This works fine already. The last wishes from me:
> 
>     * Would you add the [track-working-set-size](https://github.com/askervin/nri-plugins/blob/5RA_memtierd_host_root/deployment/overlays/memtierd/sample-configmap.yaml#L31) class to deployments/helm/memtierd/templates/configmap.yaml
> 
>     * Is it possible to make [this](https://github.com/askervin/nri-plugins/blob/5RA_memtierd_host_root/deployment/overlays/memtierd/daemonset.yaml#L46) volume mount a configurable option in helm installation?
> 
> 
> After adding the track-working-set-size class and `helm install` it should be possible to deploy test/e2e/files/nri-memtierd-test-pod.yaml without errors.
> 
> Deploying test/e2e/files/nri-memory-qos-test-pod.yaml already works fine after `helm install`ing nri-memory-qos. Great!

@askervin done. PTAL.

askervin:
@fmuyassarov, sorry, I overlooked a couple of things... there are also other cases where `deployments/helm/memtierd/templates/daemonset.yaml` has diverged from [deployment/overlays/memtierd/daemonset.yaml](https://github.com/askervin/nri-plugins/blob/5RA_memtierd_host_root/deployment/overlays/memtierd/daemonset.yaml).

We do not want to pass the `--host-root` parameter, nor do we want to mount the whole host root filesystem. Also the `--run-dir /run-dir` is an essential argument in order to get memtierd.output files into the directory that user gives with outputDir to helm.

Setting outputDir seems to work here nicely.

deployment/helm/memtierd/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+appVersion: "main"
+description: |
+  The memtierd NRI plugin enables managing workloads with Memtierd in Kubernetes.
+name: nri-resource-policy-memtierd
```
marquiz, line 5: nri-memtierd?

deployment/helm/memtierd/templates/daemonset.yaml
```diff
@@ -0,0 +1,79 @@
...
+      - name: memtierd-config-vol
+        configMap:
+          name: nri-memtierd-config.default
+      - name: nrisockets
+        hostPath:
+          path: /var/run/nri
+          type: Directory
+      - name: host-vol
+        hostPath:
+          path: /
```
marquiz, line 65: Let's get rid of this mount

deployment/helm/memory-qos/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+appVersion: "main"
+description: |
+  The memory-qos NRI plugin adds two methods for controlling cgroups v2
+  memory.* parameters: QoS class and direct memory annotations.
+name: nri-resource-policy-memory-qos
```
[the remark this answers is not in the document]
fmuyassarov: good catch. Fixing it...

deployment/helm/memory-qos/templates/_helpers.tpl
```diff
@@ -0,0 +1,8 @@
+{{/*
+Common labels
+*/}}
+{{- define "memory-qos.labels" -}}
+app: nri-memory-qos
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
```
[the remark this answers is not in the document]
fmuyassarov: Hmm good question...TBH, I don't have a strong opinion.

deployment/helm/memtierd/templates/daemonset.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - name: host-bitmap
+        hostPath:
+          path: /sys/kernel/mm/page_idle/bitmap
+      - name: cgroups-vol
+        hostPath:
+          path: /sys/fs/cgroup
+          type: Directory
+      {{- if .Values.outputDir }}
+      - name: run-dir-vol
+        hostPath:
```
marquiz, line 74: My question (primarily to @askervin) regarding this was should we have emptyDir mount by default, to persist data over container restarts?

askervin:
> @askervin I think we also need ` - --cgroups-dir - /sys/fs/cgroup` ?

Yes we do.

It's good to check that `helm template --set outputDir=/var/run/nri-memtierd nri-memtierd deployment/helm/memtierd` output includes the daemonset that will be created wtih kustomize, too, but it actually has the outputDir mounted that is just commented in the kustomize [daemonset](https://github.com/askervin/nri-plugins/blob/5RA_memtierd_host_root/deployment/overlays/memtierd/daemonset.yaml).

deployment/helm/memtierd/templates/daemonset.yaml
marquiz: Got that.

deployment/helm/memory-qos/templates/_helpers.tpl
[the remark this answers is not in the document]
marquiz: Now that we're talking about these, I took a look at https://helm.sh/docs/chart_best_practices/labels/. I'd suggest to add the recommended `helm.sh/chart` label, too.

One other change I'd suggest to do is to keep the labels and "selector labels" separate. Put  everything in labels but only `app.kubernetes.io/name` and possibly `app.kubernetes.io/instance` in selector labels. WDYT @fmuyassarov @askervin ?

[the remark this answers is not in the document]
fmuyassarov: Makes sense. Updated accordingly.

deployment/helm/memory-qos/templates/_helpers.tpl
```diff
@@ -0,0 +1,17 @@
+{{/*
+Common labels
+*/}}
+{{- define "memory-qos.labels" -}}
+helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+app.kubernetes.io/name: nri-memory-qos
+app.kubernetes.io/instance: {{ .Release.Name }}
```
marquiz, line 7: nit: I think we could include the selectorLabels here to keep these always automatically in sync. Like:

```suggestion
{{ include "memory-qos.selectorLabels" . }}
```

fmuyassarov: Good point. Fixed

docs/resource-policy/installation.md
marquiz, line 1: Just a mental note that the documentation is a total mess. Installation instructions are under "Resource policy plugins" but the actual plugin documentation under "Memory plugins" 🤪 I don't think should fix the mess in this PR

marquiz (approved):
Thanks @fmuyassarov for working on this. I don't have anything to add for the initial version. We should carry over some of the things (like labels vs selectorLabels) to the other charts as well.
LGTM

docs/resource-policy/installation.md
fmuyassarov: ow right. I will submit a patch for that and also for selectorLabel for TA and balloons plugins. Thanks for noticing.

fmuyassarov:
> Thanks @fmuyassarov for working on this. I don't have anything to add for the initial version. We should carry over some of the things (like labels vs selectorLabels) to the other charts as well. LGTM

Submitted in https://github.com/containers/nri-plugins/pull/124. Thanks for the reviews @marquiz

PR #120, opened 2023-09-28
deployment: refactor config manager to support NRI enabling in CRI-O 

fmuyassarov:
This PR extends config manager code and the plugins helm charts
so that CRI-O users are also able to enable NRI via our charts if they
wish to. Same parameter is used to opt in for the feature in Helm
charts and we don't require users to indicate what container runtime
is being used. Instead the config manager auto-detects the runtime
and does the necessary changes to its configuration file. In scenarios
with multiple active runtimes (e.g., CRI-O and containerd), the
manager gracefully exits and throws an error.

### DIFF
--- a/cmd/config-manager/main.go
+++ b/cmd/config-manager/main.go
@@ -29,33 +29,71 @@ import (
 )
 
 const (
-	tomlFilePath = "/etc/containerd/config.toml"
-	nriPluginKey = "io.containerd.nri.v1.nri"
-	disableKey   = "disable"
-	replaceMode  = "replace"
-	resultDone   = "done"
-	unit         = "containerd.service"
+	containerdConfigFile = "/etc/containerd/config.toml"
+	crioConfigFile       = "/etc/crio/crio.conf.d/10-enable-nri.conf"
+	nriPluginKey         = "io.containerd.nri.v1.nri"
+	replaceMode          = "replace"
+	resultDone           = "done"
+	containerdUnit       = "containerd.service"
+	crioUnit             = "crio.service"
 )
 
 func main() {
-	tomlMap, err := readConfig(tomlFilePath)
+	unit, err := detectRuntime()
 	if err != nil {
-		log.Fatalf("Error reading TOML file: %v", err)
+		log.Fatalf("failed to autodetect container runtime: %v", err)
 	}
 
-	updatedTomlMap := updateNRIPlugin(tomlMap)
-
-	err = writeConfig(tomlFilePath, updatedTomlMap)
-	if err != nil {
-		log.Fatalf("failed to write updated config into a file %q:, %v", tomlFilePath, err)
+	switch unit {
+	case containerdUnit:
+		err = enableNriForContainerd()
+	case crioUnit:
+		err = enableNriForCrio()
+	default:
+		log.Fatalf("unknown container runtime %q", unit)
 	}
 
-	err = restartSystemdUnit(unit)
 	if err != nil {
-		log.Fatalf("failed to restart containerd: %v", err)
+		log.Fatalf("error enabling NRI: %v", err)
 	}
+
+	if err = restartSystemdUnit(unit); err != nil {
+		log.Fatalf("failed to restart %q unit: %v", unit, err)
+	}
+
+	log.Println("enabled NRI for", unit)
 }
-func writeConfig(file string, config map[string]interface{}) error {
+
+func enableNriForContainerd() error {
+	tomlMap, err := readConfig(containerdConfigFile)
+	if err != nil {
+		return fmt.Errorf("error reading TOML file: %w", err)
+	}
+
+	updatedTomlMap := updateContainerdConfig(tomlMap)
+
+	err = writeToContainerdConfig(containerdConfigFile, updatedTomlMap)
+	if err != nil {
+		return fmt.Errorf("failed to write updated config into a file %q: %w", containerdConfigFile, err)
+	}
+	return nil
+}
+
+func enableNriForCrio() error {
+	f, err := os.Create(crioConfigFile)
+	if err != nil {
+		return fmt.Errorf("error creating a drop-in file for CRI-O: %w", err)
+	}
+	defer f.Close()
+
+	_, err = f.WriteString("[crio.nri]\nenable_nri = true\n")
+	if err != nil {
+		return fmt.Errorf("error writing a drop-in file for CRI-O: %w", err)
+	}
+	return nil
+}
+
+func writeToContainerdConfig(file string, config map[string]interface{}) error {
 	var buf bytes.Buffer
 	enc := tomlv2.NewEncoder(&buf)
 	enc.SetIndentTables(true)
@@ -90,10 +128,10 @@ func readConfig(file string) (map[string]interface{}, error) {
 	return tomlMap, nil
 }
 
-func updateNRIPlugin(config map[string]interface{}) map[string]interface{} {
+func updateContainerdConfig(config map[string]interface{}) map[string]interface{} {
 	plugins, exists := config["plugins"].(map[string]interface{})
 	if !exists {
-		log.Println("Top level plugins section not found, adding it to enable NRI...")
+		log.Println("top level plugins section not found, adding it to enable NRI...")
 		plugins = make(map[string]interface{})
 		config["plugins"] = plugins
 	}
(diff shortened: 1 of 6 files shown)


cmd/config-manager/main.go
```diff
@@ -90,23 +117,41 @@ func readConfig(file string) (map[string]interface{}, error) {
...
-		plugins[nriPluginKey] = nri
+		plugins[containerdNriTable] = nri
+	}
+
+	nri[containerdNriKey] = false
+	log.Println("Enabled NRI in containerd...")
+	return config
+}
+
+func patchCrioConfig(config map[string]interface{}) map[string]interface{} {
```
klihub, line 140: This would be less intrusive (wrt. losing comments in the config file, etc.) if you used a cri-o configuration drop-in snippet instead of rewriting the main configuration file:

```bash
test-node ~ $ cat /etc/crio/crio.conf.d/10-enable-nri.conf
[crio.nri]
enable_nri = true
```

kad:
just as an idea: would it make sense to try to autodetect?

cmd/config-manager/main.go
fmuyassarov: We can add a drop in file but what about the restarting the CRI-O afterwards from within the Pod?

fmuyassarov: One option is, we mount the drop in file and from within the main.go we do the restart if the runtime is CRI-O.

fmuyassarov: Based on the last call, I will replace the current table modification via drop in file.

fmuyassarov:
> just as an idea: would it make sense to try to autodetect?

Done. PTAL

fmuyassarov:
Rebased on top of the latest main.

cmd/config-manager/main.go
```diff
@@ -29,33 +29,66 @@ import (
...
-	err = writeConfig(tomlFilePath, updatedTomlMap)
+		updatedTomlMap := updateContainerdConfig(tomlMap)
+
+		err = writeToContainerdConfig(containerdConfigFile, updatedTomlMap)
+		if err != nil {
+			log.Fatalf("failed to write updated config into a file %q:, %v", containerdConfigFile, err)
+		}
+	}
+
+	err = writeToCrioConfig()
```
klihub, line 62: Isn't this now unconditionally always trying to update the CRI-O configuration ?

cmd/config-manager/main.go
```diff
@@ -29,33 +29,66 @@ import (
...
+			log.Fatalf("Error reading TOML file: %v", err)
+		}
 
-	err = writeConfig(tomlFilePath, updatedTomlMap)
+		updatedTomlMap := updateContainerdConfig(tomlMap)
+
+		err = writeToContainerdConfig(containerdConfigFile, updatedTomlMap)
+		if err != nil {
+			log.Fatalf("failed to write updated config into a file %q:, %v", containerdConfigFile, err)
+		}
```
klihub, line 59: Wouldn't it be easier to understand what is going on here if you only called either a containerd- or a crio-specific configuration update function (something like `updateContainerdConfig()` and `updateCRIOConfig()`) based on the detection result in and then do the unit restart if everything went fine....  You could then fold the read/modify/write back block for containerd now directly in main into `updateContainerdConfig()` and  `writeToCrioConfig()` could become `updateCRIOConfig()` as such.

cmd/config-manager/main.go
```diff
@@ -29,33 +29,79 @@ import (
...
+		return fmt.Errorf("error reading TOML file: %v", err)
 	}
 
-	err = restartSystemdUnit(unit)
+	updatedTomlMap := updateContainerdConfig(tomlMap)
+
+	err = writeToContainerdConfig(containerdConfigFile, updatedTomlMap)
 	if err != nil {
-		log.Fatalf("failed to restart containerd: %v", err)
+		return fmt.Errorf("failed to write updated config into a file %q:, %v", containerdConfigFile, err)
```
klihub, line 77: nit: unintentional '%q:, %v', guessing was meant to be '%q: %v'

fmuyassarov: thanks, fixed

klihub (approved):

fmuyassarov:
cc @marquiz

cmd/config-manager/main.go
```diff
@@ -29,33 +29,79 @@ import (
...
+		return fmt.Errorf("error reading TOML file: %v", err)
 	}
 
-	err = restartSystemdUnit(unit)
+	updatedTomlMap := updateContainerdConfig(tomlMap)
+
+	err = writeToContainerdConfig(containerdConfigFile, updatedTomlMap)
 	if err != nil {
-		log.Fatalf("failed to restart containerd: %v", err)
+		return fmt.Errorf("failed to write updated config into a file %q: %v", containerdConfigFile, err)
```
marquiz, line 77:
```suggestion
		return fmt.Errorf("failed to write updated config into a file %q: %w", containerdConfigFile, err)
```

marquiz (commented):
Thx @fmuyassarov. A few small comments below

deployment/helm/balloons/templates/daemonset.yaml
```diff
@@ -17,14 +17,15 @@ spec:
...
       initContainers:
-      - name: patch-containerd
+      - name: patch-runtime
         image: {{ .Values.initContainerImage.name }}:{{ .Values.initContainerImage.tag | default .Chart.AppVersion }}
         imagePullPolicy: {{ .Values.initContainerImage.pullPolicy }}
+        restartPolicy: Never
         volumeMounts:
-        - name: containerd-config
-          mountPath: /etc/containerd/config.toml
+        - name: etc
```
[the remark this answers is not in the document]
fmuyassarov: Yes, this is a good point. Replaced.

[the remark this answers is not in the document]
fmuyassarov: @marquiz Actually now that I think a bit more, if we are running on a host where there is no containerd we still have to create containerd config.toml forcefully via FileOrCreate and the vice versa to ensure that init containerd doesn't fail. Whereas mounting full etc would not require that.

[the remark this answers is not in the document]
marquiz: How about mounting `/etc/containerd/` with `DirectoryOrCreate`, WDYT?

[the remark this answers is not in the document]
fmuyassarov: Isn't it the same thing. If containerd isn't installed then there would not be /etc/containerd right? Of course, we can use DirectoryOrCreate but then if it is CRI-O, then we leave the host with unnecessary directory created.

deployment/helm/balloons/templates/daemonset.yaml
```diff
@@ -91,11 +94,15 @@ spec:
         hostPath:
           path: /var/run/nri
           type: DirectoryOrCreate
-      {{- if .Values.nri.patchContainerdConfig }}
+      {{- if .Values.nri.patchRuntimeConfig }}
       - name: containerd-config
         hostPath:
           path: /etc/containerd/config.toml
-          type: File
+          type: FileOrCreate
```
marquiz, line 101: After chatting with @fmuyassarov:

```suggestion
          path: /etc/containerd/
          type: DirectoryOrCreate
```

WDYT @klihub @kad @askervin? Now this would be safer but create empty config dir for the "other" runtime

askervin: Creating an empty directory on host doesn't sound a major sin to me. Creating an empty file would be worse as it might conflict with packages. And in my opinion mounting full /etc would be worse in security point of view. So this would be the best solution I can think of.

marquiz: > Creating an empty directory on host doesn't sound a major sin to me.

My thinking, too

deployment/helm/topology-aware/values.yaml
```diff
@@ -20,7 +20,7 @@ resources:
   memory: 512Mi
 
 nri:
-  patchContainerdConfig: false
+  patchRuntime: false
```
askervin, line 23: Should `patchRuntime` be `patchRuntimeConfig`?

deployment/helm/balloons/values.yaml
```diff
@@ -20,7 +20,8 @@ resources:
   memory: 512Mi
 
 nri:
-  patchContainerdConfig: false
+  patchRuntime: false
```
askervin, line 23: `patchRuntimeConfig`?

deployment/helm/balloons/templates/daemonset.yaml
```diff
@@ -91,11 +94,15 @@ spec:
...
+      - name: crio-config
+        hostPath:
+          path: /etc/crio/crio.conf.d/
+          type: Directory
```
askervin, line 105: `DirectoryOrCreate`? Doesn't container creation fail otherwise if this directory does not exist?

askervin (commented):
Few questions.

deployment/helm/balloons/templates/daemonset.yaml
fmuyassarov: Agreed, fixed.

deployment/helm/balloons/templates/daemonset.yaml
fmuyassarov: Yes it will. fixed.

deployment/helm/topology-aware/values.yaml
fmuyassarov: good catch.

marquiz (approved):
LGTM

But should we have the same thing for memtierd and memory-qos plugins (and the example plugin)?

fmuyassarov:
> LGTM
> 
> But should we have the same thing for memtierd and memory-qos plugins (and the example plugin)?

I'm +1 for that. @kad @klihub @askervin thoughts?

fmuyassarov:
> LGTM
> 
> But should we have the same thing for memtierd and memory-qos plugins (and the example plugin)?

Regardless of that, can we perhaps proceeding merging this PR if it looks good enough. And for adding init container to memtierd and memory-qos can be done as separate PR?

deployment/helm/balloons/templates/daemonset.yaml
kad: I think you're overcomplicating that. Just mount whole /etc. that directory must exist. Mounting it is also a common pattern, e.g. in CNI installation manifests.

klihub:
> > LGTM
> > But should we have the same thing for memtierd and memory-qos plugins (and the example plugin)?
> 
> I'm +1 for that. @kad @klihub @askervin thoughts?

I think it would be more consistent repo-behavior if all the helm charts would allow one to also enable NRI if necessary during deployment.

fmuyassarov:
submitted https://github.com/containers/nri-plugins/pull/131 for that

PR #121, opened 2023-09-30
nri-memtierd: mount only needed dirs from host

askervin:
Hardening nri-memtierd deployment.

### DIFF
--- a/deployment/overlays/memtierd/daemonset.yaml
+++ b/deployment/overlays/memtierd/daemonset.yaml
@@ -25,8 +25,10 @@ spec:
             - "45"
             - --config
             - /etc/nri/memtierd/config.yaml
-            - --host-root
-            - /host
+            - --run-dir
+            - /run-dir
+            - --cgroups-dir
+            - /sys/fs/cgroup
             - -v
           image: IMAGE_PLACEHOLDER
           imagePullPolicy: IfNotPresent
@@ -41,9 +43,12 @@ spec:
             mountPath: /etc/nri/memtierd
           - name: nri-sockets-vol
             mountPath: /var/run/nri
-          - name: host-vol
-            mountPath: /host
-          - name: host-bitmap
+          # # Uncomment to access memtierd.output files from host
+          # - name: run-dir-vol
+          #   mountPath: /run-dir
+          - name: cgroups-vol
+            mountPath: /sys/fs/cgroup
+          - name: idlepage-vol
             mountPath: /sys/kernel/mm/page_idle/bitmap
       volumes:
       - name: memtierd-config-vol
@@ -53,10 +58,15 @@ spec:
         hostPath:
           path: /var/run/nri
           type: Directory
-      - name: host-vol
+      # # Uncomment to access memtierd.output files from host
+      # - name: run-dir-vol
+      #   hostPath:
+      #     path: /var/run/nri-memtierd
+      #     type: DirectoryOrCreate
+      - name: cgroups-vol
         hostPath:
-          path: /
+          path: /sys/fs/cgroup
           type: Directory
-      - name: host-bitmap
+      - name: idlepage-vol
         hostPath:
           path: /sys/kernel/mm/page_idle/bitmap
--- a/deployment/overlays/memtierd/sample-configmap.yaml
+++ b/deployment/overlays/memtierd/sample-configmap.yaml
@@ -28,3 +28,27 @@ data:
             mover:
               intervalms: 20
               bandwidth: 50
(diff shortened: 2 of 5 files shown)


deployment/overlays/memtierd/daemonset.yaml
```diff
@@ -53,10 +57,14 @@ spec:
         hostPath:
           path: /var/run/nri
           type: Directory
-      - name: host-vol
+      - name: run-dir-vol
```
klihub, line 60: Why is this a host volume/bind mount ? To save data/avoid losing state across potential restarts ?

askervin: It exposes `memtierd.output` that enables, for instance, reading results in the `track-working-set-size` class (or use case).

It could be useful also to protect against restarts due to containing memtierd.pid's. That would enable killing memtierd processes even after nri-memtierd would have crashed/restarted. However, this is not implemented.

marquiz: Should/could we make this optional (i.e. remove from kustomize and behind a parameter in Helm)?

marquiz: Or perhaps use emptyDir volume if we want persisent storage over the lifetime of the pod

askervin: I suppose that would be a good and safe default.

deployment/overlays/memtierd/sample-configmap.yaml
marquiz, line 1: Is the change to the sample configmap somehow related?

askervin: Somewhat related... memtierd output files under /var/run/nri-memtierd/POD/CONTAINER become useful with this sample configmap change. The change introduces workload class `track-working-set-size`. Memory activity of workloads in this class can be read from their output files.

If we would disable mounting /var/run/nri-memtierd from the host, then having this example class would not make much sense as users would lack access to track output.

However, if we keep /var/run/nri-memtierd mount in the deployment files, remove from kustomize files, and make it conditional on helm charts, I think keeping this class as an example makes sense. Yet it could be added in a separate commit, too.

deployment/overlays/memtierd/daemonset.yaml
askervin: > Is there something currently in that memtierd.output which would be difficult/computationally heavy to calculate/produce on demand, because for instance some context is already lost, etc. ? If not, then an alternative would be to provide an HTTP endpoint in memtierd for querying such data.

In the track-working-set-size example memtierd.output contains full history of working set size reports, each of them being a snapshot from last access times to data in memory.

In any case, I really like the idea of having an http endpoint to serve the data. Because the output is produced by each memtierd separately, we could actually use an emptyDir to store the data from nri-memtierd container, and run our http endpoint in an unprivileged container: it can read the emptyDir and serve the data from there.

Still hesitating a bit to go this way, because right now it feels like k8s deployment would become an integral part of the implementation and architecture...

askervin: We can cache data from a reasonably big time window into RAM or storage. Between these two I'd prefer storage because speed is not an issue and the storage gives bigger time windows with lower costs.

If we'd go with caching to storage, one reasonably simple solution could be using the `fallocate()` syscall on memtierd.output files: we simply push a bigger and bigger hole in the beginning of these files (thus deallocating first blocks), and we would add an http endpoint to the nri-memtierd container to serve the rest of the data. That would keep RAM and storage usage limited, and it would allow us to keep redirecting memtierd process output directly into a file without any reader/writer threads or any other data gathering logic. Though this may sound a bit hackish, too. @klihub, @marquiz, what do you think?

klihub: I don't know if it's really worth the effort/makes sense to roll anything home-grown for file-based storage. The big question for me was whether file based storage could be avoided altogether. Since looks like it cannot be, I am happy with whatever you think works best.

marquiz (approved):
LGTM

deployment/overlays/memtierd/daemonset.yaml
askervin: I definitely agree with you that this is worth addressing in the default deployment yamls in the future. For the first release I'd consider current deployments to be good enough for trying out the nri-memtierd plugin in action. The concern regarding (host) disk space is definitely valid but unfortunately not the biggest one when thinking of memtierd in production use.

klihub (approved):

PR #125, opened 2023-10-05
helm: fix config-manager image name reference in charts defaults

fmuyassarov:
Fix incorrect naming of the config-manager container image in the Helm charts values.yaml.


marquiz (approved):
LGTM

PR #126, opened 2023-10-06
balloons: add PreferSpreadOnPhysicalCores

askervin:
- Enable using different CPU allocators in different balloon types. So far the same CPU allocator was used with the same options for all CPU allocations in the policy.
- Add new PreferSpreadOnPhysicalCores option to toggle CPU allocations packing or spreading on physical CPU cores. This option is specific to balloon type with a policy level default.
- Make existing AllocatorTopologyBalancing option specific to balloon type, too.


klihub:
@askervin Please add the missing sign-offs ?

klihub:
@askervin Thats 1 out of 2 ;-) If you could fix the other commit, too ?

askervin:
> @askervin Thats 1 out of 2 ;-) If you could fix the other commit, too ?

Uh-oh. Luckily this monkey doesn't need to type all the Shakespeare. That would take longer. Thanks @klihub!

klihub (approved):

PR #144, opened 2023-10-11
fixes: install containerd from release tarball for e2e tests.

klihub:
By default, install containerd from a release tarball, instead of expecting a compiled source tree and copying it from there.

Notes: This is still mark a draft, because I'd still need to
- plug back in support for installing from sources to allow testing against an unreleased source tree
- make similar changes for CRI-O


klihub:
Closing for #154

PR #145, opened 2023-10-11
workflow: fix the tagging docker images build from new tags

fmuyassarov:
When docker image build  workflow is triggered based on the new tag push which usually includes v as prefix, the workflow drops the v prefix. As such, when `v0.2.0` tag is pushed to GitHub, we end up with `0.2.0` tags in the docker images. This commit fixes the workflow to keep the v prefix. Currently, our deployments are tagging `v0.2.0` tags for the plugins while published images don't include the `v` and deployment are failing.

### DIFF
--- a/.github/workflows/publish-images.yaml
+++ b/.github/workflows/publish-images.yaml
@@ -47,7 +47,7 @@ jobs:
               echo "  - digging out tag from git ref $GITREF..."
               case $GITREF in
                   refs/tags/v*)
-                      tag="${GITREF#refs/tags/v}"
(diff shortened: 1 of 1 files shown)


marquiz (approved):
I think this makes sense. Just use the same vX.Y.Z version number everywhere (git tag, documentation, helm chart, image tags...). I don't know why `v` was dropped off(?)

ping @klihub @kad @askervin

kad (approved):

marquiz (approved):
Actually, even `make images` creates images with the v0.1.2 tag

PR #146, opened 2023-10-11
[release-0.2] images: fix the tagging docker images build from new tags

fmuyassarov:
When docker image build  workflow is triggered based on the new tag push which usually includes v as prefix, the workflow drops the v prefix. As such, when `v0.2.0` tag is pushed to GitHub, we end up with `0.2.0` tags in the docker images. This commit fixes the workflow to keep the v prefix. Currently, our deployments are tagging `v0.2.0` tags for the plugins while published images don't include the `v` and deployment are failing.


marquiz (approved):
LGTM

PR #150, opened 2023-10-12
Update Helm index for release v0.2.1

marquiz:


fmuyassarov (approved):

fmuyassarov:
Thanks @marquiz for this. I wanted to ask if we can merge https://github.com/containers/nri-plugins/pull/151 first. But even if we do, GitHub action won't get triggered anyways because the v0.2.1 tag is already pushed. So merging this patch.

PR #152, opened 2023-10-12
deployment: adjust Pod restartPolicy to OnFailure

fmuyassarov:
This PR fixes incorrectly used restartPolicy field. It should be at the Pod level and adjusts restartPolicy from default `Always` to `OnFailure`.


marquiz (approved):
LGTM

PR #158, opened 2023-10-16
fixes: fix image tagging scheme in publishing workflow.

klihub:
- use 'v'-prefixed image version tags
- use 'v'-prefixed versions in binaries


fmuyassarov (approved):
LGTM

PR #159, opened 2023-10-17
[release-0.2] drop DaemonSet unsupported OnFailure restartPolicy

fmuyassarov:
The DaemonSet controller currently only supports the `Always` restart policy, which is set by default. This commit addresses the issue of incorrectly set `OnFailure` restart policy in DaemonSet objects of plugins, ensuring that only the `Always` restart policy is used as intended.

Ref: https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#pod-template
Backport of https://github.com/containers/nri-plugins/pull/155


askervin (approved):
LGTM

klihub (approved):

PR #161, opened 2023-10-17
scripts: ditch versions for .rpm and .deb packages.

klihub:
Remove version string mangling for rpm and deb packages. We don't build packages, only binaries and container images now.


askervin (approved):
LGTM

PR #162, opened 2023-10-17
deployment/helm: always use DirectoryOrCreate for /var/run/nri.

klihub:
Use DirectoryOrCreate hostPath type for /var/run/nri in those Helm charts that are not set up like that yet. This should fix failures when NRI in the runtime is enabled by the config-manager init-container, so the /var/run/nri does not exist yet.


askervin (approved):
LGTM

askervin:
Thanks @klihub!

fmuyassarov (approved):
LGTM

PR #163, opened 2023-10-17
[release-0.2] always use DirectoryOrCreate for /var/run/nri

fmuyassarov:
Use DirectoryOrCreate hostPath type for /var/run/nri. This should fix failures when NRI in the runtime is enabled by the config-manager init-container, so the /var/run/nri does not exist yet.

Backport of https://github.com/containers/nri-plugins/pull/162


askervin (approved):
LGTM

klihub (approved):

PR #164, opened 2023-10-18
RFC: rework configuration using custom resources

klihub:
Note: This is work in progress. Big parts of the code is monkey puke. None of the e2e tests have been touched to switch from config files and ConfigMaps to the new configuration...

- define v1alpha1 version of a configuration API (IOW per (resource policy) plugin custom resource for configuration)
- implement new agent for monitoring configuration, delivering configuration data
- update resource manager to wait for initial configuration before starting up
- update the policy plugins to use the new CR-based configuration


klihub:
One way to give this a test using the Helm charts is to generate the images, copy and import them to the test node (podman or ctr depending on your runtime), then use a Helm override like this with the name/tag matching your build and how you imported the image...
 
```yaml
# Test overrides for TA Helm Chart. You can use an almost identical one for balloons.
#   - assume/use pre-imported plugin image
#   - never try to pull
#   - turn on full debugging
---
image:
  name: localhost/nri-resource-policy-topology-aware
  tag: v0.1.0-138-g0c402fe
  pullPolicy: Never

policyConfig:
  reservedResources:
    cpu: 750m

commonConfig:
  control:
    enable:
      - cpu
  log:
    debug:
      "*": true
    logSource: true
    klog:
      skip_headers: true
  instrumentation:
    reportPeriod: 60s
    samplingRatePerMillion: 0

hostPort: 8891

resources:
  cpu: 500m
  memory: 512Mi

nri:
  patchRuntimeConfig: false

initContainerImage:
  name: ghcr.io/containers/nri-plugins/nri-config-manager
  # If not defined Chart.AppVersion will be used
  #tag: unstable
  pullPolicy: Always

tolerations: []
```

askervin (commented):
Minor notes on installing tooling in Makefile.

test/e2e/run.sh
```diff
@@ -242,135 +243,101 @@ instantiate() { # script API
...
+                ;;
+            *)
+                error "Can't wait for plugin $plugin to start, daemonset_name not set"
+                return 0
+                ;;
+        esac
+    fi
+
+    vm-command "kubectl wait -n kube-system ds/${ds_name} --timeout=$timeout \
+             --for=jsonpath='{.status.numberAvailable}'=1" || \
```
askervin, line 309: This turned out to be a source of some flakiness. It seems that DaemonSet may have numberAvailable=1 for a little while when a nri-resource-policy plugin is launched with an illegal configuration. The policy will fail to start with an error, like in balloons test07-maxballoons with the impossible configuration, but helm-launch may report success.

But I don't know if we should grep `up and running` from resource policy logs or what... Any ideas?

askervin (commented):
An e2e test flakiness issue without a good idea how to solve it...

test/e2e/run.sh
klihub: Once possibility would be to extract the configuration generation once you'v updated it (`kubectl get topologyawareconfigs/default -o jsonpath="{.metadata.generation}"`), then `kubectl wait` for the nodes status' generation in the same object to become the same (and check if the status in there is `"Success"` or `"Failure"`

So maybe something like this...
```bash
$ G=$(kubectl get -n kube-system topologyawareconfigs/default -o jsonpath="{.metadata.generation}")
$ kubectl wait -n kube-system topologyawareconfigs/default --for=jsonpath="{.status.nodes['$(hostname)'].generation}"=$G
$ S=$(kubectl get -n kube-system topologyawareconfigs/default -o jsonpath="{.status.nodes['$(hostname)'].status}")
$ [ "$S" = "Success" ] || {
    E=$(kubectl get -n kube-system topologyawareconfigs/default -o jsonpath="{.status.nodes['$(hostname)'].errors}")
    error "configuration update failed with error \"$E\""
}
```

It is also possible that more seasoned/tougher jsonpath-ninjas are able to reduce this to a fewer set of more incomprehensible commands.

klihub: @askervin One possibility would be to extract the configuration generation once you'v updated it (`kubectl get topologyawareconfigs/default -o jsonpath="{.metadata.generation}"`), then `kubectl wait` for the nodes status' generation in the same object to become the same (and check if the status in there is `"Success"` or `"Failure"`

So maybe something like this...
```bash
$ G=$(kubectl get -n kube-system topologyawareconfigs/default -o jsonpath="{.metadata.generation}")
$ kubectl wait -n kube-system topologyawareconfigs/default --for=jsonpath="{.status.nodes['$(hostname)'].generation}"=$G
$ S=$(kubectl get -n kube-system topologyawareconfigs/default -o jsonpath="{.status.nodes['$(hostname)'].status}")
$ [ "$S" = "Success" ] || {
    E=$(kubectl get -n kube-system topologyawareconfigs/default -o jsonpath="{.status.nodes['$(hostname)'].errors}")
    error "configuration update failed with error \"$E\""
}
```

It is also possible that more seasoned/tougher jsonpath-ninjas are able to reduce this to a fewer set of more incomprehensible commands.

Makefile
```diff
@@ -434,3 +434,96 @@ site-serve: .$(DOCKER_SITE_BUILDER_IMAGE).image.stamp
...
+CLIENT_GEN_VERSION ?= v0.28.1
+CODE_GENERATOR_REPO ?= https://github.com/kubernetes/code-generator
+CODE_GENERATOR_VERSION ?= v0.28.1
+GENERATE_GROUPS ?= $(BUILD_AUX)/code-generator/generate-groups.sh
+
+# controller-gen local installation rules
+.PHONY: controller-gen
+controller-gen: $(CONTROLLER_GEN)
+$(CONTROLLER_GEN): $(LOCALBIN)
+	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
```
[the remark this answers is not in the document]
klihub: Updated.

pkg/resmgr/policy/policy.go
```diff
@@ -162,7 +166,7 @@ type Metrics interface{}
 //	for resource types (socket, die, NUMA node, etc.) and use them
 //	in policies (for instance for TA pool 'kind's)
 const (
-	// MemoryResource is resource name for memory
+	// MemoryResource is resource name fgggor memory
```
askervin, line 168: s/ggg//g

askervin (commented):
a typo

fmuyassarov (commented):
Great work @klihub . Just had a first quick round of the review and will go through it more deeply later today. 

One two more comments apart from below comments, 
1. memory-qos and memtierd templates/crds changes are not there. Is ts intentional? 
2. I'm aware that controller-gen generates all the CRDs in the `config/crd/bases`directory. However, currently, it seems like there is no practical use for them. This is because for installation via Helm, we store each CRD under the `templates/crds` directory. Therefore, the CRDs located at https://github.com/klihub/nri-plugins/tree/devel/cr-based-config/config/crd/bases appear to be redundant right?

pkg/agent/watch/watch.go
```diff
@@ -0,0 +1,214 @@
...
+
+	reopenDelay = 5 * time.Second
+)
+
+// CreateFn creates a watch for the named object.
+type CreateFn func(ctx context.Context, ns, name string) (Interface, error)
+
+// Watch is a wrapper for apimachinery watches that monitor a single object.
+// The wrapper transparently reopens the watch when it expires.
+type Watch struct {
```
askervin, line 48: This is probably balancing between matter of taste and worth the trouble, feel free to disagree and close this comment... But I was wondering if it would make sense to keep the types and consts in this file and split the [object]watch-related types to `object.go` file, that might then be quite similar to `file.go` in structure. Then there would be clearly an object watch, a file watch, and common watch source files.

klihub:
> Great work @klihub . Just had a first quick round of the review and will go through it more deeply later today.
> 
> One two more comments apart from below comments,
> 
> 1. memory-qos and memtierd templates/crds changes are not there. Is ts intentional?
> 2. I'm aware that controller-gen generates all the CRDs in the `config/crd/bases`directory. However, currently, it seems like there is no practical use for them. This is because for installation via Helm, we store each CRD under the `templates/crds` directory. Therefore, the CRDs located at https://github.com/klihub/nri-plugins/tree/devel/cr-based-config/config/crd/bases appear to be redundant right?

Those plugins are not resource management plugins and thus have not been using the same common ConfigMap-based configuration infrastructure that the resource management plugins have been. Therefore the overhaul of the config infra has no effect on those plugins.

pkg/agent/agent.go
```diff
@@ -0,0 +1,610 @@
...
+var (
+	defaultNamespace  string
+	defaultGroupLabel string
+	defaultKubeConfig string
+	defaultConfigFile string
+)
+
+func init() {
+	flag.StringVar(&defaultNamespace, "config-namespace", "kube-system",
+		"namespace for configuration CustomResources")
```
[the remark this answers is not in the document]
klihub: I think our current Helm charts in main/HEAD and the patched version of those Helm charts in this PR does not provide an explicit way to opt out from installing ConfigMaps or custom resources for configuration.

The charts install a default ConfigMap/custom resource, so the user has two ways of overriding the configuration. One is the traditional Helm way of providing custom values for Helm, since the charts are written so that the configuration is templated. The other way is to install a node- or group specific configuration that overrides the defaults.

In all cases, the agent only monitors the indicated configuration namespace (which defaults to kube-system) and nothing else. Also, just like there is no way to dynamically move a daemonset from one namespace to another, we neither need (IMO) nor provide a way to dynamically change the namespace for our configuration custom resources (which the agent monitors). The Helm charts are deliberately written so that the same namespace is used for daemonsets and the configuration resources.

deployment/helm/topology-aware/values.yaml
```diff
@@ -22,6 +35,9 @@ resources:
 nri:
   patchRuntimeConfig: false
 
+plugin:
+  enableTestAPIs: false
```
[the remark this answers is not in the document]
klihub: Yeah, probably `plugin` is not a very good choice here. I'm not too keen on `config` because we already have the two config sections `policyConfig` and `commonConfig` (which I am not too fond of either). I pulled `plugin` for the name of this section mostly because it enables (any) extra test APIs that might be compiled into the plugin binary.

Maybe this should be instead something like `testAPIs.enable: true`, or then we could combine the config bits into a single one with 3 sections `config.policy`, `config.common`, and `config.test[APIs]` and then under `config.test` we'd have `enable: false`...

pkg/agent/watch/watch.go
klihub: I think it is worth at least a try. We ended up with the current split to `{watch,file}.go` as hiding local file watching behind the same interface was one of the last things I implemented, and the current split also bothered me. I can take a look at it.

deployment/helm/topology-aware/templates/clusterrole.yaml
```diff
@@ -22,3 +22,16 @@ rules:
...
+  - topologyawareconfigs
+  - topologyawareconfigs/status
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - update
+  - patch
+  - delete
```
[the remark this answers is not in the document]
klihub: I'll check if it works without and remove them if it does.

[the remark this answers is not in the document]
klihub: While we're talking about roles and bindings, there is another thing bothering me. Shouldn't it be possible to have finer-grained RBAC rules in clustterrole ? Basically we could split the current single `apigroup` for `config.nri` into two:
- one for `${policy}configs`, allowing only `get, list, and watch` (IOW, no create, delete, update, or patch)
- another for `${policy}configs/status`, allowing only `get, list, update, and patch` (no create, delete or watch)

Then there is also the naming. Is `config.nri` and the current custom resource names fine ?

pkg/resmgr/policy/policy.go
klihub: Fixed.

pkg/agent/watch/watch.go
klihub: Also, I think that it might make sense to turn `Agent` into an `interface`...

klihub:
> 2. I'm aware that controller-gen generates all the CRDs in the `config/crd/bases` directory. However, currently, it seems like there is no practical use for them. This is because for installation via Helm, we store each CRD under the `templates/crds` directory. Therefore, the CRDs located at https://github.com/klihub/nri-plugins/tree/devel/cr-based-config/config/crd/bases appear to be redundant right?

Yes, we can control where CRDs are generated, but that needs to be a single directory and the tooling will generate all the CRDs there. So we can't really generate them directly under the Helm charts. However to keep these in sync, the Makefile rule goes through each plugin and if it finds a Helm chart with a CRD it updates that copy, too.

deployment/helm/topology-aware/templates/clusterrole.yaml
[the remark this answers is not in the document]
fmuyassarov: As discussed yesterday offline, I think it makes sense to decuple permissions(verbs) between the CRDs and their sub-resources (_status_).

deployment/helm/topology-aware/values.yaml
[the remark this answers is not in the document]
fmuyassarov: Ah I missed that `commonConfig `and `policyConfig `is already there. Maybe it is actually fine for now to keep it as it is. 

But I also kind of like the idea of `config.policy`, `config.common`, and `config.test.....`

pkg/apis/config/v1alpha1/types.go
```diff
@@ -0,0 +1,264 @@
...
+
+// TopologyAwareConfig represents the configuration for the topology-aware policy.
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +genclient
+type TopologyAwareConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	TopologyAware topologyaware.Config `json:"topologyAware"`
```
askervin, line 38: Suggesting  `json:"topology-aware"` because users see the policy name written with `-` in documentation, helm commands, deployment files and binary name, too.

askervin (commented):
\*cfg* files need fixes so that they work with `nri-resource-policy-(balloons|topology-aware) -config-file`

pkg/apis/config/v1alpha1/types.go
askervin: (Hesitating... looking at TopologyAwareConfig CRs (like topologyaware-config.yaml), having a hyphen there in "topologyAware:" would not really fit in the picture too well, if it is even allowed. Probably better keep it without hyphen after all.)

pkg/apis/config/v1alpha1/resmgr/control/config.go
```diff
@@ -0,0 +1,26 @@
...
+
+package control
+
+import (
+	"github.com/containers/nri-plugins/pkg/apis/config/v1alpha1/resmgr/control/cpu"
+)
+
+// +k8s:deepcopy-gen=true
+type Config struct {
+	Enable []string `json:"enable,omitempty"`
```
askervin, line 23: In `common.control(.cpu)` I would prefer the opposite enable/disable logic. If there is a (non-empty) `cpu:...` definition, then it should be enabled by default.

deployment/helm/topology-aware/templates/clusterrole.yaml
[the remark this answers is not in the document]
klihub: It already should be like that. I pushed such an update yesterday evening.

pkg/apis/config/v1alpha1/types.go
askervin: let's forget about this, bad idea after all, based on first feeling when trying to play with config files.

askervin:
Final remarks on the configuration in user perspective.

I would like to name our CRDs so that names would include a hint to a resource policy or  NRI. In our e2e vms `kubectl api-resources` include entries like `NetworkPolicy` and `CiliumNetworkPolicy`. Maybe `BalloonsConfig` could be `BalloonResourcePolicy` or `BalloonsNodeResourcePolicy`? (Personally I could be fine without "config" or "configuration" in the name, because pretty much everything in api-resources look configuration to me.)

I was also considering if we could somehow flatten the configuration, especially in keys/paths like `common.control.cpu`. For a user it could be more convenient to have just `cpuControl` on the top level of the configuration. However, when I look at the implementation, I very well understand why this is so convenient, especially when new resource policies are added. So this is a remark where I cannot decide: perhaps slightly inconvenient for user, but the code shines...

test/e2e/run.sh
klihub: Pushed a slightly cleaner version of the above mentioned idea.

deployment/helm/topology-aware/values.yaml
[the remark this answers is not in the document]
klihub: @fmuyassarov I updated the helm charts and the e2e tests to use the above outlined helm chart template with config.{policy,common,test}. PTAL, if this looks/feels like a better fit. I am already past the point where I could really trust my judgement. I can barely see the forest from the trees any more.

klihub:
> So this is a remark where I cannot decide: perhaps slightly inconvenient for user, but the code shines...

I suspect the only shining related to this code in its current state must be the Stanley Kubrickian kind... It sure gives me nightmares and urges me to put an axe in my own chest.

pkg/agent/watch/watch.go
klihub: @askervin Updated the watch code splitting the implementation into `{watch,object,file}.go`. I also adjusted the configuration interface slightly (provide an `Unmarshal()` instead of a `WatchFile()`).

klihub:
> *cfg* files need fixes so that they work with `nri-resource-policy-(balloons|topology-aware) -config-file`

Fixed by removing them. Those were leftovers. Config files are now simply custom resources YAML-marshalled to a file. There is no need for multiple samples with slightly different indentation/syntax.

pkg/apis/config/v1alpha1/resmgr/control/config.go
klihub: @askervin Reworked controller configuration and setup code. Eliminated configuration for explicitly enabling individual controllers by name. Instead, each controller now decides for itself whether it should be enabled or disabled and reports this back to the main controller during startup and reconfiguration. Updated CPU controller to disable itself if there are no CPU classes defined.

klihub:
> Final remarks on the configuration in user perspective.
> 
> I would like to name our CRDs so that names would include a hint to a resource policy or NRI. In our e2e vms `kubectl api-resources` include entries like `NetworkPolicy` and `CiliumNetworkPolicy`. Maybe `BalloonsConfig` could be `BalloonResourcePolicy` or `BalloonsNodeResourcePolicy`? (Personally I could be fine without "config" or "configuration" in the name, because pretty much everything in api-resources look configuration to me.)

@askervin I also disliked like the tautologic *config.config.nri CRD naming scheme. I updated things so that we now have '{balloons,topologyaware,template}policies.config.nri' as the CRDs, and '{Balloons,TopologyAware,Template}Policy' as the actual CR kinds. Additionally, I changed the '{balloons,topologyAware,template}' fields in all the custom resources to simply be 'policy'. It feels a bit more consistent now.

fmuyassarov (commented):
1. I think running `make generate` is missing and need to run at least once to generate Custom Resource Definitions (CRDs) under `config/crd/bases/`. The missing CRDs in this directory leads failure of `make update-helm-crds`, which expects CRDs to be present at `config/crd/bases/`.

    ```shell
    $ make update-helm-crds

    Updating Helm chart CRDs for topology-aware plugin...
    cp: cannot stat 'config/crd/bases/*_topologyaware*.yaml': No such file or directory
    Updating Helm chart CRDs for balloons plugin...
    cp: cannot stat 'config/crd/bases/*_balloons*.yaml': No such file or directory
    Updating Helm chart CRDs for template plugin...
    cp: cannot stat 'config/crd/bases/*_template*.yaml': No such file or directory
    No generated CRD found for memory-qos plugin...
    No generated CRD found for memtierd plugin...
    No generated CRD found for sgx-epc plugin...
    ```

2. The generated CRD name in the Helm directory has diverged from its initial version. The `config.nri_topologyawarepolicies.yaml` file is added to `deployment/helm/topology-aware/crds`, deviating from the already existing `noderesourcetopology.yaml`. I think first `make generate` needs to be run to get the latest CRD generated and then fix the naming of the CRD or remove the old `noderesourcetopology.yaml` and replace it with the new one `config.nri_topologyawarepolicies.yaml` ?
3. `build-aux` directory would be good to add in `.gitignore` because generated binaries are now being tracked by the git.
4. https://github.com/klihub/nri-plugins/blob/devel/cr-based-config/sample-configs/blockio.cfg can be removed perhaps? I'm not sure if it has any use.
5. We need to add template plugin reference in the root [README](https://github.com/klihub/nri-plugins/blob/devel/cr-based-config/README.md) 
6. TA plugin helm chart is missing its CRD at https://github.com/klihub/nri-plugins/tree/devel/cr-based-config/deployment/helm/topology-aware/crds.

Makefile
```diff
@@ -434,3 +405,96 @@ site-serve: .$(DOCKER_SITE_BUILDER_IMAGE).image.stamp
...
+.PHONY: generate-type
+generate-types: controller-gen
+	$(CONTROLLER_GEN) object:headerFile="./docs/license-header.go.txt" paths="./pkg/apis/..."
+
+# client generation rules 
+.PHONY: generate-clients
+generate-clients: $(GENERATE_GROUPS)
+	$(GENERATE_GROUPS) client \
+	    github.com/containers/nri-plugins/pkg/generated \
+	    github.com/containers/nri-plugins/pkg/apis \
```
fmuyassarov, line 474: we don't have these directories here I guess? At least, `apis/` is under `pkg/`. Should we fix it here?

klihub:
> 1. I think running `make generate` is missing and need to run at least once to generate Custom Resource Definitions (CRDs) under `config/crd/bases/`. The missing CRDs in this directory leads to failure of `make update-helm-crds`, which expects CRDs to be present at `config/crd/bases/`.
>    ```shell
>    $ make update-helm-crds
>    
>    Updating Helm chart CRDs for topology-aware plugin...
>    cp: cannot stat 'config/crd/bases/*_topologyaware*.yaml': No such file or directory
>    Updating Helm chart CRDs for balloons plugin...
>    cp: cannot stat 'config/crd/bases/*_balloons*.yaml': No such file or directory
>    Updating Helm chart CRDs for template plugin...
>    cp: cannot stat 'config/crd/bases/*_template*.yaml': No such file or directory
>    No generated CRD found for memory-qos plugin...
>    No generated CRD found for memtierd plugin...
>    No generated CRD found for sgx-epc plugin...
>    ```

Thanks. Forgot to commit the new CRD's in `config/crd/bases` after changing the CRD group name.

> 2. The generated CRD name in the Helm directory has diverged from its initial version. The `config.nri_topologyawarepolicies.yaml` file is added to `deployment/helm/topology-aware/crds`, deviating from the already existing `noderesourcetopology.yaml`. I think first `make generate` needs to be run to get the latest CRD generated and then fix the naming of the CRD or remove the old `noderesourcetopology.yaml` and replace it with the new one `config.nri_topologyawarepolicies.yaml` ?

Yes, you only need to run `make generate` which already should take care of updating everything. You don't need to run the helm CRD updating rule separately. You can if you want, though... for you it was failing since I forgot to add+commit the renamed CRD.
```
klitkey1-mobl1 cr-based-config $ grep generate: Makefile
generate: generate-go generate-types generate-manifests generate-clients update-helm-crds
``` 

> 3. `build-aux` directory would be good to add in `.gitignore` because generated binaries are now being tracked by the git.

Added.

> 4. https://github.com/klihub/nri-plugins/blob/devel/cr-based-config/sample-configs/blockio.cfg can be removed perhaps? I'm not sure if it has any use.

Removed.

> 5. We need to add template plugin reference in the root [README](https://github.com/klihub/nri-plugins/blob/devel/cr-based-config/README.md)

Hmm, yes. Maybe we should link this in to the documentation. However, I don't get a documentation build error about this. Do you really get one ?

> 6. TA plugin helm chart is missing its CRD at https://github.com/klihub/nri-plugins/tree/devel/cr-based-config/deployment/helm/topology-aware/crds.

Thanks. I forgot to add+commit that one too, across CRD renames. Added.

fmuyassarov:
> Hmm, yes. Maybe we should link this in to the documentation. However, I don't get a documentation build error about this. Do you really get one ?

It is true that there isn't an error but it is missing though. It is not a bug or a bid deal and just would be good to follow the same pattern as other plugins. But I understand that it is a hassle to rebase since this is giant PR. So, I'm absolutely fine to postpone it to be done within some other future PR.

Makefile
fmuyassarov: Please ignore.

fmuyassarov (commented):
This looks good to me. I tried all there plugins with Helm and so far all have started fine.
One question though, should not we be sticker when it comes to RBAC? Instead of using clusterRole/clusterRoleBinding perhaps we should limit the permissions to a particular namespace via role/roleBinding where the plugin is running? I understand that there are mix of resources where some are cluster-scope and some namespaced. And I guess we can decouple namespaced ones into role? For instance, topologyawarepolicies?

klihub:
> This looks good to me. I tried all there plugins with Helm and so far all have started fine. One question though, should not we be sticker when it comes to RBAC? Instead of using clusterRole/clusterRoleBinding perhaps we should limit the permissions to a particular namespace via role/roleBinding where the plugin is running? I understand that there are mix of resources where some are cluster-scope and some namespaced. And I guess we can decouple namespaced ones into role? For instance, topologyawarepolicies?

Good question. AFAICT, the only things we need to be able to access outside of our own namespace are: read-write (create/update) access to NRT's and read-only (list/get) access to nodes to get the configuration grouping label for the node we're running on.

klihub:
> > Hmm, yes. Maybe we should link this in to the documentation. However, I don't get a documentation build error about this. Do you really get one ?
> 
> It is true that there isn't an error but it is missing though. It is not a bug or a big deal and just would be good to follow the same pattern as other plugins. But I understand that it is a hassle to rebase since this is giant PR. So, I'm absolutely fine to postpone it to be done within some other future PR.

I also think it should be linked to in the documentation. I actually already added and force-pushed it among the other fixes to your comments. I was just curious if you get a doc build error for it...

fmuyassarov:
> Good question. AFAICT, the only things we need to be able to access outside of our own namespace are: read-write (create/update) access to NRT's and read-only (list/get) access to nodes to get the configuration grouping label for the node we're running on.

**BEFORE**:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: nri-resource-policy-topology-aware
  labels:
    {{- include "nri-plugin.labels" . | nindent 4 }}
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  verbs:
  - get
  - watch
- apiGroups:
  - topology.node.k8s.io
  resources:
  - noderesourcetopologies
  verbs:
  - create
  - get
  - list
  - update
  - delete
- apiGroups:
  - config.nri
  resources:
  - topologyawarepolicies
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - config.nri
  resources:
  - topologyawarepolicies/status
  verbs:
  - get
  - list
  - update
  - patch
```
**AFTER**: (Role and ClusterRole)
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: nri-resource-policy-topology-aware
  labels:
    {{- include "nri-plugin.labels" . | nindent 4 }}
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  verbs:
  - get
  - watch
- apiGroups:
  - topology.node.k8s.io
  resources:
  - noderesourcetopologies
  verbs:
  - create
  - get
  - list
  - update
  - delete
```

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: nri-resource-policy-topology-aware
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "nri-plugin.labels" . | nindent 4 }}
rules:
- apiGroups:
  - config.nri
  resources:
  - topologyawarepolicies
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - config.nri
  resources:
  - topologyawarepolicies/status
  verbs:
  - get
  - list
  - update
  - patch
```

fmuyassarov:
> t

I'm not sure I can find it. 
Screenshot from 2023-11-21 14-33-19

cmd/agent-test/main.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/containers/nri-plugins/pkg/agent"
+	cfgapi "github.com/containers/nri-plugins/pkg/apis/config/v1alpha1"
+	logger "github.com/containers/nri-plugins/pkg/log"
+)
+
+var (
+	log        = logger.Get("agent-test")
```
fmuyassarov, line 15: I'm wondering if it would be better to keep all the tests under root `test/` directory?

klihub: Sorry, I pushed this accidentally. It was not supposed to be part of the PR at all.

klihub:
> > t
> 
> I'm not sure I can find it. Screenshot from 2023-11-21 14-33-19

You mean the template plugin ? It only has a Helm chart description now. I can add a short description for it and link it among the rest of the plugins.

klihub:
> > t
> 
> I'm not sure I can find it. Screenshot from 2023-11-21 14-33-19

Added a very short (single-paragraph) description of the template policy and linked it among the others.

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
marquiz, line 1: Is there a reason to have these deeply nested packages under `pkg/apis/config/v1alpha1`? We could have all the types defined in the same directory, we could have separate files, though, like `types_balloons.go`, `types_instrumentation.go` etc.

This is merely a code structuring question and doesn't (afaiu) affect the API. Can be addressed at a later time if desired.

pkg/apis/config/v1alpha1/types.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// AnyConfig provides access to metadata of configuration resources.
+// All top level configuration types must implement this interface.
+// Among other things, the agent uses it to filter out unnecessary
+// update notifications.
+// +kubebuilder:object:generate=false
+type AnyConfig interface {
+	GetObjectMeta() *metav1.ObjectMeta
+}
```
marquiz, line 185: You shouldn't need this. E.g. k8s apimachinery (`k8s.io/apimachinery/pkg/apis/meta/v1`) provides helpers - you should be able to use the generic `Object` type (`metav1.Object`) from there to access object meta fields of any api object. The k8s `client-go` provides some similar functionalities iirc.

```suggestion
```

pkg/apis/config/v1alpha1/types.go
```diff
@@ -0,0 +1,276 @@
...
+
+type patchConfig struct {
+	Status patchStatus `json:"status,omitempty"`
+}
+
+type patchStatus struct {
+	Nodes map[string]*NodeStatus `json:"nodes,omitempty"`
+}
+
+func NodeStatusPatch(node string, status *NodeStatus) ([]byte, types.PatchType, error) {
```
marquiz, line 161: I'd suggest to keep the `types.go` strictly for typedefs. Put funcs somewhere else, preferably in a separate package (really separating the API from implementation details).

Again, this is a code-structuring thing that can be addressed later.

pkg/apis/config/v1alpha1/types.go
```diff
@@ -0,0 +1,276 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +genclient
+type TopologyAwarePolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Policy topologyaware.Config `json:"policy"`
+	// +optional
+	Common CommonConfig `json:"common,omitempty"`
```
marquiz, line 42: Two comments/questions:

1. Maybe I'm just too stuck with the common spec vs. status pattern in the APIs but the immediate question rose that why not follow it? 😇 Spec specifies the desired state (i.e. complete configuration) and status the tells where we are in getting there. WDYT?

2. Another note is that the `common` field just looks strange from the users point of view (imho). Common what(?) Policy as a separate field makes more sense. I'd suggest putting the common stuff just as fields on the top level

```golang
type TopologyAwarePolicy struct {
	metav1.TypeMeta
	metav1.ObjectMeta

	Policy topologyaware.Config
	// Common stuff for the resource manager
	Control control.Config
	Log log.Config
	Instrumentation instrumentation.Config
}
```

Nothing prevents having the same getter func/method (`CommonConfig()`) as before for the common config bits, but it doesn't need to be visible in the user-facing API.

With these changes the the CR yaml would look smth like:

```yaml
apiVersion: config.nri/v1alpha1
kind: TopologyAwarePolicy
metadata:
  name: default
spec:
  policy:
    pinCPU: true
    pinMemory: true
    reservedResources:
      cpu: 750m
  log:
    klog:
      skip_headers: true
  instrumentation:
    reportPeriod: 60s
```

Thoughts?

marquiz (commented):
Wow @klihub , absolutely huge PR 😅 Generally looks really good. Some comments below.

pkg/apis/config/v1alpha1/types.go
klihub: That (providing a uniform access API via common mandatory interfaces as opposed to a common mandatory struct type) is a very good point. I completely missed/overlooked that possibility. I need to think about it and probably give it a try.

pkg/apis/config/v1alpha1/types.go
klihub: Generally I agree, but I still prefer to keep code with the types when the implementation details are tightly coupled with the actual (structure of the) configuration data. And this is such a case.

pkg/apis/config/v1alpha1/types.go
klihub: > 1. Maybe I'm just too stuck with the common spec vs. status pattern in the APIs but the immediate question rose that why not follow it? 😇 Spec specifies the desired state (i.e. complete configuration) and status the tells where we are in getting there. WDYT?

Uhmmm, isn't that what we have here ? At least that's what I was aiming for. The /status subresource describes the status of applying the configuration described by the resource for all nodes that it is applicable to. No ?

pkg/apis/config/v1alpha1/types.go
marquiz: A sidenote to this that I think you need to inform the kubebuilder to implement the interface with
```
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
in the type definition

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
klihub: Whichever way we look at it, whenever there is a possibility for naming conflicts you need to disambiguate. You can do that by encoding scoping in the type (base)name, or you can use scoping provided by the language, which in golang is packages. I opted here for the latter.

That said, I did flatten the (other )control-related configuration, which I hated a lot (partially for other reasons). After flattening now I dislike it very much (partially for other reasons) and I'm not sure if that was really an improvement.

Generally speaking I really don't know if

```
...
    Xyzzy  XyzzyConfig
    Foobar FoobarConfig
...
```

is universally always/any better than

```
...
    Xyzzy  xyzzy.Config
    Foobar foobar.Config
...
```

I suspect it very much depends both on the actual case and the reader.

pkg/apis/config/v1alpha1/types.go
marquiz: > Uhmmm, isn't that what we have here ?

Not really, all the "desired state" stuff are on the main level, same as status, i.e. we currently have like

```yaml
apiVersion: config.nri/v1alpha1
kind: TopologyAwarePolicy
metadata:
  name: default
policy:
  pinCPU: true
  ...
log:
  ...
status:
 ...
```

What I was suggesting have a separate `spec` field on the main level, under which all the "desired state" stuff would go, like


```yaml
apiVersion: config.nri/v1alpha1
kind: TopologyAwarePolicy
metadata:
  name: default
spec:
  policy:
    pinCPU: true
    ...
  log:
    ...
status:
 ...
```

pkg/apis/config/v1alpha1/types.go
marquiz: We can debate over this at some other time 😅 Even in this case I think it would be good to keep the api package cleaner. In general keeping all implementation somewhere else e.g. prevents the automatic code generation from accidentally picking up types that are not part of the API etc.

But as said, this can be addressed later if desired. In any case, I'd suggest to move the helper/accessor funcs in a separate file to keep types.go tidy. WDYT?

sample-configs/balloons-config.yaml
```diff
@@ -0,0 +1,246 @@
...
+#    source: true
+#    klog:
+#       skip_headers: true
+---
+# Configuration for node 'custom-node0'.
+# Only used for custom-node0.
+apiVersion: config.nri/v1alpha1
+kind: BalloonsPolicy
+metadata:
+  name: node.custom-node0
```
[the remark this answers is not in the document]
klihub: I think @askervin then needs to come up with an alternative way of providing documentation about the subtleties of balloons in a way that is both comprehensible and maintainable.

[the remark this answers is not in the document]
askervin: ...which does not sound bad at all for users and maintainers. 😄

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
marquiz: I think there is value (and I think it's a rather common practice) to have all needed types defined in the api package itself (plus referencing some other api packages if needed).

I don't know if 
```golang
...
    Xyzzy  XyzzyConfig
    Foobar FoobarConfig
...
```

is that bad.

But again, this is merely discussion and not blocking the PR.

sample-configs/balloons-config.yaml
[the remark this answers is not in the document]
marquiz: > I think @askervin then needs to come up with an alternative way of providing documentation about the subtleties of balloons in a way that is both comprehensible and maintainable.

I think that could be handled directly in the documentation. OR then alternatively have another balloons sample config (with multiple objects) that sheds lights on the subtleties but is much shorter and does not document every single option in every single object – just highlights the bits that are crucial for the multi-config setup. WDYT?

Makefile
```diff
@@ -434,3 +405,96 @@ site-serve: .$(DOCKER_SITE_BUILDER_IMAGE).image.stamp
...
+	cd $(dir $@) && \
+	for d in ./cmd/*; do go build $$d; done
+
+# CRD generation rules
+.PHONY: generate-manifests
+generate-manifests: controller-gen
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd paths="./pkg/apis/..." output:crd:artifacts:config=$(CRD_BASE_DIR)
+
+# code generation rules for CRD types (DeepCopy, DeepCopyInto, etc.)
+.PHONY: generate-type
```
fmuyassarov, line 465:
```suggestion
.PHONY: generate-types
```

fmuyassarov (changes requested):

Makefile
```diff
@@ -434,3 +405,96 @@ site-serve: .$(DOCKER_SITE_BUILDER_IMAGE).image.stamp
...
+CLIENT_GEN ?= $(LOCALBIN)/client-gen
+
+GIT_CLONE    := git clone
+GIT_CHECKOUT := git checkout
+
+# External generation tool versions
+CONTROLLER_TOOLS_VERSION ?= v0.12.0
+CLIENT_GEN_VERSION ?= v0.28.1
+CODE_GENERATOR_REPO ?= https://github.com/kubernetes/code-generator
+CODE_GENERATOR_VERSION ?= v0.28.1
```
fmuyassarov, line 434: This branch name is the one that we create when cloning the repo. Should we rather do the cloning based on the upstream `release-1.28` branch?

`$(GIT_CLONE) $(CODE_GENERATOR_REPO) -b $(CODE_GENERATOR_VERSION) origin/release-1.28?`

The reason is, if we are following particular k8s version, perhaps we should stick to that version of the code generators here too to avoid any surprises that may happen on the master/main branch?

pkg/apis/config/v1alpha1/types.go
klihub: But I will split all but the types to separate files in the package. I agree that it is cleaner that way.

Makefile
```diff
@@ -434,3 +405,96 @@ site-serve: .$(DOCKER_SITE_BUILDER_IMAGE).image.stamp
...
+	    "config:v1alpha1" --output-base=. \
+	    --go-header-file=./docs/license-header.go.txt && \
+	rm -fr pkg/generated && \
+	mv github.com/containers/nri-plugins/pkg/generated pkg && \
+	rm -fr github.com
+
+# golang code generation rules
+generate-go:
+	$(Q)$(GO_CMD) generate ./...
+
```
fmuyassarov, line 484: what's the use of this target? I could not find a single go file that is using go code generator markers (`//go:generate`)?

klihub: klitkey1-mobl1 cr-based-config $ tail -1 pkg/apis/config/v1alpha1/log/klogcontrol/doc.go
//go:generate go run gen-config.go klogcontrol Config config.go

pkg/apis/config/v1alpha1/types.go
marquiz: Oh, I forgot about another possibility is to make the common part "inline", similar to e.g. ObjectMeta, e.g.
```suggestion
	Policy topologyaware.Config `json:"policy"`
	// +optional
	Common CommonConfig `json:",inline"
```

Optionally maybe renaming the type to `CommonPolicyConfig` or `CommonResourcePolicyConfig`, with a look forward to some common config patterns of other type of plugins...

Makefile
klihub: Based on my own and @marquiz's past experiment with controller-gen's stability, I deliberately want to version-lock `code-generator` to an exact version which is known to work for us. It might be better/more API-stable now and I am willing to give it a try... once we have figured out the exact details of how to switch from using the deprecated `generate-groups.sh` to `kbe_codegen.sh`. I took a look at that initially during this work, but it was very much not a drop-in replacement, so I decided against it as this particular problem is IMO extremely low on our TODO list at the moment.

Makefile
```diff
@@ -434,3 +405,96 @@ site-serve: .$(DOCKER_SITE_BUILDER_IMAGE).image.stamp
...
+$(LOCALBIN):
+	mkdir -p $(LOCALBIN)
+
+## External generation tool binaries
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+CLIENT_GEN ?= $(LOCALBIN)/client-gen
+
+GIT_CLONE    := git clone
+GIT_CHECKOUT := git checkout
```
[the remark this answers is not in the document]
klihub: Removed.

Makefile
klihub: Fixed.

pkg/apis/config/v1alpha1/types.go
klihub: Updated, splitting out any extra code from the type definitions.

pkg/apis/config/v1alpha1/types.go
klihub: Removed and switched to using `metav1.ObjectMeta` and `metav1.ObjectMetaAccessor` type assertions.

sample-configs/balloons-config.yaml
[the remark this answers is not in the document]
klihub: Nuked the node- and group-specific config resources. Added comment before `metadata.name` describing how to define node- and group-specific configuration resources. Ditto for `topologyaware-config.yaml`.

klihub:
> Wow @klihub , absolutely huge PR 😅 Generally looks really good. Some comments below.

It is indeed borderline unmanageably huge and I am truly sorry for that. 🙈 😞 Yet, still to date I haven't figured out how we could have done this in a less intrusive way, considering how widespread its blast radius is.

Makefile
fmuyassarov: @klihub is that really a go code relying on go generator? Isn't that line you got from grepping is just a comment ?

fmuyassarov: Ah it is needed actually. Let's keep it.

Makefile
fmuyassarov: > Based on my own and @marquiz's past experiment with controller-gen's stability, I deliberately want to version-lock `code-generator` to an exact version which is known to work for us.

Isn't it the same with the command I suggested above? It is using [release-1.28](https://github.com/kubernetes/code-generator/tree/release-1.28) release branch of the controller-gen right?

fmuyassarov (approved):
There is one comment left about code-generator version but this is a nit and we can iterate over and agree how we want to continue using it. Otherwise this is pretty good looking work already. Thank you @klihub

Makefile
klihub: I don't think `release-1.28` would be version-/sha1-locked. If/when a `1.28.2` or any other `1.28.x` gets tagged it will point to the tip of `release-1.28`. IOW, `release-1.28` will receive as yet unknown commits.

pkg/apis/config/v1alpha1/types.go
```diff
@@ -0,0 +1,148 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   TopologyAwarePolicySpec `json:"spec"`
+	Status ConfigStatus            `json:"status,omitempty"`
+}
+
+// TopologyAwarePolicySpec describes a topology-aware policy.
+type TopologyAwarePolicySpec struct {
+	Policy topologyaware.Config `json:"policy"`
```
marquiz, line 42: Now that `Common` was eliminated gave another thought to the `Policy` field and wonder if we should eliminate it, too

```suggestion
	topologyaware.Config `json:",inline"`
```

It would move all the policy config fields to the top (spec) level.. As an example, it would make the config yaml look like this

```yaml
apiVersion: config.nri/v1alpha1
kind: TopologyAwarePolicy
metadata:
  name: default
  namespace: kube-system
spec:
  pinCPU: true
  pinMemory: true
  reservedPoolNamespaces:
    - special-namespace
    - extra-reserved-namespace
  reservedResources:
    cpu: 750m
  log:
    klog:
      skip_headers: true
  instrumentation:
    reportPeriod: 60s
```

I'm a bit indeterminate on this one. as in my some earlier comment I said that the `policy` field makes sense. But now that I look at it, it doesn't necessarily so much. The object type name is already TopologyAware**Policy** so the `policy` field feels unnecessary and tautological. WDYT?

fmuyassarov: > I'm a bit indeterminate on this one. as in my some earlier comment I said that the `policy` field makes sense. But now that I look at it, it doesn't necessarily so much. The object type name is already TopologyAware**Policy** so the `policy` field feels unnecessary and tautological. WDYT?

+1

klihub: Changed all the former `Policy` fields inlined, dropping out `policy` from the configuration data on user-visible interfaces.

marquiz (approved):
Thank you @klihub for this huge effort 👏 I think the CRD API now looks nice and clean. And everything else (if we find something) we can easily change later without touching the UI (I hope) 😇

LGTM

PR #166, opened 2023-10-18
[gh-pages] Helm: add artifacthub-repo.yml to get publisher verification.

fmuyassarov:
Add artifacthub-repo.yml to get "verified publisher" label in artifacthub.io for our charts.


klihub (approved):

klihub:
I think this should be fine. I was worried for a moment that the next automatic update of/push to `gh-pages` would nuke this, but it looks like that should not happen.

PR #168, opened 2023-10-19
build/docs: make site-build work again from git worktrees.

klihub:
Fix up careless commit a5f7ddd0a134 which broke `make site-build`... when run from a git worktree.


fmuyassarov (approved):
LGTM

PR #171, opened 2023-10-24
.github: add workflow for rejecting PRs that introduce whitespace errors.

klihub:
This PR adds a workflow to check and reject PRs that introduce whitespace errors. The method used to check for whitespace errors is a minimalistic `git diff-index --check` against the base ref of the PR.

### DIFF
--- a/.github/workflows/project-checks.yaml
+++ b/.github/workflows/project-checks.yaml
@@ -30,6 +30,15 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
+      - name: Check PR for whitespace errors
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          if ! git diff-index --check origin/${{ github.base_ref }} --; then
+              echo "This PR would introduce the errors shown above."
+              echo "Please fix them and update the PR."
+              exit 1
+          fi
+
       - name: Verify dependencies
         run: |
           make verify-godeps


.githooks/run-hooks
```diff
@@ -0,0 +1,31 @@
...
+
+exec 1>&2
+
+for h in "$dotd"/???*; do
+    case "$h" in
+        *~|*.swp)
+            continue
+            ;;
+        [0-9]*-*)
+            ;;
```
marquiz, line 15: What is the significance of this?

.githooks/run-hooks
```diff
@@ -0,0 +1,31 @@
...
+    $h $@
+    status=$?
+    if [ $status != 0 ]; then
+        exit $status
+    fi
+done
+
+if [ -x "$orig" ]; then
+    exec $orig
+fi
```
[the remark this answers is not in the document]
klihub: Yes. If the user has an enabled git hook in the stock location, IMO we should run it.

.github/workflows/project-checks.yaml
```diff
@@ -27,6 +27,15 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
+      - name: Git-check PR for conflict markers and trailing whitespace errors.
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          git fetch -q origin ${{ github.base_ref }}
```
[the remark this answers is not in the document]
klihub: Hmm, true. Fixed.

.githooks/run-hooks
klihub: Hmm, in its current form it's merely documenting/suggesting how the hook-lets should be named with/for explicit ordering. But perhaps we should explicitly ignore the rest... which I suspect might have been my original intention at some point.

Added a default case for that effect and updated the PR.

.githooks/run-hooks
```diff
@@ -0,0 +1,34 @@
...
+orig="${0%/*}/../.git/hooks/$hook"
+
+exec 1>&2
+
+for h in "$dotd"/???*; do
+    case "$h" in
+        *~|*.swp)
+            continue
+            ;;
+        [0-9]*-*)
```
marquiz, line 14: Maybe we should document somehow (more clearly, somewhere), that the hooks in the drop-in dir have strict naming requirements (otherwise they are silently ignored). WDYT?

klihub:
Okay, I ditched all the git hooks, their installation and related additions, leaving only a minimal whitespace error check in the workflows.

fmuyassarov (approved):

PR #157, opened 2023-10-16
helm: add an option to set tolerations

fmuyassarov:
In some clusters, some of the Nodes might have taints which may avoid NRI plugins DaemonSet Pods not to get scheduled on those Nodes. Having an option to pass custom tolerations to plugins DaemonSet object would help to ensure that Pods get scheduled on all the desired Nodes.

PR #178, opened 2023-10-25
[release-0.2] helm: add an option to set tolerations (backport of #157)

klihub:
In some clusters, some of the Nodes might have taints which may avoid NRI plugins DaemonSet Pods not to get scheduled on those Nodes. Having an option to pass custom tolerations to plugins DaemonSet object would help to ensure that Pods get scheduled on all the desired Nodes.


klihub:
@fmuyassarov @marquiz Should we cherry-pick this, too ?

fmuyassarov (approved):
LGTM
I think it is good to have this too.

PR #180, opened 2023-10-25
config-manager: don't crash if no runtime is found running.

klihub:


marquiz (approved):
Not pretty but LGTM 😅 I wouldn't necessarily spend too much time on this

fmuyassarov (approved):
LGTM
Thanks @klihub . I will spend some time to really understand why is the real culprit in the CRI-O.
But for now, I also agree that this is good.

PR #188, opened 2023-11-02
balloons: use new config more consistently, fix incorrect test config keys.

klihub:
Don't delay updating stored configuration when reconfiguring the policy. Some functions in the call-tree (for instance `newBalloon()`), use the stored options, so they could end up using old data otherwise.

Additionally, fix a few incorrect keys (`AllocationPriority` should be `AllocatorPriority`) in the balloons tests.

@askervin I bumped into this while fixing failing tests for #164 and I realized it is present also in current main/HEAD.   
This patch might not be a fully proper fix (I didn't try to check all the implications). This change might alter behavior on the reconfiguration error code paths, albeit that should be fixed (at least in principle) by the subsequent successful (re)application of the previous configuration, triggered externally by resmgr... Still, I think we probably prefer the more correct behavior on the normal/errorfree code paths.


askervin:
Great, thanks for spotting and fixing this @klihub!

marquiz (approved):
Oh, #190 went in already 🙄 Anyway, this one looks good to me but I'll let @askervin to make the call

askervin (approved):
LGTM

PR #189, opened 2023-11-03
[release-0.2] balloons: use new config more consistently.

klihub:
Don't delay updating stored configuration when reconfiguring the policy. Some functions in the call-tree (for instance `newBalloon()`), use the stored options, so they could end up using old data otherwise.

This PR is a cherry-pick of #188.


askervin (approved):
LGTM

PR #193, opened 2023-11-14
test/e2e: fix vagrant usage, make cilium version configurable.

klihub:
This PR fixes a number of problems related to vagrant setup and usage in the e2e tests. In particular:

- use Vagrantfile.in as a vagrant ERB template instead of a pre-created Vagrantfile (which latest vagrant init rejects)
- check vagrant status and fail tests if vagrant (init, provision, or up) commands fail
- try using ssh directly, falling back to vagrant ssh only if this fails
- allow default (latest) cilium version to be overridden for tests

These commits have been split out from #164 as a self-contained and largely unrelated set that can be reviewed and merged separately.

### DIFF
--- a/test/e2e/files/Makefile.in
+++ b/test/e2e/files/Makefile.in
@@ -1,3 +1,7 @@
+ifneq ($(V),1)
+.SILENT:
+endif
+
 all: up
 
 install: .plugins.installed.stamp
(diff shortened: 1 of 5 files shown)


klihub:
@askervin PTAL if you have time. If we get this merged, I'll rebase #164 on top of it (now the same commits are at the bottom of the commit stack there).

fmuyassarov (approved):
LGTM

askervin (approved):
LGTM

PR #195, opened 2023-11-16
RFC WIP: experimenting with building debug binaries and images

askervin:
Builds debug versions of binaries and images with make DEBUG=1. When a debug version of an NRI resource policy image is running, you can attach debugger to it with

`kubectl exec -n kube-system -it nri-resource-policy-... -- dlv attach 1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,11 @@ LDFLAGS    = \
              -X=github.com/containers/nri-plugins/pkg/version.Build=$(BUILD_BUILDID) \
              -B 0x$(RANDOM_ID)"
 
+ifeq ($(DEBUG),1)
+    GCFLAGS ?= -gcflags "all=-N -l"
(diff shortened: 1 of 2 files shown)


askervin:
Polling opinions on this debug image approach. In the first (proof-of-concept) draft version of this PR debug image building is enabled only on the balloons policy image. What do you think, would this be worth adding to other policies, too? Or would this kind of debug images be something that shouldn't be built at all or not with the same names or Dockerfiles as production images?

klihub:
This is akin to STATIC=1, so I'd make it applicable to all images. Can't we twist this so, that DEBUG=1 will build an image which is either named or tagged differently ? So you could always swap in a DEBUG image (if it was built) to a helm chart for debugging ?

klihub:
Closing this as ancient draft.

PR #196, opened 2023-11-17
helm: introduce affinity to select specific nodes for the DaemonSet

fmuyassarov:
This patch allows users to define custom affinity rules in `values.yaml,` specifying preferred nodes for plugin DaemonSet Pods. Follows Kubernetes structure for affinity fields.


klihub (commented):
This LGTM. I just wonder if we should also add support for simple label-based `nodeSelector`s ?

fmuyassarov:
> This LGTM. I just wonder if we should also add support for simple label-based `nodeSelector`s ?

Done.

klihub (approved):

PR #202, opened 2023-11-27
topology: support more devices under /sys/devices

askervin:
Extra device support removes limitations on referring to some devices under /sys/devices in cases where referenced devices have notion of local CPUs yet the file has other name than local_cpulist. This supports e2e tests with device locality because added devices have non-trivial set of local CPUs in e2e vms.

### DIFF
--- a/pkg/topology/topology.go
+++ b/pkg/topology/topology.go
@@ -106,8 +106,16 @@ func getTopologyHint(sysFSPath string) (*Hint, error) {
 
 	hint := Hint{Provider: sysFSPath}
 	fileMap := map[string]*string{
+		// match /sys/devices/system/node/node0/cpulist
+		"cpulist": &hint.CPUs,
+		// match /sys/devices/pci0000:00/pci_bus/0000:00/cpulistaffinity
+		"cpulistaffinity": &hint.CPUs,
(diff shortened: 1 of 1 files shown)


pkg/topology/topology.go
```diff
@@ -106,8 +106,16 @@ func getTopologyHint(sysFSPath string) (*Hint, error) {
 
 	hint := Hint{Provider: sysFSPath}
 	fileMap := map[string]*string{
-		"local_cpulist": &hint.CPUs,
-		"numa_node":     &hint.NUMAs,
+		// match /sys/devices/system/node/node0/cpulist
+		"cpulist":         &hint.CPUs,
+		// match /sys/devices/pci0000:00/pci_bus/0000:00/cpulistaffinity
+		"cpulistaffinity": &hint.CPUs,
```
kad, line 112: I'm not 100% sure about this one. In subfunction entries of PCI devices, the topology func should go up one layer in tree via `device` and should see `local_cpulist` entry. Have you seen scenarios where this would be more reliable compared to root function of the pci device?

kad (approved):
Generally I'm ok with those changes. I think we can merge it now, and observe, if some corner cases would pop-up.

PR #206, opened 2023-11-29
sysfs, cpuallocator: topology discovery fixes and improvements.

klihub:
Notes: this patch series is stacked on top of #294.

This patch series improves hardware topology discovery and CPU allocation. The improvements include:
- fix cache discovery and enable it by default
- add support for querying CPUs by last-level cache
- add support for discovering CPU cluster information
- add support for querying CPUs by die and cluster
- filtered/logical cluster querying: all 'hyperthread cluster's reported as a single cluster
- clean up detection and interfaces for online, isolated and offline CPU detection
- add support for querying possible and present CPUs
- implement cluster-aware allocation in the CPU allocator.


kad (approved):
LGTM

askervin (approved):
LGTM

pkg/sysfs/system.go
```diff
@@ -732,6 +736,118 @@ func (c *cpu) SetFrequencyLimits(min, max uint64) error {
...
+
+	for idx := len(c.caches) - 1; idx >= 0; idx-- {
+		cch := c.caches[idx]
+		if level == -1 {
+			level = cch.level
+		}
+		if level != cch.level {
+			break
+		}
+		caches = append(caches, cch)
```
marquiz, line 921: This loop somehow looks haphazard. We always assume `c.caches` to be sorted? Is c.caches[idx].level sometimes `-1`?

pkg/sysfs/system.go
```diff
@@ -635,13 +669,38 @@ func (sys *system) discoverCPUs() error {
...
 
+	sys.coreKindCPUs = make(map[CoreKind]idset.IDSet)
+	for kind, entry := range coreKindCPUPath {
+		cpus := idset.NewIDSet()
+		_, err = readSysfsEntry(sys.path, entry, &cpus)
+		if err != nil {
+			sys.Error("failed to get set of %s cpus: %v", kind, err)
+			if kind == PerformanceCore {
+				cpus = sys.onlineCPUs.Clone()
+			}
```
marquiz, line 683: minor nit: in case of any error, should we bail out clearly with smth like

```golang
    sys.coreKindCPUs = map[CoreKind]idset.IDSet{PerformanceCore: sys.onlineCPUs.Clone()}
```

Otherwise we could theoretically end up with overlapping set. Also would make the code more explicit. WDYT?

marquiz: Also, should we consider the possibility of a 3rd core kind being added, like "atom-low-power" á la Meteor Lake. Basically ensure that every CPU gets assigned to some kind? Can be left as a future exercise (with a note?)

klihub: > minor nit: in case of any error, should we bail out clearly with smth like
> 
> ```go
>     sys.coreKindCPUs = map[CoreKind]idset.IDSet{PerformanceCore: sys.onlineCPUs.Clone()}
> ```
> 
> Otherwise we could theoretically end up with overlapping set. Also would make the code more explicit. WDYT?

https://github.com/containers/nri-plugins/pull/282 adds sanity checks for the detected CoreKinds, including a check for non-overlapping sets.

klihub: > Also, should we consider the possibility of a 3rd core kind being added, like "atom-low-power" á la Meteor Lake. Basically ensure that every CPU gets assigned to some kind? Can be left as a future exercise (with a note?)

Definitely we should add it. I didn't know it's a thing already...

marquiz: I don't think that exists (haven't checked, though). That was just a note that maybe in the (not so distant) future we see more than two types

pkg/sysfs/system.go
klihub: It's set to -1 before the loop then, if there are any caches, it gets set to the `level` of the `index0` cache, to collect both types of caches at that level if the cache is not unified (which it usually is not).

pkg/sysfs/system.go
klihub: @marquiz I reworked the patch series excluding anything related to hybrid architectures from this PR. It now only has the cache detection, cluster detection and clustered allocation bits.

I moved all infra hybrid architecture discovery bits and the related policy-specific bits to #282. Now I'm also considering splitting that into two parts: the infra bits and policy-specific bits. WDYT ?

pkg/sysfs/system.go
klihub: But you were right, it haphazard. Not only that, it was also broken. Fixed and rewrote it to be hopefully more easy to follow.

marquiz (approved):
LGTM

fmuyassarov (approved):
lgtm

PR #208, opened 2023-12-15
Add nri-plugins operator

fmuyassarov:
Move the operator code from:
1. https://github.com/containers/nri-plugins-operator 
2. https://github.com/containers/nri-plugins-operator/pull/3
3. https://github.com/containers/nri-plugins-operator/pull/4

Also, refactor the nri-plugins so that we can reuse the already existing CRD yamls instead of duplicating them.
There is one thing left to be added, which is image build refactoring. I will add it soon, but the rest is ready for review.


klihub:
A general comment about the PR... Although this is definitely not a showstopper/biggie, but since we're lifting existing code over from another repository, I wonder if the current PR commit history makes sense for each commit, or if it could be squashed to fewer ones. For instance, cannot we create the operator directly in the final location instead of first adding it in one location then moving it to another ? Also, it looks to me like the first two commits could as well be squashed together.

If I try to take a milky-way perspective at the full set of additions in this PR, it seems to me that a more logical set of commits would be:
1. add operator (commits 1, 2, 5, 6 squashed)
2. add operator bundle (commits 4, 7 squashed)
3. add operator scorecard testing  (commit 3)
4. add/link in documentation for the operator (commit 8)

operator/config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: fmuyassarov/ansible
-  newTag: v10
```
klihub: IMO these should be squashed into the original commit that introduces the operator.

operator/config/rbac/role.yaml
```diff
@@ -25,17 +25,13 @@ rules:
...
-      - customresourcedefinitions
+    - noderesourcetopologies
     verbs:
-      - create
-      - delete
-      - get
-      - list
-      - patch
-      - update
-      - watch
```
klihub: These could be squashed into the first commit as well.

operator/config/crd/bases/topology.node.k8s.io_noderesourcetopologies.yaml
```diff
@@ -0,0 +1,270 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870
+    controller-gen.kubebuilder.io/version: v0.11.2
+  creationTimestamp: null
+  name: noderesourcetopologies.topology.node.k8s.io
+spec:
```
klihub: Don't we have a copy of this already in the repo ? Then we probably shouldn't add this again, just use a single copy.

operator/roles/plugin-deployment/tasks/main.yml
```diff
@@ -19,6 +19,7 @@
         wait: True
         create_namespace: True
         values: "{{ values | default({}) }}"
+        skip_crds: True
```
klihub: Could be squashed into the first commit.

klihub:
First commit message has an outdated reference to `PluginConfig` which I think should be `NriPluginDeployment` instead.

sample-configs/noderesourcetopology.yaml
```diff
@@ -0,0 +1,103 @@
+---
```
klihub, line 1: Does this sample make sense ? `NodeResourceTopology` is semantically not really a configuration custom resource. Instead it reflects the current state of the node.

klihub:
@fmuyassarov Here is a rewritten history which is not squashed to a single commit and is more along the lines what I've been thinking of: https://github.com/klihub/nri-plugins/tree/devel/plugins-operator

sample-configs/noderesourcetopology.yaml
fmuyassarov: I didn't want to add it myself too but unfortunately I had to to pass the operator-sdk validation warning.
I understand that we don't expect users to interact with this CRD but olm expects you to place the CR instance of the CRD that the operator is owning.

fmuyassarov:
@klihub @marquiz @askervin @kad PTAL. It is ready now.

deployment/operator/README.md
```diff
@@ -0,0 +1,69 @@
...
+nri-plugins. The operator deploys community maintained [nri-plugins](https://github.com/containers/nri-plugins) in
+Kubernetes cluster. When operator is installed, it doesn't do anything apart from watching for custom resources called
+NriPluginDeployment. When NriPluginDeployment object is created, reconciliation loops kicks off and installs the
+nri-plugin specified in the NriPluginDeployment. 
+
+## Installation
+
+Build the operator image and push it to some registry
+```shell
+make docker-build docker-push IMG="my-registry.com/nri-plugins-operator:unstable"
```
klihub, line 15: Shouldn't this be
```
make -C deployment/operator docker-build docker-push IMG="my-registry.com/nri-plugins-operator:unstable"
```

deployment/operator/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+Deploy the operator in your cluster
+```shell
+make deploy
```
klihub, line 20: ditto for this

deployment/operator/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+Uninstall the operator
+```shell
+make undeploy
```
klihub, line 25: ditto

sample-configs/noderesourcetopology.yaml
klihub: Hmm... I wonder if this CRD will get removed when we remove the operator. Also wonder what happens, if this CRD is already present when we install the operator.

klihub (approved):

klihub:
@askervin @marquiz @kad PTAL.

Makefile
```diff
@@ -235,6 +240,14 @@ image.%:
...
 	$(DOCKER) image save "$${NRI_IMAGE_REPOTAG}" > "$${NRI_IMAGE_TAR}";
 
+operator-image:
+	$(Q)mkdir -p $(IMAGE_PATH); \
+	$(MAKE) -C deployment/operator image-save
+
+operator-bundle-image:
+	$(Q)mkdir -p $(IMAGE_PATH); \
+	$(MAKE) -C deployment/operator bundle-save
+
```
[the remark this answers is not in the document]
fmuyassarov: Agreed. Fixed.

deployment/operator/README.md
fmuyassarov: fixed

kad (commented):
I don't have any concerns about this PR, except deletion of files under docs/deployment/. Is that expected?

fmuyassarov:
> I don't have any concerns about this PR, except deletion of files under docs/deployment/. Is that expected?

We are not deleting docs but rather re-structuring them to accommodate the operator document. You can also see from the changes lines at the top of the PR.

kad (approved):
ok. mine bad, didn't notice moving files around. LGTM

PR #209, opened 2023-12-18
test/e2e: use ansible.builtin.shell for tasks that redirecting output.

klihub:
We can't use `ansible.builtin.command` for tasks that create or update files by redirecting output. These all need `ansible.builtin.shell` instead.


fmuyassarov (approved):

askervin (approved):
LGTM

PR #211, opened 2023-12-21
 config-manager: improve systemd unit listing

fmuyassarov:
Don't rely on ListUnitsByPatternsContext method as it might be too new
and missing in some of the distros like Amazon Linux 2. Instead
filter out active units and then loop over those to check if either
containerd or cri-o units exist.

### DIFF
--- a/cmd/config-manager/main.go
+++ b/cmd/config-manager/main.go
@@ -180,22 +180,29 @@ func detectRuntime() (string, *dbus.Conn, error) {
 	// (either containerd or CRI-O).If more than one container runtime systemd unit is found
 	// to be in an active state, the process fails.
 	log.Infof("looking for active runtime units on D-Bus...")
-	units, err := conn.ListUnitsByPatternsContext(context.Background(), []string{"active"}, []string{containerdUnit, crioUnit})
+	units, err := conn.ListUnitsFilteredContext(context.Background(), []string{"active"})
 	if err != nil {
 		return "", nil, fmt.Errorf("failed to detect container runtime in use: %w", err)
 	}
 
-	if len(units) == 0 {
+	expectedUnits := make([]dbus.UnitStatus, 0)
+	for _, unit := range units {
+		if unit.Name == containerdUnit || unit.Name == crioUnit {
+			expectedUnits = append(expectedUnits, unit)
+		}
+	}
+
+	if len(expectedUnits) == 0 {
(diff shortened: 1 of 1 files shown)


cmd/config-manager/main.go
```diff
@@ -180,22 +180,30 @@ func detectRuntime() (string, *dbus.Conn, error) {
...
+	units, err := conn.ListUnitsFilteredContext(context.Background(), []string{"active"})
+
+	expectedUnits := make([]dbus.UnitStatus, 0)
+	for _, unit := range units {
+		if unit.Name == "containerd.service" || unit.Name == "crio.service" {
+			expectedUnits = append(expectedUnits, unit)
+		}
+	}
+
 	if err != nil {
```
kad, line 192: probably error handling would be better before `make()`

cmd/config-manager/main.go
```diff
@@ -180,22 +180,30 @@ func detectRuntime() (string, *dbus.Conn, error) {
 	// (either containerd or CRI-O).If more than one container runtime systemd unit is found
 	// to be in an active state, the process fails.
 	log.Infof("looking for active runtime units on D-Bus...")
-	units, err := conn.ListUnitsByPatternsContext(context.Background(), []string{"active"}, []string{containerdUnit, crioUnit})
+	units, err := conn.ListUnitsFilteredContext(context.Background(), []string{"active"})
+
+	expectedUnits := make([]dbus.UnitStatus, 0)
+	for _, unit := range units {
+		if unit.Name == "containerd.service" || unit.Name == "crio.service" {
```
klihub, line 187: nitpick: There are `containerdUnit` and `crioUnit` consts defined for these, so probably those should be used here instead.

cmd/config-manager/main.go
klihub: +1

klihub (changes requested):
Small nit, otherwise LGTM.

cmd/config-manager/main.go
fmuyassarov: Ops, I accidentally moved already existing error handling down. But I have now fixed it.

cmd/config-manager/main.go
fmuyassarov: Good catch. Fixed.

klihub (approved):
LGTM.

kad (approved):
Lgtm

PR #217, opened 2023-12-28
operator: switch to community defined category

fmuyassarov:
https://github.com/k8s-operatorhub/community-operators doesn't support adding custom categories and such dropping using custom categories and switching to Openshift Optional to pass CI checks in https://github.com/k8s-operatorhub/community-operators/pull/3685.


askervin (approved):
LGTM

PR #218, opened 2023-12-28
operator: point containerImage tag to the latest release (v0.3.2)

fmuyassarov:
Ensure that we `sed` the image tag to the next release tag before pushing a new tag.
This will ensure that when new tag is pushed and image build is started, our images 
will have the right manifests.


kad (approved):
it means we need to document that prep step for future releases.

kad (approved):
lgtm. 
please don't forget to open PR for main branch with same script change.

askervin (approved):
LGTM

PR #219, opened 2023-12-29
operator: refactor release prep script to patch operator manifests too

fmuyassarov:
Ensure that we `sed` the image tag to the next release tag before pushing a new tag.
This will ensure that when new tag is pushed and image build is started, our images
will have the right manifests.


deployment/operator/config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: ghcr.io/containers/nri-plugins-operator
+  newName: ghcr.io/containers/nri-plugins/nri-plugins-operator
```
klihub, line 7: While we're fixing this... I'd prefer to have a `yaml.in` file (which is part of the repository), which is then copied and [s]edited into the resulting `.yaml` (which is not part of the repository). Could you fix this as well ?

Otherwise in the current state `make images`  (`make  nri-plugins-operator-bundle-image` more specifically) always leaves the repository with changed files in an uncommitted state which one needs to undo with a `git checkout deployment/operator/config/manager/kustomization.yaml` or something else to the same effect.

hack/release-helper.sh
```diff
@@ -22,6 +22,8 @@ fi
 version=$1
```
klihub, line 22: While at it, could you move this into the (already existing) `scripts/hack` directory where other similar scripts are ?

hack/release-helper.sh
```diff
@@ -34,3 +36,9 @@ find deployment/helm -name Chart.yaml | xargs -I '{}' \
 find deployment/helm -name values.yaml | xargs -I '{}' \
     sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" -i '{}'
 
+# Patch deployment templates
+echo Patching kustomize templates to use $container_image
+find deployment/operator/config/manifests/bases -name nri-plugins-operator.clusterserviceversion.yaml | xargs -I '{}' \
+    sed -E -e s",^([[:space:]]+)containerImage:.+$,\1containerImage: $container_image," -i '{}'
+find deployment/operator/config/manager -name kustomization.yaml | xargs -I '{}' \
+    sed -E -e s",^([[:space:]]+)newTag:.+$,\1newTag: $version," -i '{}'
```
klihub, line 44: So these inline-[s]edits would be IMO better to replace with [s]editing `.yaml.in`s into corresponding `.yaml`s to keep these actions idempotent wrt. to both the repo content and state.

deployment/operator/config/manager/kustomization.yaml
klihub: @fmuyassarov For instance something like this: #224

hack/release-helper.sh
klihub: @fmuyassarov And with #224 in place, you'd need to do something like this instead:

```
...
make -C deployment/operator kustomizations && \
    find deployment/operator/config/manager -name kustomization.yaml | xargs -I '{}' \
        sed -E -e s",^([[:space:]]+)newTag:.+$,\1newTag: $version," -i '{}'
```

hack/release-helper.sh
fmuyassarov: yes

fmuyassarov:
@klihub 
I noticed that my changes were pushed directory to the upstream branch instead of to my work. 
As such, I have to close this PR here and we can continue reviewing the changes in https://github.com/containers/nri-plugins/pull/235.

Issue #194, opened 2023-11-16
Question: Adding Kubernetes priority class to NRI plugin deployment manifests


ffuerste:
### Description

As described in the [NRI (security) docs](https://github.com/containerd/nri#security-considerations) NRI plugins should be considered as part of the container runtime. Therefore, maybe the Kubernetes priority class [`system-node-critical`](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical) could be added for the NRI plugin deployment manifests.

Or maybe adding the priority class not in the deployment manifest as a default, but discussing it as an option/recommendation as part of the README of the plugins?

### Rationale

Adding the Kubernetes built-in priority class `system-node-critical` would ensure, that the particular NRI plugin is not evicted (easily) and available when pods are scheduled for which the respective plugin would be responsible for.

E.g. if a pod is created by a user for which a limit for EPC memory is defined, the sgx-epc NRI plugin must be up and running in order to ensure, that the limit is configured correctly as part of the misc cgroup of the container. Currently the sgx-epc NRI plugin supports only creating new containers. However, if containers are already existing when the plugin registers itself to the container runtime, it would need to request an update for existing containers as described [here](https://github.com/containerd/nri#plugin-registration). This use case is currently not supported by the sgx-epc plugin. Therefore, adding the `system-node-critical` priority class to the deployment manifest for it, could mitigate potential risk in a running system under load.

However, I am not sure, if this makes sense for all currently available plugins or future ones. But it would be great, if we could define a common policy/best practice for it (as suggested by @mythi [here](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1582#discussion_r1387948398)).

mythi:
> Therefore, maybe the Kubernetes priority class [`system-node-critical`](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical) could be added for the NRI plugin deployment manifests.

@fmuyassarov any thoughts on this? I think it makes sense but before I add it to my deployment, I thought we could align with the direction.

fmuyassarov:
I don't see reason to say not and that can be made optional in our Helm charts so that whoever is interested can opt in for the feature by setting a flag during the Helm installation. Any objections @kad @klihub @marquiz ?

kad:
yes, I think it makes sense to add priorityClass

PR #220, opened 2024-01-06
feat(helm): Introduce priorityClassName system-node-critical

ffuerste:
# What does this PR do?

Introducing in the helm charts of all NRI plugins an optional boolean field for setting `priorityClassName: system-node-critical`.

As suggest [here](https://github.com/containers/nri-plugins/issues/194#issuecomment-1833377246) the `priorityClassName` is introduced as an optional helm flag analog to `patchRuntimeConfig`. However, in this way the feature can only be enabled with no option to define a custom `priorityClassName`. 

## Related issues

Resolves containers/nri-plugins#194


marquiz (commented):
Thanks for the patch @ffuerste. This totally makes sense.

I'm just wondering should be the desired default be `true`, for at least some of the plugins 🧐 If a resource policy plugin (like balloons or topology-aware) is deployed I think it is kinda critical and should not be evicted (thus default to `true`). WDYT? @klihub @fmuyassarov @askervin @kad

askervin (approved):
I think this is a very nice improvement.

I'm fine both merging this PR as it is, or with some/all existing plugins set as `system-node-critical` by default.

Maybe I'd even prefer taking this PR is as it is, as it only enables configuring the priorityClassName but does not change the default behavior. And then have a separate commit that changes the default behavior for chosen plugins.

fmuyassarov:
> Thanks for the patch @ffuerste. This totally makes sense.
> 
> I'm just wondering should be the desired default be `true`, for at least some of the plugins 🧐 If a resource policy plugin (like balloons or topology-aware) is deployed I think it is kinda critical and should not be evicted (thus default to `true`). WDYT? @klihub @fmuyassarov @askervin @kad

I don't have a strong opinion about that to be frank. I'm fine with the current state as well.

kad:
I think it makes sense.

klihub (approved):
LGTM.

klihub:
> Thanks for the patch @ffuerste. This totally makes sense.
> 
> I'm just wondering should be the desired default be `true`, for at least some of the plugins 🧐 If a resource policy plugin (like balloons or topology-aware) is deployed I think it is kinda critical and should not be evicted (thus default to `true`). WDYT? @klihub @fmuyassarov @askervin @kad

I'm fine with the default either way for the current plugins.

deployment/helm/balloons/values.yaml
```diff
@@ -42,6 +42,10 @@ resources:
 
 nri:
   patchRuntimeConfig: false
+  # NRI plugins should be considered as part of the container runtime.
+  # Therefore, adding the system-node-critical priority class to the DaemonSet,
+  # could mitigate potential risk in a running system under load.
+  setPodPriorityClassNodeCritical: false
```
marquiz, line 48: Two suggestions:

1. Don't put the new config field under `nri:` but on the top level. I think this is better in line with the other pod level settings like `toleration` and `nodeSelector`
2. Streamline the field name a bit, drop the "set" from the beginning

```suggestion
podPriorityClassNodeCritical: false
```

deployment/helm/balloons/README.md
```diff
@@ -98,6 +98,7 @@ customize with their own values, along with the default values.
 | `hostPort`               | 8891                                                                                                                          | metrics port to expose on the host                   |
 | `config`                 | see [helm chart values](tree:/deployment/helm/balloons/values.yaml) for the default configuration                       | plugin configuration data                            |
 | `nri.patchRuntimeConfig` | false                                                                                                                         | enable NRI in containerd or CRI-O                    |
+| `nri.setPodPriorityClassNodeCritical` | false                                                                                                                         | enable [marking Pod as node critical](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical)                       |
```
marquiz, line 101: Accompaniying my other comment:

```suggestion
| `podPriorityClassNodeCritical` | false                                                                                                                         | enable [marking Pod as node critical](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical)                       |
```

...and ditto all charts

marquiz (changes requested):
Hold until comments are resolved

deployment/helm/balloons/values.yaml
ffuerste: @marquiz I have updated the helm charts as suggested. I have added the "new" value `podPriorityClassNodeCritical` at the end of the `values.yaml` files to indicate that it is similar to the `toleration` and `nodeSelector` values. Please check again. Thank you very much!

deployment/helm/balloons/README.md
ffuerste: @marquiz I have updated the helm charts as suggested. I've added the description of the "new" value `podPriorityClassNodeCritical` at the end of the helm value table, reflecting its position in the `values.yaml` files. I am resolving this thread in favour of the first thread.

marquiz (approved):
Thanks @ffuerste. LGTM on my part

fmuyassarov (approved):

PR #222, opened 2024-01-09
pkg/apis, helm: update balloons config/CRD.

klihub:
It looks like the latest balloons configuration additions have been committed then merged without updating the configuration, CRDs and clientsets. Let's fix it. Also, while at it let's add generated artifact verification to our github verification workflow.


config/crd/bases/config.nri_balloonspolicies.yaml
```diff
@@ -147,6 +159,8 @@ spec:
                       type: string
                   required:
                   - name
+                  - preferCloseToDevices
```
askervin, line 162: `preferCloseToDevices` is fully optional preference in balloon types. Does listing it in `required` make it mandatory in some validation?

I would hide `preferFarFromDevices` from CRDs.

config/crd/bases/config.nri_balloonspolicies.yaml
```diff
@@ -106,6 +106,18 @@ spec:
                       items:
                         type: string
                       type: array
+                    preferCloseToDevices:
+                      description: 'PreferCloseToDevices: prefer creating new balloons
+                        of this type close to listed devices.'
+                      items:
+                        type: string
+                      type: array
+                    preferFarFromDevices:
```
askervin, line 115: Could we hide `preferFarFromDevices` from CRDs? It's undocumented, untested (as an explicitly given option), and the policy fills it internally to reflect `preferCloseToDevices` configurations from other balloon types.

"Internal filling" does not mean "overwriting", but only "adding to it". In other words, it could be made part of public interface, too, but I'm not sure if we want to do that.

askervin (commented):
Great that you spotted this, @klihub!

It seems that I added `preferCloseToDevices` only to helm deployent crd in commit 101642756d4f6b31e2e48796fc6a2b024107b39a. Should have done more careful grep <existing-option> to find other CRD files to be modified. As explained, `preferFarFromDevices` was intentionally left out from CRDs for now.

fmuyassarov:
I wonder if we should extend one of the workflows or introduce a new one to check if all CRDs are re-generated correctly.

config/crd/bases/config.nri_balloonspolicies.yaml
fmuyassarov: > `preferCloseToDevices` is fully optional preference in balloon types. Does listing it in `required` make it mandatory in some validation?

Required means, when user creates CR, it will be mandatory to set `preferCloseToDevices `to some value. If there is default value, then it the default value will be used if the user doesn't set it. But you don't have default here, so user must set it.

fmuyassarov: Since [preferCloseToDevices ](https://github.com/containers/nri-plugins/blob/2308d1f5a296283e4b0ec0ad0fbdf841311dcebb/pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go#L191 ) is omitempty, I think it should be under optional in the CRD YAML. I wonder if our CRD generation has a bug or did you add it manually @klihub ?

klihub: Yes, I think so, too. Thanks for spotting that.

config/crd/bases/config.nri_balloonspolicies.yaml
klihub: I'm not sure, but I think/recall that if you set its golang struct member tag to "-" instead of `preferCloseToDevices` then it will not be available for unmarshalling (hence you shouldn't be able to set it from a CRD).

config/crd/bases/config.nri_balloonspolicies.yaml
klihub: Actually, no. `omitempty` would imply `// +optional` had it been correctly added to the tags like it was for the rest of the struct members.

fmuyassarov: Yes, correct. We need `// +optional` marker if we want the field to be optional.

klihub:
> Great that you spotted this, @klihub!
> 
> It seems that I added `preferCloseToDevices` only to helm deployent crd in commit [1016427](https://github.com/containers/nri-plugins/commit/101642756d4f6b31e2e48796fc6a2b024107b39a). Should have done more careful grep to find other CRD files to be modified. As explained, `preferFarFromDevices` was intentionally left out from CRDs for now.

All our CRD's are generated and should not be touched by hand. The proper way of updating them is to update the corresponding types somewhere under `pkg/apis/config/vialpha1/**` then run `make generate`. This should generate the CRDs, copy them to the necessary duplicate places and make sure that running `make generate` after that is idempotent until further similar changes. We'll probably want to add a github verification workflow which just runs `make generate` and verifies that it does not induce any changes to the repo.

klihub:
> I wonder if we should extend one of the workflows or introduce a new one to check if all CRDs are re-generated correctly.

I had exactly the same idea. I think too that we should.

config/crd/bases/config.nri_balloonspolicies.yaml
klihub: Okay, so I chose to tag `PreferFarFromDevices` with `"json:-"` to prevent it from being marshalled or unmarshalled from a CRD to the actual configuration. That's the most proper way I could come up with. The alternative maybe would have been to add a validation to force it to be `false`, but I think this is both simpler and better.

klihub:
> > I wonder if we should extend one of the workflows or introduce a new one to check if all CRDs are re-generated correctly.
> 
> I had exactly the same idea. I think too that we should.

Hooked in new `verify-generate` to the top level `verify` target. Now our github CI workflows should verify that all generated artifacts are properly committed and up-to-date in the repo.

fmuyassarov (approved):
Thank you @klihub for the fixes.
LGTM

askervin (approved):
LGTM

klihub:
#230 is a backport of this to the release-0.3 branch.

PR #223, opened 2024-01-10
resmgr: inject mount before container state update.

klihub:
Inject our 'downward API' mount into containers before updating their state to 'created'. Otherwise, if a policy doesn't make any changes to a container during its creation, we incorrectly consider the container as already created and fail to insert the mount.


fmuyassarov (approved):
LGTM

fmuyassarov:
@klihub do we want to backport it to other release branches?

klihub:
> @klihub do we want to backport it to other release branches?

Yes. #226

PR #227, opened 2024-01-11
helm: allow setting NRI plugin index via values.

klihub:
Update helm charts to allow setting the index our plugins use to register with NRI. The plugin-specific defaults remain the same.


klihub:
@fmuyassarov I added the `nri` section to the values verification schema. I added range verification for the newly introduced integer plugin index, but it does not work. Can you spot and tell me what I am doing wrong there ?

klihub:
> @fmuyassarov I added the `nri` section to the values verification schema. I added range verification for the newly introduced integer plugin index, but it does not work. Can you spot and tell me what I am doing wrong there ?

Okay, to answer myself, I figured it out. Looks like all of our validation schema files a misnamed. They are called `values.scheme.json` while they should be `values.schema.json`.

fmuyassarov (approved):
LGTM
Thank you for the fixes @klihub

fmuyassarov:
And we defiantly need it to be backported to release branch as well. At least to `release-0.3`

klihub:
> And we defiantly need it to be backported to release branch as well. At least to `release-0.3`

#229

PR #231, opened 2024-01-12
fixes: rename default config group label, support/fall back to deprecated labels. 

klihub:
This patch set
- changes the default configuration group label to `config.nri/group`
- falls back to look for deprecated labels (`group.config.nri`, `resource-policy.nri.io/group`) if the configured label is absent
- allows overriding the default configuration group label on the command line via helm charts


marquiz (commented):
I think this is ok in the project in this early stage. But generally I think deprecating things in a patch release are questionable (or other functional changes)

klihub:
> I think this is ok in the project in this early stage. But generally I think deprecating things in a patch release are questionable (or other functional changes)

IMO as long as we keep the deprecated things intact/supported throughout the whole main release cycle where the patch release belongs to (IOW in this case 0.3.x) and if timing requires (IOW deprecation is too close to the end of the main release lifecycle) also during the next one (IOW in our case that would be 0.4.x), then it should be perfectly fine to deprecate things in any patch release.

askervin (approved):
Looks great! Thank you @klihub!

fmuyassarov (approved):
LGTM

PR #234, opened 2024-01-15
instrumentation: allow proper reconfiguration of tracing.

klihub:
Allow tracing to be properly reconfigured via configuration updates, including proper restart of tracing if necessary. 
We can't simply shut down our (global) tracing provider since the global provider can neither be subsequently changed nor restarted. Therefore update the tracing code to
  - only flush the provider without shutdown during Stop()
  - disable sampling when no endpoint is set
  - log any errors encountered during restart

### DIFF
--- a/pkg/instrumentation/instrumentation.go
+++ b/pkg/instrumentation/instrumentation.go
@@ -91,7 +91,13 @@ func Restart() error {
 	defer lock.Unlock()
 
 	stop()
-	return start()
+
+	err := start()
+	if err != nil {
+		log.Error("failed to start tracing: %v", err)
+	}
+
+	return err
 }
 
 // Reconfigure our instrumentation services.
--- a/pkg/instrumentation/tracing/sampler.go
+++ b/pkg/instrumentation/tracing/sampler.go
@@ -1,0 +1,65 @@
+// Copyright The NRI Plugins Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tracing
+
+import (
+	"sync"
+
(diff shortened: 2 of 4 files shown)


pkg/instrumentation/instrumentation.go
```diff
@@ -91,7 +91,13 @@ func Restart() error {
 	defer lock.Unlock()
 
 	stop()
-	return start()
+
+	err := start()
+	if err != nil {
+		log.Error("failed to start tracing: %v")
```
askervin, line 97: missing err in the error message.

pkg/instrumentation/tracing/sampler.go
```diff
@@ -0,0 +1,65 @@
...
+	drop = sdktrace.SamplingResult{
+		Decision:   sdktrace.Drop,
+		Tracestate: trace.TraceState{},
+	}
+
+	_ sdktrace.Sampler = &sampler{}
+)
+
+type sampler struct {
+	sync.RWMutex // TODO(klihub): should we use an atomic type instead ?
```
askervin, line 34: Are you sure we want to keep // TODO in the code base? The comment looks more like a consideration than something that needs to be done.

pkg/instrumentation/tracing/tracing.go
```diff
@@ -103,28 +104,33 @@ func Start(options ...Option) error {
 
 // Stop tracing.
 func Stop() {
-	trc.shutdown()
+	trc.flush()
```
askervin, line 107: Why not call `trc.shutdown()`, as there is such a function. If calling flush is better, then perhaps consider removing `shutdown()` as it does not seem to be called anywhere.

askervin (commented):
Found one issue, and added two questions to make sure there are no accidents...

pkg/instrumentation/tracing/tracing.go
klihub: Because I was supposed to remove `shutdown()` altogether, but I forgot. Thanks for spotting it, updated.

pkg/instrumentation/instrumentation.go
klihub: Thanks, fixed.

pkg/instrumentation/tracing/sampler.go
klihub: Well, I wanted to leave that in for myself for reconsideration... but removed it now.

marquiz (approved):
lgtm

fmuyassarov:
LGTM

fmuyassarov (approved):

PR #244, opened 2024-01-25
agent: fix the (short) import path for NRT api

fmuyassarov:


marquiz (approved):
LGTM

klihub (approved):
Out of curiosity: why is this a 'fix' ?

marquiz:
> Out of curiosity: why is this a 'fix' ?

looks like unifying the import name

fmuyassarov:
Yea probably I should be careful with the wording here. This is not a fix indeed and just a cosmetic change to follow the same pattern.

PR #245, opened 2024-01-25
resmgr: log containers by pretty name during startup.

klihub:
Log containers by pretty name during initial sync with the runtime. Log discovered containers simply according to their state as 'created/running' or 'stopped'. Don't refer to them as 'out-of-sync running' or 'stale stopped', since we anyway try to force reallocation of resources.


marquiz (approved):
LGTM

fmuyassarov (commented):
LGTM

PR #246, opened 2024-01-25
helm: reduce CRD duplication by reusing CRDs from root directory

fmuyassarov:
Rather than having redundant copies of CRDs within the helm directories, reuse CRDs from the top-level 'config/crd/bases' directory.


klihub:
> The thing I really dislike about this approach is that...

But since you have that 'copy-helm-crds' Makefile target which takes care of the missing Helm CRDs, I am also fine with merging this if @marquiz and @askervin think that's better.

askervin:
I'd suggest writing a proper rule for creating CRDs from base CRDs. In a proper rule the target (resulting CRD) is followed by a complete set of prerequisites, that is base CRDs and whatever else affects the result. Then make will take care of creating and updating produced CRDs when necessary, and avoid doing so unnecessarily.

@klihub, I still suppose that your worry about developer workflow is not solved by this suggestion. Should there be something totally different in your point of view?

klihub:
> I'd suggest writing a proper rule for creating CRDs from base CRDs. In a proper rule the target (resulting CRD) is followed by a complete set of prerequisites, that is base CRDs and whatever else affects the result. Then make will take care of creating and updating produced CRDs when necessary, and avoid doing so unnecessarily.

I cannot come up with a ruleset that is proper yet concise/easy to comprehend. This is the only one I could quickly throw together and, sans Stockholm syndrome, I suspect this might not be viewed as a clear improvement over the current situation or the proposed one:

```
#
# rules to update generated CRDs in our helm charts
#

# canned rule for declaring per-plugin Helm chart CRDs
define declare-crds
    $(foreach crd,$(2),$(HELM_TOP_DIR)/$(1)/crds/$(crd).yaml)
endef

# canned recipe for updating one CRD
define update-crd =
	echo "Updating CRD $@ (from $<)..."; \
	cp $< $@
endef

# per-plugin set of CRDs
BALLOONS_PLUGIN_CRDS := \
    config.nri_balloonspolicies topology.node.k8s.io_noderesourcetopologies
TEMPLATE_PLUGIN_CRDS := \
    config.nri_templatepolicies topology.node.k8s.io_noderesourcetopologies
TOPOLOGYAWARE_PLUGIN_CRDS := \
    config.nri_topologyawarepolicies topology.node.k8s.io_noderesourcetopologies

# colective rule to update all (out-of-date) CRDs
.PHONY: update-helm-crds
update-helm-crds: \
    $(call declare-crds,balloons,$(strip $(BALLOONS_PLUGIN_CRDS))) \
    $(call declare-crds,template,$(strip $(TEMPLATE_PLUGIN_CRDS))) \
    $(call declare-crds,topology-aware,$(strip $(TOPOLOGYAWARE_PLUGIN_CRDS)))

# per-plugin rules for updating CRDs.
$(HELM_TOP_DIR)/balloons/crds/%.yaml: $(CRD_BASE_DIR)/%.yaml
	@$(call update-crd)

$(HELM_TOP_DIR)/template/crds/%.yaml: $(CRD_BASE_DIR)/%.yaml
	@$(call update-crd)

$(HELM_TOP_DIR)/topology-aware/crds/%.yaml: $(CRD_BASE_DIR)/%.yaml
	@$(call update-crd)
```


> @klihub, I still suppose that your worry about developer workflow is not solved by this suggestion. Should there be something totally different in your point of view?

For my workflow it is enough to have a single rule I can use to make sure my helm charts are up-to-date. Currently I have been using to top-level `generate` rule for that (which then used `update-helm-crds`). And with the proposed changes that would be/should become `copy-helm-crds` (which is now missing from generate, AFAICT).

marquiz (commented):
Hmm, I also dislike this approach (breaking simple helm install from git clone). I also very much dislike complicated (and not generilable) makefile rules for this kind of simple copying.

Why not just use symlinks? I think Helm should handle those

fmuyassarov (commented):
Thanks for the feedback folks. Since it seems this is not the right approach I willl close the PR for now and we can come back to it whenever it is needed and when we have some better ideas.

PR #249, opened 2024-01-28
cache: support annotations to preserve CPU and memory pinning

askervin:
Add "ignore" annotation to enable special containers to opt out from resource allocation of any resource policy. Resource policies can find these containers and their resource requirements from the cache, but they are not requested to allocate, update or release resources of these containers during their lifecycle and on policy Sync().

### DIFF
--- a/pkg/resmgr/cache/cache.go
+++ b/pkg/resmgr/cache/cache.go
@@ -59,6 +59,11 @@ const (
 
 	// TopologyHintsKey can be used to opt out from automatic topology hint generation.
 	TopologyHintsKey = "topologyhints" + "." + kubernetes.ResmgrKeyNamespace
+
+	// PreserveCpuKey means that CPU resources should not be touched.
+	PreserveCpuKey = "cpu.preserve." + kubernetes.ResmgrKeyNamespace
+	// PreserveMemoryKey means that memory resources should not be touched.
+	PreserveMemoryKey = "memory.preserve." + kubernetes.ResmgrKeyNamespace
 )
 
 // PodState is the pod state in the runtime.
@@ -244,6 +249,13 @@ type Container interface {
 	GetMemoryLimit() int64
 	// GetMemorySwap gets the swap limit in bytes for the container.
 	GetMemorySwap() int64
+
+	// PreserveCpuResources() returns true if CPU resources of the
+	// container must not be changed.
+	PreserveCpuResources() bool
+	// PreserveMemoryResources() returns true if memory resources
+	// of the container must not be changed.
+	PreserveMemoryResources() bool
 
 	// GetPendingAdjusmentn clears and returns any pending adjustment for the container.
 	GetPendingAdjustment() *nri.ContainerAdjustment
--- a/pkg/resmgr/cache/container.go
+++ b/pkg/resmgr/cache/container.go
@@ -761,6 +761,16 @@ func (c *container) GetMemorySwap() int64 {
 	return c.Ctr.GetLinux().GetResources().GetMemory().GetSwap().GetValue()
 }
 
+func (c *container) PreserveCpuResources() bool {
+	value, ok := c.GetEffectiveAnnotation(PreserveCpuKey)
+	return ok && value == "true"
+}
+
+func (c *container) PreserveMemoryResources() bool {
+	value, ok := c.GetEffectiveAnnotation(PreserveMemoryKey)
+	return ok && value == "true"
+}
+
 var (
 	// More complext rules, for Kubelet secrets and config maps
 	ignoredTopologyPathRegexps = []*regexp.Regexp{


klihub (commented):
If I understand this correctly, this tries to implement a common way of ignoring containers by trying to hide them from the policy. IOW, instead of putting in some common mechanism to let policies knowingly ignore a container, (which you mostly have there, maybe sans a convenience function added to Cache.Container), this tries to hide the mere existence of such containers from policies (although AFAICT these containers are not fully hidden from policies).

My gut feeling (although I can't really put my finger on it why) is that it would be better at least to consider doing this the other way around: put in the necessary common infra for policies to query whether they should touch or ignore the container, but let the policies consciously ignore or handle the container-related operations.

klihub:
Some more practical details related to my reservations about this approach. 1) An ignored container can have resources assigned to it which now cannot be taken into account for subsequent allocations by the policy. 2) If a policy implements an unsolicited reallocation/adjustment of resource assignments (for instance a metrics-based 'reconciliation loop') and it does so by querying all containers from the cache, extra logic is still needed in the policy to ignore containers it did not get informed about.

klihub:
> I agree. It's better if policies are informed about every container that runs. Calling "AllocateResources" makes sense even if a container should not be pinned.
> 
> Do you think we could extended cache.Container interface with something like IgnoreCpuPinning() and IgnoreMemoryPinning()? And with this annotation both would return True?

Yes, I was also thinking something like that... adding some convenience functions to cache.Container, which would allow the policy to detect when a container's assignments should not be touched and bail out early on the allocation and update branches.

klihub:
> I agree. It's better if policies are informed about every container that runs. Calling "AllocateResources" makes sense even if a container should not be pinned.
> 
> Do you think we could extended cache.Container interface with something like IgnoreCpuPinning() and IgnoreMemoryPinning()? And with this annotation both would return True?

Maybe we can come up with a name that aptly suggests that the container's existing resource assignments should not be touched/altered, at most accounted for...

askervin:
Very nice! Perhaps this could enable use cases where a special-purpose resource plugin runs first in the chain, detects a special container, pins it memory and CPU, and adds an annotation that tells remaining plugins not to touch CPU and memory pinning of this container anymore? And a general-purpose resource plugin could just try to cope with the situation, as it notices that container's cpuset.cpus and cpuset.mems are "readonly" or "immutable" or "fixed" or "locked" or "unchangeable" or "unmodifiable" or whatever we wish to call them.

Alternatives how this could look in the cache.container interface:
- `IsCpusetCpuReadOnly()` and `container.IsCpusetMemsReadOnly()`? This could be better if it is possible for one plugin to change cpuset.cpus and other cpuset.mems.
- `container.IsCpusetReadOnly()` could be better if touching cpuset.cpus and cpuset.mems by different plugins causes a conflict error in NRI.

klihub:
> Very nice! Perhaps this could enable use cases where a special-purpose resource plugin runs first in the chain, detects a special container, pins it memory and CPU, and adds an annotation that tells remaining plugins not to touch CPU and memory pinning of this container anymore? And a general-purpose resource plugin could just try to cope with the situation, as it notices that container's cpuset.cpus and cpuset.mems are "readonly" or "immutable" or "fixed" or "locked" or "unchangeable" or "unmodifiable" or whatever we wish to call them.
> 
> Alternatives how this could look in the cache.container interface:
> 
> * `IsCpusetCpusReadOnly()` and `container.IsCpusetMemsReadOnly()`? This could be better if it is possible for one plugin to change cpuset.cpus and other cpuset.mems.
> * `container.IsCpusetReadOnly()` could be better if touching cpuset.cpus and cpuset.mems by different plugins causes a conflict error in NRI.

How about something like
- `PreserveAssignedResources()` or `ShouldPreserveAssignedResources()`: if we don't want to split CPU from memory, and
- `PreserveAssignedCpuResources()` and `PreserveAssignedMemoryResources()` or `ShouldPreserveAssignedCpu()` and `ShouldPreserveAssignedMemory()`: with accurately documented expectations about how a policy should react to these if we want to split CPU from memory

The implementation could simple check for the annotations you proposed in this PR and return true if it's set (for the container or the whole pod), false otherwise.

klihub:
@askervin Then again, recalling the offline discussion the other day that triggered filing this PR, the original/real itch to scratch is the fact that there exists a class of containers which need access to the full (or some other predetermined) set of CPU and memory nodes to successfully perform their designated tasks. This PR was filed with the assumption that simply ignoring such containers will always result in correct resource assignment.

Considering this, another alternative would be to call the annotations something like `NeedAccessToAllCpuset()` and `NeedAccessToAllMemoryNodes()` and then have the corresponding handling code in policies make sure that the containers indeed have the necessary access.

askervin:
Appetite comes with eating.

What I like about your idea regarding `PreserveAssigned{Cpu,Memory}Resources()` or setting specific attributes ReadOnly is that this would be generic enough to be extended to other adjustments as well. Not now, but somewhere in the future, if there is a need. By extending I mean that all NRI resource plugins could use similar interface for saying to other plugin later in the chain: "do not touch these attributes, they have been adjusted already". Currently NRI plugins do not know what they are allowed to touch and what causes a conflict.

If we choose `NeedAccessToAllCpuset()` and alike, I do not see similar benefits, but potentially a growing set of custom needs that may be harder to take into account in policies.

klihub:
> Appetite comes with eating.
> 
> What I like about your idea regarding `PreserveAssigned{Cpu,Memory}Resources()` or setting specific attributes ReadOnly is that this would be generic enough to be extended to other adjustments as well. Not now, but somewhere in the future, if there is a need. By extending I mean that all NRI resource plugins could use similar interface for saying to other plugin later in the chain: "do not touch these attributes, they have been adjusted already". Currently NRI plugins do not know what they are allowed to touch and what causes a conflict.
> 
> If we choose `NeedAccessToAllCpuset()` and alike, I do not see similar benefits, but potentially a growing set of custom needs that may be harder to take into account in policies.

Yes, I also tend to lean towards that... a resource preservation abstraction feels more generic.

askervin:
Thanks for the feedback & discussion, @klihub!

Now this PR is modified to support two separate annotations for preserving cpu and memory, and providing that information from the cache.

klihub (approved):

marquiz (approved):
LGTM

marquiz:
I updated the PR title

marquiz:
ping @klihub

klihub:
> ping @klihub

Let's wait for #260 to go in first....

PR #250, opened 2024-01-29
helm: add workaround for Helm int/float snafu.

klihub:
Helm is struggling with a regression where integer values get converted to float64 and therefore fail to get later properly injected into templates. For us this hits our integer [nri.]pluginIndex values.

Add a workaround to first force plugin indices to integers before injecting them through templates.
See https://github.com/helm/helm/commit/44a81f63 for more details.

The other alternative considered was this in `values.schema.json`:
```
                "pluginIndex": {
                    "type": "string",
                    "pattern": "^[0-9][0-9]$"
                }
```
Since indices are integers and the above alternative would have also required us to always quote indices in `values.yaml`, this simpler choice of representing integer indices as integer values was chosen.

### DIFF
--- a/deployment/helm/balloons/templates/daemonset.yaml
+++ b/deployment/helm/balloons/templates/daemonset.yaml
@@ -55,7 +55,7 @@ spec:
             - -metrics-interval
             - 5s
             - --nri-plugin-index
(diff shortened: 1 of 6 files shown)


fmuyassarov (approved):
LGTM

PR #252, opened 2024-02-05
topology-aware: log resource operations at info level.

klihub:
Log resource allocation/release/pinning operations at info level. Use pretty-name when logging container pinning.

PR #253, opened 2024-02-05
    resmgr: handle UpdateContainer requests better.

klihub:
Note: this PR is stacked on top of #252.

Try doing a better job of detecting and handling redundant UpdateContainer requests which do not really change any of a container's parameters. Such requests are typically sent by an active CPU Manager with the static policy enabled in kubelet.
    
Before evaluating an update request for redundancy fill in any missing values from the container's current resources. Then if the request looks redundant short-circuit handling it by sending a reply with the current resources. Otherwise pass it on to the policy.

### DIFF
--- a/pkg/resmgr/nri.go
+++ b/pkg/resmgr/nri.go
@@ -432,13 +432,36 @@ func (p *nriPlugin) UpdateContainer(ctx context.Context, pod *api.PodSandbox, co
 	}
 
 	if realUpdates := c.SetResourceUpdates(res); !realUpdates {
-		p.Warn("UpdateContainer with identical resources, ignoring it...")
-		return nil, nil
-	}
-	//r := cache.EstimateResourceRequirements(res, c.GetQOSClass())
-
-	if err := m.policy.UpdateResources(c); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/resmgr/nri.go
```diff
@@ -432,13 +432,33 @@ func (p *nriPlugin) UpdateContainer(ctx context.Context, pod *api.PodSandbox, co
...
+		}
+		if v := c.GetCpusetCpus(); v != "" {
+			c.SetCpusetCpus(v)
+		}
+		if v := c.GetCpusetMems(); v != "" {
+			c.SetCpusetMems(v)
+		}
+		if v := c.GetMemoryLimit(); v != 0 {
+			c.SetMemoryLimit(v)
+		}
```
askervin, line 453: Would it be good to short-circuit MemorySwap, too? Just for symmetry, if nothing else.

askervin (commented):
One question, otherwise LGTM.

askervin (approved):
LGTM

fmuyassarov (approved):
lgtm

PR #254, opened 2024-02-06
Resize the reserved balloon

askervin:
This patch makes the built-in balloon types first-class citizens in their configurability and in CPU allocation order.

### DIFF
--- a/docs/resource-policy/policy/balloons.md
+++ b/docs/resource-policy/policy/balloons.md
@@ -3,9 +3,10 @@
 ## Overview
 
 The balloons policy implements workload placement into "balloons" that
-are disjoint CPU pools. Balloons can be inflated and deflated, that is
-CPUs added and removed, based on the CPU resource requests of
-containers. Balloons can be static or dynamically created and
+are disjoint CPU pools. Size of a balloon can be fixed, or the balloon
+can be dynamically inflated and deflated, that is CPUs added and
+removed, based on the CPU resource requests of containers running in
+the balloon. Balloons can be static or dynamically created and
 destroyed. CPUs in balloons can be configured, for example, by setting
 min and max frequencies on CPU cores and uncore.
 
@@ -52,8 +53,7 @@ other policy. See [deployment](../../deployment/index.md) for more details.
 
 ## Configuration
 
-The balloons policy is configured using the yaml-based configuration
-system of nri-resource-policy.
+The balloons policy is configured using BalloonsPolicy Custom Resources.
 See [setup and usage](../setup.md#setting-up-nri-resource-policy) for
 more details on managing the configuration.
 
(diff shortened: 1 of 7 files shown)


askervin:
This patch still sticks with two questionable old design assumptions:
1. There is only one "default" balloon instance
2. If the default balloon has no explicit CPU-configuration, then it'll use the same CPUs as the "reserved" balloon. (This looks very suspicious now in the code and makes the logic difficult to understand.)

I'd be happy to simplify this design into one where the default balloon is just like any user-defined balloon. The only thing how it would differ from them is that if there is no user-defined "default" balloon type, then the balloons policy would append a built-in "default" balloon type that captures pods that are not assigned into any other balloon type.

What do you think, @klihub, @marquiz ,@kad? Would you agree, and would you prefer including this design change into this same PR? I think it would make this PR simpler to review.

klihub:
> This patch still sticks with two questionable old design assumptions:
> 
> 1. There is only one "default" balloon instance
> 2. If the default balloon has no explicit CPU-configuration, then it'll use the same CPUs as the "reserved" balloon. (This looks very suspicious now in the code and makes the logic difficult to understand.)
> 
> I'd be happy to simplify this design into one where the default balloon is just like any user-defined balloon. The only thing how it would differ from them is that if there is no user-defined "default" balloon type, then the balloons policy would append a built-in "default" balloon type that captures pods that are not assigned into any other balloon type.
> 
> What do you think, @klihub, @marquiz ,@kad? Would you agree, and would you prefer including this design change into this same PR? I think it would make this PR simpler to review.

I you feel that it would be easier/the right course of action to cram that change into this same PR, then we should do that.

askervin:
> I you feel that it would be easier/the right course of action to cram that change into this same PR, then we should do that.

Thanks for the assurance, I have now done that. There can be any number of default balloons, if user wishes so. And that's a perfectly valid wish. Default and reserved balloons no more share CPUs. And there is no more assumption that balloons[0] would be the reserved, and balloons[1] the default balloon.

docs/resource-policy/policy/balloons.md
```diff
@@ -99,8 +101,15 @@ Balloons policy parameters:
   value set here is the default for all balloon types, but it can be
   overridden with the balloon type specific setting with the same
   name.
-- `balloonTypes` is a list of balloon type definitions. Each type can
-  be configured with the following parameters:
+- `balloonTypes` is a list of balloon type definitions. The order of
+  the types is significant in balloon CPU allocation and in choosing a
+  balloon type for a new container. The first type with the highest
+  `allocatorPriority` will get CPUs before other balloons when their
+  `minBalloons` > 0. A new container without balloon annotations will
```
marquiz, line 108: Just a note that should we come up with a term for these "minBalloons > 0" balloons. "pre-allocated" or "pre-inflated" or "persistent" or "non-volatile" or "sustained" or smth. So we could discuss the division between the two types of balloons. WDYT?

klihub (approved):

marquiz (approved):
One note about documentation, but not a blocker (LGTM)

docs/resource-policy/policy/balloons.md
klihub: Good point. Perhaps in a follow-up PR.

askervin: @marquiz, I agree.

PR #255, opened 2024-02-06
e2e: fix balloons test CPU frequencies

askervin:
Use kHZ instead of MHz in minFreq and maxFreq as required in the real interface. Using wrong units in the tests did not fail because configuring CPU frequencies with any values does not really work in our test VMs. However, using wrong units in tests is misleading for anyone who tries to use similar configuration on bare metal.

### DIFF
--- a/test/e2e/policies.test-suite/balloons/balloons-config.yaml.in
+++ b/test/e2e/policies.test-suite/balloons/balloons-config.yaml.in
@@ -65,15 +65,15 @@ spec:
     cpu:
       classes:
         default:
-          minFreq: ${CPU_DEFAULT_MIN:-800}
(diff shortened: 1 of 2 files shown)


marquiz (approved):
LGTM

klihub:
@askervin This is a bit late and only tangentially related, but having the unit be allowed in those configuration frequency values would allow the user to specify them in a human-friendly syntax and reduce the possibility for mistakes. If we don't do that, then using parameter names like minFreqkHz and maxFreqkHz would make the settings' unit self-documenting/obvious.

askervin:
> allow the user to specify them [units] in a human-friendly syntax

Definitely something we should do.

PR #259, opened 2024-02-09
balloons: fix logged balloon name.

klihub:
Log balloon name as a string, not integer.


askervin (approved):
Thanks @klihub! LGTM.

PR #263, opened 2024-02-12
build: update to controller-tools 0.14.0.

klihub:
Update controller-tools to 0.14.0. It handles some errors better than simply throwing a panic, leaving the user with little to no hint about what is really wrong. Regenerate CRDs since this version also generates multiline YAML strings differently than 0.12.0.


askervin (approved):
LGTM

marquiz:
But let's wait for #254

PR #264, opened 2024-02-12
pkg/apis: move resmgr apis to pkg/apis/resmgr.

klihub:
Notes:
- this PR is stacked on top of #263

Move resmgr API and types to pkg/apis/resmgr/v1alpha1, next to other generated APIs (currently config). This re-enables automatic updating/generation of these APIs and allows annotating the types for better CRD validation which is useful if we use them in our config CRDs.


askervin (approved):
LGTM

PR #271, opened 2024-02-14
.github: package and publish unstable Helm charts in image registry.

klihub:
This PR publishes unstable Helm Charts by pushing them to the OCI registry, triggered by any push to `main`. The chart version is calculated from the last tag `x.y.z` on `main` then turning it into a chart version `x.y.z+1-unstable`. The image/app version used in the chart is always `unstable`.

[skopeo](https://github.com/containers/skopeo) can be used to list helm charts hosted in an OCI registry. For instance, to list all the available charts for the template policy one can use
`skopeo list-tags docker://ghcr.io/klihub/nri-plugins/helm-charts/nri-resource-policy-template`.

Note that Helm itself is currently unable to search OCI registries, although there is a WiP implementation for Helm OCI registry search here: https://github.com/sabre1041/helm/tree/search-registry. Before this or something similar gets merged, the unstable Helm chart is not discoverable using Helm alone. One has to know both the charts location and its tag to be able to pull or use it. IOW, for instance, with a command like this:
`helm install -n kube-system test oci://ghcr.io/klihub/nri-plugins/helm-charts/nri-resource-policy-template --version v0.4-unstable`

### DIFF
--- a/.github/workflows/package-helm.yaml
+++ b/.github/workflows/package-helm.yaml
@@ -4,12 +4,20 @@ on:
   push:
     tags:
       - "v*.*.*"
+    branches:
+      - main
 
 env:
   CHARTS_DIR: deployment/helm/
+  IS_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') && 'true' || 'false' }}
+  UNSTABLE_CHARTS: unstable-helm-charts
+  REGISTRY: ghcr.io
+  REGISTRY_USER: ${{ github.repository_owner }}
+  REGISTRY_PATH: ${{ github.repository }}
 
 jobs:
   release:
+    if: ${{ github.env.IS_RELEASE == 'true' }}
     permissions:
       contents: write
     runs-on: ubuntu-latest
@@ -20,7 +28,7 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@v4.0.0
 
-      - name: Package Helm charts
+      - name: Package Stable Helm Charts
         run: |
           find "$CHARTS_DIR" -name values.yaml | xargs -I '{}' \
               sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" -i '{}'
@@ -30,10 +38,66 @@ jobs:
             mv $SRC_FILE $DEST_FILE
           done
 
-      - name: Upload Helm packages to GitHub releases
+      - name: Upload Stable Helm Charts to GitHub Release
         uses: softprops/action-gh-release@v1
         with:
           name: ${{ github.ref_name }}
           draft: true
           append_body: true
           files: nri-*helm-chart*.tgz
+
+  unstable:
+    concurrency:
+      group: unstable-helm-charts
+      cancel-in-progress: false
+    if: ${{ github.env.IS_RELEASE != 'true' }}
+    permissions:
+      packages: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Deep Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4.0.0
+
+      - name: Package Unstable Helm Charts
+        id: package-charts
+        run: |
+          # For unstable chart verison we use:
+          #   - chart version: x.y-unstable derived from the latest tag x.y.z
+          #   - image version: 'unstable'.
+          majmin="$(git describe | sed -E 's/v([0-9]*\.[0-9]*).*$/\1/')"
+          CHART_VERSION="${majmin}-unstable"
+          APP_VERSION=unstable
+          # Package charts
+          find "$CHARTS_DIR" -name values.yaml | xargs -I '{}' \
+              sed -e s"/pullPolicy:.*/pullPolicy: Always/" -i '{}'
+          helm package --version "$CHART_VERSION" --app-version $APP_VERSION "$CHARTS_DIR"/*
+          find "$CHARTS_DIR" -name values.yaml | xargs -I '{}' \
+              git checkout '{}'
+          mkdir ../$UNSTABLE_CHARTS
+          find . -name '*.tgz' -print | while read SRC_FILE; do
+            DEST_FILE=$(echo $SRC_FILE | sed 's/v/helm-chart-v/g')
+            mv -v $SRC_FILE ../$UNSTABLE_CHARTS/$DEST_FILE
+          done
+
+      - name: Log In To Registry
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | \
+              helm registry login ${{ env.REGISTRY }}/${{ env.REGISTRY_PATH }} -u ${{ env.REGISTRY_USER }} --password-stdin
+
+      - name: Push Unstable Helm Charts To Registry
+        shell: bash
+        run: |
+          # Notes:
+          #   Currently we only publish unstable Helm charts from main/HEAD.
+          #   We have no active cleanup of old unstable charts in place. In
+          #   between new tags unstable chart have the same version, though.
+          pushd ../$UNSTABLE_CHARTS
+          for i in ./*.tgz; do
+              helm push $i oci://${{ env.REGISTRY }}/${{ env.REGISTRY_PATH }}
+          done
+          popd
--- a/docs/deployment/helm/index.md
+++ b/docs/deployment/helm/index.md
@@ -1,5 +1,7 @@
 # Helm
 
+## Stable Helm Charts
+
 All the available charts can be found in [artifacthub.io](https://artifacthub.io/packages/search?ts_query_web=nri&verified_publisher=true&official=true&sort=relevance&page=1).
 
 **NOTE:** NRI-plugins Helm installation has been successfully verified in both local clusters and major Cloud Providers' managed clusters, including:
@@ -18,6 +20,26 @@ All the available charts can be found in [artifacthub.io](https://artifacthub.io
         - node image: Azure Linux Container Host, Ubuntu 20.04
 
 While Ubuntu 20.04/22.04 was used across all three CSP environments, it's worth noting that node images are not limited to Ubuntu 20.04/22.04 only. The majority of widely recognized Linux distributions should be suitable for use.
+
+## Unstable Helm Charts
+
+Helm charts are also published from the main/development branch after each merge.
+These charts reference the latest development images tagged as `unstable` and are
+are stored alongside plugin images in the OCI image registry.
+
+### Discovering Unstable Helm Charts
+
+Unstable charts can be discovered using [skopeo](https://github.com/containers/skopeo).
+For instance, one can list the available charts for the balloons plugin using this
+skopeo command:
+`skopeo list-tags docker://ghcr.io/containers/nri-plugins/nri-resource-policy-balloons`
+
+### Using Unstable Helm Charts
+
+Once discovered, unstable Helm charts can be used like any other. For instance, to use
+the `$X.$Y-unstable` version of the chart to install the development version of the
+balloons plugin one can use this command:
+`helm install --devel -n kube-system test oci://ghcr.io/containers/nri-plugins/nri-resource-policy-balloons --version $X.$Y-unstable --set image.tag=unstable --set image.pullPolicy=Always`
 
 ```{toctree}
 ---


fmuyassarov:
> @fmuyassarov Would this work ? Would such versioning scheme confuse Helm ? 

If we use `0.0.0-unstable` I think there won't be any confusion/issue from Helm side, because Helm only requires following semantic versioning and `0.0.0-unstable` is valid semantic version.

> Would it consider such a version the highest ? Unfortunately I couldn't find a way in helm package to explicitly mark a chart unstable or a development version.

Yes, there is no way to indicate/mark the package as development package. But 0.0.0-unstable is a good indicator from versioning perspective that it is unstable version and should not be used in production.

I'm also considering whether we could label the charts based on the git tag for each committed push. This approach would allow us to test Helm charts with any version of the code, compared to relying solely on '0.0.0-unstable', which limits us to testing only the latest or current state (HEAD) of the Helm charts. However, one potential challenge with storing charts for every git commit is the space it might require.

fmuyassarov:
> I'm also considering whether we could label the charts based on the git tag for each committed push. This approach would allow us to test Helm charts with any version of the code, compared to relying solely on '0.0.0-unstable', which limits us to testing only the latest or current state (HEAD) of the Helm charts. However, one potential challenge with storing charts for every git commit is the space it might require.

Or we really want that, perhaps we can also come up with some proper retention logic

.github/workflows/package-helm.yaml
```diff
@@ -30,10 +35,79 @@ jobs:
...
           name: ${{ github.ref_name }}
           draft: true
           append_body: true
           files: nri-*helm-chart*.tgz
+
+  unstable:
+    concurrency:
+      group: gh-pages
+      cancel-in-progress: false
+    if: ${{ github.env.IS_RELEASE != 'true' }}
```
fmuyassarov, line 50: I think you can simplify it a little 
```suggestion
    if: ${{ !github.env.IS_RELEASE }}
```

.github/workflows/package-helm.yaml
```diff
@@ -30,10 +35,79 @@ jobs:
...
+    permissions:
+      contents: write
```
fmuyassarov, line 52: to avoid duplication, I would set it at the root level
```

name: Package Helm charts

on:
  push:
    tags:
      - "v*.*.*"
    branches:
      - main
      
permissions:
  contents: write

...
```

fmuyassarov (commented):
Since it is still WIP, I didn't go through the change fully. Just gave it a quick review.

klihub:
> > @fmuyassarov Would this work ? Would such versioning scheme confuse Helm ?
> 
> If we use `0.0.0-unstable` I think there won't be any confusion/issue from Helm side, because Helm only requires following semantic versioning and `0.0.0-unstable` is valid semantic version.
> 
> > Would it consider such a version the highest ? Unfortunately I couldn't find a way in helm package to explicitly mark a chart unstable or a development version.
> 
> Yes, there is no way to indicate/mark the package as development package. But 0.0.0-unstable is a good indicator from versioning perspective that it is unstable version and should not be used in production.
> 
> I'm also considering whether we could label the charts based on the git tag for each committed push. This approach would allow us to test Helm charts with any version of the code, compared to relying solely on '0.0.0-unstable', which limits us to testing only the latest or current state (HEAD) of the Helm charts. However, one potential challenge with storing charts for every git commit is the space it might require.

Well, this current scheme a bit surprisingly seems to work as I hoped/expected. It is live in my personal fork for testing, and if I do

```shell
# helm repo add klihub https://raw.githubusercontent.com/klihub/nri-plugins/gh-pages
# helm search repo -l nri-resource-policy-balloons
NAME                                            CHART VERSION   APP VERSION     DESCRIPTION                                       
nri-plugins/nri-resource-policy-balloons        v0.3.2          v0.3.2          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.3.1          v0.3.1          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.3.0          v0.3.0          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.3          v0.2.3          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.2          v0.2.2          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.1          v0.2.1          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.0          v0.2.0          The balloons NRI resource policy plugin impleme...
# But then if I allow development/unstable packages I get:
[root@crio-devel xfer]# helm search repo --devel -l nri-resource-policy-balloons
NAME                                            CHART VERSION   APP VERSION     DESCRIPTION                                       
klihub/nri-resource-policy-balloons             v0.3.2-unstable unstable        The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.3.2          v0.3.2          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.3.1          v0.3.1          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.3.0          v0.3.0          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.3          v0.2.3          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.2          v0.2.2          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.1          v0.2.1          The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons        v0.2.0          v0.2.0          The balloons NRI resource policy plugin impleme...
```

.github/workflows/package-helm.yaml
klihub: I don't think that would work. I think everything in `github.env`  is unconditionally a `string`, even `'true'` and `'false'`... At least I could not find by a quick search any conditionals that would evaluate `github.env.*` directly as a boolean.

.github/workflows/package-helm.yaml
klihub: Packaging stable charts now require `permissions/contents: write` while packaging and publishing unstable ones require `permissions/packages: write`.

klihub:
> Since it is still WIP, I didn't go through the change fully. Just gave it a quick review.

@fmuyassarov It is not reworked to publish unstable charts to the image registry and marked as ready for review.

.github/workflows/package-helm.yaml
```diff
@@ -30,10 +38,68 @@ jobs:
...
+      - name: Install Helm
+        uses: azure/setup-helm@v4.0.0
+
+      - name: Package Unstable Helm Charts
+        id: package-charts
+        run: |
+          # For unstable chart verison we use:
+          #   - chart version: latest tag+1 patch + '-unstable'
+          #   - image version: 'unstable'.
+          majminpatch="$(git describe | sed -E 's/v([0-9]*\.[0-9]*\.[0-9]*).*$/\1/')"
```
marquiz, line 72: Should we include the `v` here?

```suggestion
          majminpatch="$(git describe | sed -E 's/(v[0-9]*\.[0-9]*\.[0-9]*).*$/\1/')"
```

I.e. make the version `v0.4.0-unstable` instead of `0.4.0-unstable`.

.github/workflows/package-helm.yaml
```diff
@@ -30,10 +38,68 @@ jobs:
...
+          majmin=${majminpatch%.*}
+          patch=${majminpatch##*.}
+          CHART_VERSION="${majmin}.$((patch+1))-unstable"
+          APP_VERSION=unstable
+          # Package charts
+          find "$CHARTS_DIR" -name values.yaml | xargs -I '{}' \
+              sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" -i '{}'
```
marquiz, line 79: I think we should use `Always` pull policy for the unstable images(?)

.github/workflows/package-helm.yaml
```diff
@@ -30,10 +38,68 @@ jobs:
...
+      - name: Package Unstable Helm Charts
+        id: package-charts
+        run: |
+          # For unstable chart verison we use:
+          #   - chart version: latest tag+1 patch + '-unstable'
+          #   - image version: 'unstable'.
+          majminpatch="$(git describe | sed -E 's/v([0-9]*\.[0-9]*\.[0-9]*).*$/\1/')"
+          majmin=${majminpatch%.*}
+          patch=${majminpatch##*.}
+          CHART_VERSION="${majmin}.$((patch+1))-unstable"
```
marquiz, line 75: Hmm, what do we get for the main branch? We get `v0.4.1-unstable` when even `v0.4.0` is not released, yet?

Also, another question I'm pondering whether it would make sense to just have per-majmin or per-release-branch unstable chart versions like `v0.4-unstable`.

I think this would make it cleaner in the future if we enable unstable chart for release branches. E.g. it would keep the registry a bit cleaner and versions like `v0.4.1-unstable` virtually becomes junk/noise when v0.4.1 is out.

docs/deployment/helm/index.md
```diff
@@ -19,6 +21,26 @@ All the available charts can be found in [artifacthub.io](https://artifacthub.io
...
+
+Unstable charts can be discovered using [skopeo](https://github.com/containers/skopeo).
+For instance, one can list the available charts for the balloons plugin using this
+skopeo command:
+`skopeo list-tags docker://ghcr.io/containers/nri-plugins/nri-resource-policy-balloons`
+
+### Using Unstable Helm Charts
+
+Once discovered, unstable Helm charts can be used like any other. For instance, to use
+the `0.3.2-unstable` version of the chart to install the development version of the
```
marquiz, line 40: We don't have unstable for the release branches so maybe we should just refer to the main branch here? I.e.`v0.4.0-unstable`

.github/workflows/package-helm.yaml
klihub: Unfortunately, that's not a valid version for a Helm chart. It needs to be semver-compatible, so it can't start with a 'v'.

.github/workflows/package-helm.yaml
klihub: > Also, another question I'm pondering whether it would make sense to just have per-majmin or per-release-branch unstable chart versions like v0.4-unstable.
>
> I think this would make it cleaner in the future if we enable unstable chart for release branches. E.g. it would keep the registry a bit cleaner and versions like v0.4.1-unstable virtually becomes junk/noise when v0.4.1 is out.

True. It makes much more sense to have per-majmin unstable version.

docs/deployment/helm/index.md
klihub: Yes, but then I guess it would be better to not spell out any particular version. So maybe I'll update it as `$X.$Y-unstable` and trust the user to figure out what that means in practice.

.github/workflows/package-helm.yaml
klihub: Updated to use per-majmin versions.

.github/workflows/package-helm.yaml
klihub: Updated.

.github/workflows/package-helm.yaml
klihub: Although you are right that probably we could use a 'v' in the OCI registry tag and omit it from the version in the chart metadata... I can try it out in practice, and see if Helm is happy with such a setup...

klihub: Hmm, but we don't get to chose the tag or anything. Helm figures all that out by itself, so there is not much I can do here, I think.

marquiz: But wait, how are the stable helm charts versioned starting with v:

```
$ helm search repo nri-plugins/nri-resource-policy-balloons -l
NAME                                    	CHART VERSION	APP VERSION	DESCRIPTION
nri-plugins/nri-resource-policy-balloons	v0.4.1       	v0.4.1     	The balloons NRI resource policy plugin impleme...
nri-plugins/nri-resource-policy-balloons	v0.3.2       	v0.3.2     	The balloons NRI resource policy plugin impleme...
...
```

klihub: Updated to use 'v'-prefixed versions and to push helm charts to `ghcr.io/$org/$repo/helm-charts/$helmchart`.

marquiz (approved):
LGTM

PR #272, opened 2024-02-19
cache: store creation time of pod and containers cache objects

askervin:
Enable sorting containers based on the time when they and their pods have been added to the cache.

### DIFF
--- a/pkg/resmgr/cache/cache.go
+++ b/pkg/resmgr/cache/cache.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	nri "github.com/containerd/nri/pkg/api"
 	v1 "k8s.io/api/core/v1"
@@ -81,6 +82,8 @@ type Pod interface {
 	GetName() string
 	// GetNamespace returns the namespace of the pod.
 	GetNamespace() string
+	// GetCtime returns the creation time of the pod cache object.
+	GetCtime() time.Time
 	// GetQOSClass returns the PodQOSClass of the pod.
 	GetQOSClass() v1.PodQOSClass
 	// GetLabel returns the value of the given label and whether it was found.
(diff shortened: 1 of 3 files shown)


klihub (approved):
LGTM.

On a related note, I can extend NRI and add the CreatedAt and StartedAt timestamps to the protocol. That would allow us to sort existing containers also during synchronization at will.

askervin:
> On a related note, I can extend NRI and add the CreatedAt timestamp to the protocol. That would allow us to sort existing containers also during synchronization at will.

I'm all for it.

Before implementing it this way, I was first wondering if the whole thing should be based on actual creation time of containers in the system. However, for this particular use case (PR #273), the order in which the policy became aware of containers is even more important than their creation times in the system, so I thought that storing cache object creation times is useful anyway.

That said, there can be use cases for real create and start timestamps, too.

klihub:
> > On a related note, I can extend NRI and add the CreatedAt timestamp to the protocol. That would allow us to sort existing containers also during synchronization at will.
> 
> I'm all for it.
> 
> Before implementing it this way, I was first wondering if the whole thing should be based on actual creation time of containers in the system. However, for this particular use case (PR #273), the order in which the policy became aware of containers is even more important than their creation times in the system, so I thought that storing cache object creation times is useful anyway.
> 
> That said, there can be use cases for real create and start timestamps, too.

Referring to our chat in the weekly sync, as a temporary approximation with pod-level granularity you often might be able to tell which of two containers is older by looking at their pods UID, extracting the their Version() and if they are both among {1, 2, 6, 7}, then extracting the Time() fields and comparing them. These fields are then in 100s of nanoseconds since 15th Oct 1582 (according to RFC4122). If the pods have not been scheduled roughly at the same time, this should provide plenty of precision to tell them apart age wise. Probably not a Turing-award winner, but can act as a bridge gap solution if sorting by age is really needed.

PR #273, opened 2024-02-19
balloons: fix the order of assigning containers into balloons

askervin:
This PR is stacked on top of PR #272.


askervin:
This is PR implements a policy-specific order of synchronizing running containers after restarting or reconfiguring the policy.

A possible alternative would be to sort containers by the framework: sort "released" containers from newest to oldest, and "allocated" containers from oldest to newest, and then call `policy.Sync(allocated, released)`.

klihub (approved):

askervin:
@marquiz, would you have time to check this out...

ppalucki:
I can just confirm it works well in my enviornment setup :)

PR #275, opened 2024-02-29
cache, resmgr: add EvalRef(), implement key substitution.

klihub:
Expose evaluation of (potentially joint) pod and container key references using the newly added EvalRef() Evaluable interface. This allows one to evaluate the values of key references which originally were only used in expressions without having to put an expression around such references.

Additionally, introduce Substitute(source string, subject Evaluable) which takes a source string and perform substitution on it using a subject to resolve key references for plain '$key' or well-formed '${key}' references.

@askervin This is what I cooked up for the new balloons feature we discussed today in our face-to-face session. This should allow using shell-like substitutions in the balloons configuration.

### DIFF
--- a/cmd/plugins/topology-aware/policy/mocks_test.go
+++ b/cmd/plugins/topology-aware/policy/mocks_test.go
@@ -335,7 +335,10 @@ func (m *mockContainer) GetEffectiveAnnotation(key string) (string, bool) {
 	}
 	return pod.GetEffectiveAnnotation(key, m.name)
 }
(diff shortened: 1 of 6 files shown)


klihub:
@askervin I'll rewrite substitution with os.Expand...

klihub:
> @askervin I'll rewrite substitution with os.Expand...

Okay, so we can't actually do that. `os.Expand` mimicks simple shell substitution. For instance, it does not allow '/' or '.' in variable names, and we can have both of those in references...

PR #277, opened 2024-03-01
cache, resmgr: expose key evaluation, implement key substitution.

klihub:
Expose evaluating pod and container key references using the newly added `EvalRef()` Evaluable interface. This lets one evaluate values of key references, something which previously was only done as part of evaluating expressions. Also, introduce `Substitute()` which takes a source string and performs key substitution on it wrt. a subject, using the new `EvalRef()`, to resolve key references in plain ('$key') or well-formed ('${key}') notation.

### DIFF
--- a/pkg/apis/resmgr/v1alpha1/expression.go
+++ b/pkg/apis/resmgr/v1alpha1/expression.go
@@ -122,7 +122,7 @@ func (e *Expression) Evaluate(subject Evaluable) bool {
 		return true
 	}
 
-	value, ok := e.KeyValue(subject)
+	value, ok := KeyValue(e.Key, subject)
 	result := false
 
 	switch e.Op {
@@ -173,14 +173,19 @@ func (e *Expression) Evaluate(subject Evaluable) bool {
 	return result
 }
 
-// KeyValue extracts the value of the expresssion key from a container.
-func (e *Expression) KeyValue(subject Evaluable) (string, bool) {
-	log.Debug("looking up %q @ %s...", e.Key, subject)
+// String returns the expression as a string.
+func (e *Expression) String() string {
+	return fmt.Sprintf("<%s %s %s>", e.Key, e.Op, strings.Join(e.Values, ","))
+}
+
+// KeyValue extracts the value of the expression in the scope of the given subject.
+func KeyValue(key string, subject Evaluable) (string, bool) {
+	log.Debug("looking up %q @ %s...", key, subject)
 
 	value := ""
 	ok := false
 
-	keys, vsep := splitKeys(e.Key)
+	keys, vsep := splitKeys(key)
 	if len(keys) == 1 {
 		value, ok, _ = ResolveRef(subject, keys[0])
 	} else {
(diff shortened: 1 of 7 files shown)


pkg/apis/resmgr/v1alpha1/expression.go
```diff
@@ -296,9 +301,97 @@ func ResolveRef(subject Evaluable, spec string) (string, bool, error) {
...
+
+			if err := substKeyValue(); err != nil {
+				return "", err
+			}
+
+			buf.Reset()
+
+		case '{':
+			if curly || buf.Len() != 0 {
+				return "", fmt.Errorf("invalid substitution %q, unexpecetd '{' (at %q)",
```
askervin, line 360: s/unexpecetd/unexpected

askervin (commented):
This looks great. Just a typo-nit.

askervin:
When implementing the groupBy option (PR #278) on top of this, I wondered if we should scope the `Substitute()` function so that it would not be directly in `resmgr/v1alpha1` scope, but under `resmgr/expression` or something similar.  The reason for considering this is that seeing `resmgr.Substitute()` in the code might not associate with expression handling.

But I'm definitely not requesting this change, only asking about your feelings on the location of the expression part of the API. On the other hand, I would not like to complicate the API with unnecessary sub-namespaces/scopes, unless there is really need for it.

klihub:
> When implementing the groupBy option (PR #278) on top of this, I wondered if we should scope the `Substitute()` function so that it would not be directly in `resmgr/v1alpha1` scope, but under `resmgr/expression` or something similar. The reason for considering this is that seeing `resmgr.Substitute()` in the code might not associate with expression handling.

Hmm, would adding a `Substitute(src string, mustResolve bool) (string, error)` convenience function to containers be of any help in this regard ? It would not add the mental association to expressions, but it would be less intrusive (would not require reshuffling the package tree right now), and it would leave resmgrapi out of the function signature/call altogether, maybe making it more intuitive what the function does just by looking at its signature.

> 
> But I'm definitely not requesting this change, only asking about your feelings on the location of the expression part of the API. On the other hand, I would not like to complicate the API with unnecessary sub-namespaces/scopes, unless there is really need for it.

askervin:
> Hmm, would adding a `Substitute(src string, mustResolve bool) (string, error)` convenience function to containers be of any help in this regard ?

Good idea, I think it would. Then it would be next to `ctr.EvalRef()` for evaluating single references, so it would nicely fit there in my mind at least. And as it takes single evaluable as a parameter, it fits fine there.

> maybe making it more intuitive what the function does just by looking at its signature.

Would `Expand(string)` sound descriptive enough?

klihub:
> > Hmm, would adding a `Substitute(src string, mustResolve bool) (string, error)` convenience function to containers be of any help in this regard ?
> 
> Good idea, I think it would. Then it would be next to `ctr.EvalRef()` for evaluating single references, so it would nicely fit there in my mind at least. And as it takes single evaluable as a parameter, it fits fine there.
> 
> > maybe making it more intuitive what the function does just by looking at its signature.
> 
> Would `Expand(string)` sound descriptive enough?

Yes, probably `Expand` is better than `Substitute`. How about the mustResolve argument ? Should we omit it from the container-level convenience wrapper and let unresolved keys always evaluate to `""` ?

klihub:
@askervin Updated, renaming `Substitute()` to `Expand()` and adding convenience wrappers to pods and containers. The wrappers also require a `mustResolve bool` since I recall you used `Substitute()` with `mustResolve = true` in the initial placement and with `mustResolve = false` in later updates.

askervin (approved):
LGTM. This is very nice!

PR #278, opened 2024-03-04
balloons: implement groupBy option

askervin:
groupBy enables grouping containers into balloon instances.

Stacked on top of PR #277.


klihub:
@askervin Can you run locally a `make generate` and squash the resulting changes to the commit which originally updated the balloons configuration API ? I think that would be would be7ee5c3121983b88e3072e52f4190ca254dcdae78/'balloons: add groupBy balloon type option'... That would get us rid of the failure in minimal verification.

klihub (dismissed):

kad (approved):
LGTM

PR #280, opened 2024-03-05
e2e: fix k8s Fedora repo

askervin:
- Automatically detect latest stable k8s release. This becomes necessary because new repository URLs contain the version number.
- Allow user to override k8s version to be installed.


askervin:
PR #279 solves this problem.

PR #282, opened 2024-03-12
topology-aware: configurable allocation priority

klihub:
Add support for configuring the default CPU priority preference and annotating pod-/container-level CPU priority preferences. These determine the preferred priority for CPUs when doing fully or partially exclusive CPU allocation. Priorities are calculated for such allocations and passed on to the CPU allocator which then tries to fulfill these preferences.

With this PR in place, it should be possible on hybrid core architecture systems to configure the policy to prefer allocating energy-efficient cores by default and performance optimized cores to containers annotated with such a preference, or vice versa. Additionally, due to the recently introduced (CPU) cluster-awareness in the CPU allocator, this policy should now do exclusive allocations in a (CPU) cluster-aware fashion, preferring to allocate one or more full CPU clusters  when the requested number of CPUs allows this to happen.

### DIFF
--- a/cmd/plugins/topology-aware/policy/pod-preferences.go
+++ b/cmd/plugins/topology-aware/policy/pod-preferences.go
@@ -42,6 +42,8 @@ const (
 	keyColdStartPreference = "cold-start"
 	// annotation key for reserved pools
 	keyReservedCPUsPreference = "prefer-reserved-cpus"
+	// annotation key for CPU Priority preference
+	keyCpuPriorityPreference = "prefer-cpu-priority"
 
 	// effective annotation key for isolated CPU preference
 	preferIsolatedCPUsKey = keyIsolationPreference + "." + kubernetes.ResmgrKeyNamespace
@@ -53,6 +55,8 @@ const (
 	preferColdStartKey = keyColdStartPreference + "." + kubernetes.ResmgrKeyNamespace
 	// annotation key for reserved pools
 	preferReservedCPUsKey = keyReservedCPUsPreference + "." + kubernetes.ResmgrKeyNamespace
+	// effective annotation key for CPU priority preference
+	preferCpuPriorityKey = keyCpuPriorityPreference + "." + kubernetes.ResmgrKeyNamespace
 )
 
 // cpuClass is a type of CPU to allocate
@@ -151,6 +155,36 @@ func sharedCPUsPreference(pod cache.Pod, container cache.Container) (bool, bool)
 	log.Debug("%s: effective shared CPU preference %v", container.PrettyName(), preference)
 
 	return preference, true
+}
+
+// cpuPrioPreference returns the CPU priority preference for the given container
+// and whether the container was explicitly annotated with this setting.
+func cpuPrioPreference(pod cache.Pod, container cache.Container, fallback cpuPrio) cpuPrio {
+	key := preferCpuPriorityKey
+	value, ok := pod.GetEffectiveAnnotation(key, container.GetName())
+
+	if !ok {
+		prio := fallback
+		log.Debug("%s: implicit CPU priority preference %q", container.PrettyName(), prio)
+		return prio
+	}
+
+	if value == "default" {
+		prio := defaultPrio
+		log.Debug("%s: explicit default CPU priority preference %q", container.PrettyName(), prio)
+		return prio
+	}
+
+	prio, ok := cpuPrioByName[value]
+	if !ok {
+		log.Error("%s: invalid CPU priority preference %q", container.PrettyName(), value)
+		prio := fallback
+		log.Debug("%s: implicit CPU priority preference %q", container.PrettyName(), prio)
+		return prio
+	}
+
+	log.Debug("%s: explicit CPU priority preference %q", container.PrettyName(), prio)
+	return prio
 }
 
 // memoryTypePreference returns what type of memory should be allocated for the container.
@@ -370,7 +404,8 @@ func checkReservedCPUsAnnotations(c cache.Container) (bool, bool) {
 // 2. fraction: amount of fractional CPU in milli-CPU
 // 3. isolate: (bool) whether to prefer isolated full CPUs
 // 4. cpuType: (cpuClass) class of CPU to allocate (reserved vs. normal)
-func cpuAllocationPreferences(pod cache.Pod, container cache.Container) (int, int, bool, cpuClass) {
+// 5. cpuPrio: preferred CPU allocator priority for CPU allocation.
+func cpuAllocationPreferences(pod cache.Pod, container cache.Container) (int, int, bool, cpuClass, cpuPrio) {
 	//
 	// CPU allocation preferences for a container consist of
 	//
@@ -439,20 +474,21 @@ func cpuAllocationPreferences(pod cache.Pod, container cache.Container) (int, in
 	request := reqs.Requests[corev1.ResourceCPU]
 	qosClass := pod.GetQOSClass()
 	fraction := int(request.MilliValue())
+	prio := defaultPrio // ignored for fractional allocations
 
 	// easy cases: kube-system namespace, Burstable or BestEffort QoS class containers
 	preferReserved, explicitReservation := checkReservedCPUsAnnotations(container)
 	switch {
 	case container.PreserveCpuResources():
-		return 0, fraction, false, cpuPreserve
+		return 0, fraction, false, cpuPreserve, prio
 	case preferReserved == true:
-		return 0, fraction, false, cpuReserved
+		return 0, fraction, false, cpuReserved, prio
 	case checkReservedPoolNamespaces(namespace) && !explicitReservation:
-		return 0, fraction, false, cpuReserved
+		return 0, fraction, false, cpuReserved, prio
 	case qosClass == corev1.PodQOSBurstable:
-		return 0, fraction, false, cpuNormal
+		return 0, fraction, false, cpuNormal, prio
 	case qosClass == corev1.PodQOSBestEffort:
-		return 0, 0, false, cpuNormal
+		return 0, 0, false, cpuNormal, prio
 	}
 
 	// complex case: Guaranteed QoS class containers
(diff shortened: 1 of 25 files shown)


pkg/apis/config/v1alpha1/resmgr/policy/topologyaware/config.go
```diff
@@ -77,4 +101,12 @@ type Config struct {
...
 	// +kubebuilder:validation:Required
 	ReservedResources Constraints `json:"reservedResources"`
+	// DefaultCPUPriority (high, normal, low, none)
+	// This parameter is passed to CPU allocator when allocating CPUs exclusively.
+	// If a container is not annotated otherwise, this is the requested priority
+	// from the allocator.
+	// +kubebuilder:validation:Enum=high;normal;low;none
+	// +kubebuilder:default=none
+	// +kubebuilder:validation:Format:string
+	DefaultCPUPriority CPUPriority `json:"defaultCPUPriority,omitempty"`
```
fmuyassarov, line 111: nit: This field can have 4 different values out of which `none`  is considered as the default. Should not it rather be called `CPUPriority`?

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -153,6 +157,36 @@ func sharedCPUsPreference(pod cache.Pod, container cache.Container) (bool, bool)
 	return preference, true
 }
 
+// cpuPrioPreference returns the CPU priority preference for the given container
+// and whether the container was explicitly annotated with this setting.
+func cpuPrioPreference(pod cache.Pod, container cache.Container, fallback cpuPrio) (cpuPrio, bool) {
```
fmuyassarov, line 162: what's the use of the bool return value here? I see that you are ignoring the second return value when calling `cpuPrioPreference`, so I am wondering if we should rather drop it.

pkg/apis/config/v1alpha1/resmgr/policy/topologyaware/config.go
klihub: But it is not the CPU priority. It is the default CPU priority, which is used if the user has not annotated the container otherwise (and if the policy does not decide otherwise during its internal classification of the container/workload). But at least I need to update the comment to state this in more clear terms.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Good point, and I will drop it if in the end of my still ongoing iteration it still not being used. 

But the idea with the second bool is that it tells whether the preference was explicit (indicated by the user) or implicit. I am considering a change so that if the preference is implicit, and there is no default priority set y the user (or it is set to none), then policy would ask for HighPrio CPUs, if an exclusive allocation request is for a single core or two cores on a hyperthreaded system.

config/crd/bases/config.nri_topologyawarepolicies.yaml
```diff
@@ -94,6 +94,19 @@ spec:
                     - classes
                     type: object
                 type: object
+              defaultCPUPriority:
+                default: none
+                description: |-
+                  DefaultCPUPriority (high, normal, low, none)
+                  This parameter is passed to CPU allocator when allocating CPUs exclusively.
```
kad, line 101: Wait, so is it only for exclusive allocation? should it be then reflected in the name? From UX perspective, it should be a valid case for container request "low" priority cpus even from shared pool(s). e.g. E-cores on hybrids.

klihub: That is a good question. This PR does not implement low-priority shared CPU allocation. 

Currently I don't know if I will be able to stretch the limits of the T-A policy enough to implement that without breaking stuff left and right. However, if it turns out to be possible with a reasonable effort, then I'd like to have a single configuration preference and a single annotation for controlling both exclusive and shared CPU allocation in this regard (both to keep configuration simpler and to avoid breaking configuration compatibility).

That is why I chose to reflect the semantics in the description of the configuration option and not in the name.

kad: then probably better to rephrase comment and docs for this option that it is **currently** affecting only exclusive allocation. might change in future.

klihub: Updated.

pkg/apis/config/v1alpha1/resmgr/policy/topologyaware/config.go
klihub: Updated the description to better reflect the above description.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Updated.

deployment/helm/balloons/values.yaml
```diff
@@ -33,6 +33,11 @@ config:
 
 # configGroupLabel: config.nri/group
 
+# Extra environment variables to inject.
+# extraEnv:
```
marquiz, line 37: I think these should be documented in the Helm READMEs(?)

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -1105,6 +1138,33 @@ func (cs *supply) GetScore(req Request) Score {
 
 		// calculate fractional capacity
 		score.shared -= part
+
+		ecores := cs.GetNode().System().CoreKindCPUs(sysfs.EfficientCore)
+		lpCPUs := ecores
+		if ecores.Size() == 0 {
```
marquiz, line 1144:
```suggestion
		lpCPUs := cs.GetNode().System().CoreKindCPUs(sysfs.EfficientCore)
		if lpCPUs.Size() == 0 {
```

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -1105,6 +1138,33 @@ func (cs *supply) GetScore(req Request) Score {
...
+		if pcores.Size() == 0 {
+			hpCPUs = cs.GetNode().Policy().cpuAllocator.GetCPUPriorities()[highPrio]
+		}
+		hpCPUs = hpCPUs.Intersection(cs.SharableCPUs())
+		hpCnt := hpCPUs.Size()
+		score.prio[highPrio] = hpCnt*1000 - (1000*full + part)
+
+		npCPUs := pcores
+		if pcores.Size() == 0 {
+			npCPUs = cs.GetNode().Policy().cpuAllocator.GetCPUPriorities()[highPrio]
```
marquiz, line 1163: Should we get the normal instead?

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -1105,6 +1138,33 @@ func (cs *supply) GetScore(req Request) Score {
...
+		}
+		npCPUs = npCPUs.Intersection(cs.SharableCPUs())
+		npCnt := npCPUs.Size()
+		score.prio[normalPrio] = npCnt*1000 - (1000*full + part)
```
marquiz, line 1167: Looks like we get the exact same calculation and result for hpCPUs and npCPUs 🤨 Is that the intention (and if so, we don't need to calculate it twice)

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -370,7 +404,7 @@ func checkReservedCPUsAnnotations(c cache.Container) (bool, bool) {
 // 2. fraction: amount of fractional CPU in milli-CPU
 // 3. isolate: (bool) whether to prefer isolated full CPUs
 // 4. cpuType: (cpuClass) class of CPU to allocate (reserved vs. normal)
```
askervin, line 406: Comment for the 5th return value.
```
// 5. cpuPrio: CPU allocator priority for the allocation.

File: cmd/plugins/topology-aware/policy/pools.go [hunk 9]
```diff
@@ -960,7 +960,38 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
...
+		lp1, lp2 := score1.PrioCapacity(lowPrio), score2.PrioCapacity(lowPrio)
+		log.Debug("  - lp1 %d vs. lp2 %d", lp1, lp2)
+		switch {
+		case lp1 == lp2:
+			log.Debug("  - low-prio CPU capacity is a TIE")
+		case lp1 >= 0 && lp2 < 0:
+			log.Debug("  => %s WINS based on low-prio CPU capacity", node1.Name())
+			return true
+		case lp1 < 0 && lp2 >= 0:
+			log.Debug("  => %s WINS based on low-prio CPU capacity", node1.Name())
```
askervin, line 976: `node2.Name()`

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -960,7 +960,38 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
...
+	case highPrio:
+		hp1, hp2 := score1.PrioCapacity(highPrio), score2.PrioCapacity(highPrio)
+		switch {
+		case hp1 == hp2:
+			log.Debug("  - HighPrio CPU capacity is a TIE")
+		case hp1 >= 0 && hp2 < 0:
+			log.Debug("  => %s WINS based on high-prio CPU capacity", node1.Name())
+			return true
+		case hp1 < 0 && hp2 >= 0:
+			log.Debug("  => %s WINS based on high-prio CPU capacity", node1.Name())
```
askervin, line 989: node2.Name()

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -1001,7 +1032,23 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
...
+		if request.CPUPrio() == normalPrio {
+			np1, np2 := score1.PrioCapacity(normalPrio), score2.PrioCapacity(normalPrio)
+			switch {
+			case np1 == np2:
+				log.Debug("  - normal-prio CPU capacity is a TIE")
+			case np1 >= 0 && np2 < 0:
+				log.Debug("  => %s WINS based on normal-prio CPU capacity", node1.Name())
+				return true
+			case np1 < 0 && np2 >= 0:
+				log.Debug("  => %s WINS based on normal-prio capacity", node1.Name())
```
askervin, line 1046: node2.Name()

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -998,19 +1026,19 @@ func (cr *request) String() string {
 	isolated := map[bool]string{false: "", true: "isolated "}[cr.isolate]
 	switch {
 	case cr.full == 0 && cr.fraction == 0:
-		return fmt.Sprintf("<CPU request "+cr.container.PrettyName()+": ->") + mem
+		return fmt.Sprintf("<%s CPU request "+cr.container.PrettyName()+": ->", cr.prio) + mem
```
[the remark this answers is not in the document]
klihub: Hmm... maybe. If any of the concatenated strings could come from an untrusted/external source (so it could contain formatting verbs) then it would be a bad idea to use both concatenation and `Sprintf`. It's not the case here, but I'll update it anyway (and all the other cases in the same function) to only use `Sprintf()` without concatenation.

cmd/plugins/topology-aware/policy/resources.go
klihub: Yes, for normal priority I think we should.

cmd/plugins/topology-aware/policy/resources.go
klihub: Updated to use normal priority instead of trying P-cores for the normal prio request.

askervin (approved):
LGTM

marquiz (approved):
Nice job @klihub 🚀

LGTM

PR #286, opened 2024-03-15
pkg/topology: exclude sysRoot from hint provider / fix broken unit tests.

klihub:


pkg/topology/topology.go
```diff
@@ -104,7 +105,11 @@ func getDevicesFromVirtual(realDevPath string) (devs []string, err error) {
 func getTopologyHint(sysFSPath string) (*Hint, error) {
 	log.Debugf("getting topology hint for %s", sysFSPath)
 
-	hint := Hint{Provider: sysFSPath}
+	plainPath := sysFSPath
+	if sysRoot != "" {
+		plainPath = path.Join("/", strings.TrimPrefix(plainPath, sysRoot))
```
askervin, line 110: `path.Join()` gets only single string argument to join with "/". No need to call it?

klihub: Hmm. Dosen't `strings.TrimPrefix()` remove the leading '/' when we trim `sysRoot` ?

marquiz: I think it's good practice to use `Join()` as it's Clean()s the resulting path. A nano-scale philosophical nit that maybe it would be more correct to use `filepath.Join`

marquiz (approved):
LGTM

pkg/topology/topology.go
klihub: Updated.

marquiz: My nit was really meaningless. I've also been bothered (and inconsistent) by that "duality" myself

marquiz (commented):
LGTM

PTAL @askervin

marquiz (approved):

askervin (approved):
LGTM

PR #287, opened 2024-03-15
memtierd: update the nri-memtierd plugin to use memtierd v0.1.1

askervin:
memtierd requires to be built with go 1.22. This patch changes nri-memtierd build not to follow GO_VERSION of all the other plugins. Other options include downgrading memtierd required go version or upgrading GO_VERSION of all nri-plugins.


klihub (dismissed):

kad (approved):
LGTM

PR #290, opened 2024-03-15
.github: install exact golang version if necessary.

klihub:
When building and publishing documentation, install golang toolchain from release tarball if necessary.


askervin (approved):
LGTM

PR #294, opened 2024-03-18
sysfs: dump system discovery results in a more predictable order.

klihub:
Dump system discovery results in a predictable order by looping through packages, nodes and CPU cores by sorted ID order.


pfl (approved):
LGTM

fmuyassarov (approved):
LGTM

PR #295, opened 2024-03-18
sysfs,cpuallocator: add support for hybrid core discovery, preferred allocation.

klihub:
Notes: This PR is stacked on top of #294 and #206.

This patch series adds support for discovering CPU cores of various kinds on hybrid core architecture systems. It also updates the CPU allocator to make better allocation choices on such systems, both with and without explicitly provided allocation preferences.


pfl (approved):

fmuyassarov (approved):

fmuyassarov:
@klihub can you please rebase this one ?

klihub:
> @klihub can you please rebase this one ?

@fmuyassarov Rebased on latest main.

pkg/sysfs/system.go
```diff
@@ -730,6 +750,16 @@ func (sys *system) discoverCPUs() error {
 		return err
 	}
 
+	// If we had partial core type overrides while discovering CPUs, some cores
+	// did not get their type set correctly. Fix that now.
+	for kind, ids := range sys.coreKindCPUs {
```
marquiz, line 755: Should we do this "hack" only if overrides were used. I.e. do this before the detection from sysfs and not tamper with what was detected from there?

pkg/cpuallocator/allocator.go
```diff
@@ -943,6 +966,21 @@ func (c *cpuPriorities) cmpCPUSet(csetA, csetB cpuset.CPUSet, prefer CPUPriority
...
+	if cpuCnt > 0 && prefer == PriorityLow {
+		prefA := csetA.Intersection(c[prefer]).Size()
+		prefB := csetB.Intersection(c[prefer]).Size()
+		if prefA >= cpuCnt || prefB >= cpuCnt {
+			if prefB == 0 {
+				return -1
+			}
+			if prefA == 0 {
+				return 1
+			}
```
marquiz, line 997: What do we need this special cases for? Why would the abs(return value) need to be 1 in theses cases?

pkg/cpuallocator/allocator.go
```diff
@@ -981,6 +981,25 @@ func (c *cpuPriorities) cmpCPUSet(csetA, csetB cpuset.CPUSet, prefer CPUPriority
...
+			prefA = csetA.Intersection(c[PriorityNormal]).Size()
+			prefB = csetB.Intersection(c[PriorityNormal]).Size()
+		}
+		if prefA >= cpuCnt || prefB >= cpuCnt {
+			if prefB == 0 {
+				return -1
+			}
+			if prefA == 0 {
+				return 1
+			}
```
marquiz, line 1016: Same here

marquiz (commented):
Thanks @klihub. Looks pretty good

One nit about one of the commit messages: s/sorring/sorting/

pkg/sysfs/system.go
klihub: We can't do this before `checkCoreKinds()` is done as that might patch up the per kind cpusets for partial overrides. We could do this as part of `checkCoreKinds()` though, although it would be a bit more convoluted because of multiple successful code paths.

pkg/cpuallocator/allocator.go
klihub: Uhmm... if either one is zero (set has no proper CPUs) we need to favor the other one. But in all other cases we favor the tighter fit (the one with fewer but sufficient number of proper CPUs), not the loser one. But there was a snafu there (the '== 0' special cases should have been '< cpuCnt' ones) if the whole test was to be written like that. Anyway, I rewrote it now to be hopefully both correct and easier to understand.

pkg/cpuallocator/allocator.go
klihub: Ditto.

klihub:
> Thanks @klihub. Looks pretty good
> 
> One nit about one of the commit messages: s/sorring/sorting/

Thanks for spotting it. Fixed.

pkg/sysfs/system.go
marquiz: > We can't do this before `checkCoreKinds()` is done as that might patch up the per kind cpusets for partial overrides.

K, got that

> We could do this as part of `checkCoreKinds()` though, although 

Yeah, it feels like the right place for this as it's doing other fixes as well

> it would be a bit more convoluted because of multiple successful code paths.

Are there really? There is this one `if len(sys.coreKindCPUs) == 0` but that branch shouldn't need any patching?

klihub: Hmm, that is true. Lemme check if I can fold the fixup loop into `checkCoreKinds()`.

pkg/cpuallocator/allocator.go
```diff
@@ -943,6 +984,40 @@ func (c *cpuPriorities) cmpCPUSet(csetA, csetB cpuset.CPUSet, prefer CPUPriority
 		return 0
 	}
 
+	// For low prio request, favor cpuset with the tightest fit.
+	if cpuCnt > 0 && prefer == PriorityLow {
+		prefA := csetA.Intersection(c[prefer]).Size()
+		prefB := csetB.Intersection(c[prefer]).Size()
+		if prefA >= cpuCnt && prefB >= cpuCnt {
+			return prefA - prefB
```
askervin, line 992: In order to prefer the tightest fit, line 992 should prefer A (in other words return > 0) if prefA < prefB. But now `prefA - prefB` works the other way around: it returns > 0 when prefA > prefB. If I'm not mistaken, line 992 should be `return prefB - prefA` similartly to line 1028.

pkg/cpuallocator/allocator.go
```diff
@@ -943,6 +984,40 @@ func (c *cpuPriorities) cmpCPUSet(csetA, csetB cpuset.CPUSet, prefer CPUPriority
...
+		}
+		if prefA >= cpuCnt {
+			return -1
```
askervin, line 995: `return 1` in order to prefer cpuset A. (Comment on line 979.)

pkg/cpuallocator/allocator.go
```diff
@@ -943,6 +984,40 @@ func (c *cpuPriorities) cmpCPUSet(csetA, csetB cpuset.CPUSet, prefer CPUPriority
...
+		}
+		if prefB >= cpuCnt {
+			return 1
```
askervin, line 998: `return -1` in order to prefer cpuset B.

pkg/cpuallocator/allocator.go
```diff
@@ -943,6 +984,40 @@ func (c *cpuPriorities) cmpCPUSet(csetA, csetB cpuset.CPUSet, prefer CPUPriority
...
+		prefB := csetB.Intersection(c[prefer]).Size()
+		if prefA == 0 && prefB == 0 {
+			prefA = csetA.Intersection(c[PriorityNormal]).Size()
+			prefB = csetB.Intersection(c[PriorityNormal]).Size()
+		}
+		if prefA >= cpuCnt && prefB >= cpuCnt {
+			return prefA - prefB
+		}
+		if prefA >= cpuCnt {
+			return -1
```
askervin, line 1014: `return 1` to prefer A

pkg/cpuallocator/allocator.go
```diff
@@ -943,6 +984,40 @@ func (c *cpuPriorities) cmpCPUSet(csetA, csetB cpuset.CPUSet, prefer CPUPriority
...
+		}
+		if prefB >= cpuCnt {
+			return 1
```
askervin, line 1017: `return -1` to prefer B

pkg/sysfs/system.go
klihub: Okay, I moved all verification and fixup into `checkCoreKinds()`.

askervin (commented):
Need to double-check CPU set comparison

pkg/cpuallocator/allocator.go
marquiz, line 992: Wouldn't your orignal (or the like) code work just as expected?

```suggestion
		if prefA >= cpuCnt || prefB >= cpuCnt {
			return prefA - prefB
```

klihub: Damn I got really confused here, making a mental association to `slices.SortFunc()` somehow... probably due to the return type.

Anyway, @marquiz no, that wouldn't work correctly. We need to take the smaller one which is larger than cpuCnt, keeping in mind that we need to return > 0 to prefer A and < 0 to prefer B.

marquiz: Well, then `return prefA - prefB`?

It should work. If BOTH `prefA >= cpuCnt` AND `prefB >= cpuCnt` then we just return prefA-prefB as you did.

If EITHER `prefA >= cpuCnt` OR `prefB >= cpuCnt` then we return either positive or negative number (because one of those is greater than the other), just like in your code. The value doesn't matter. It doesn't need to be -1 or 1.

Or is there something I'm missing here 🤔

klihub: Hmm, wouldn't it work incorrectly, for instance, in this case:
cpuCnt: 2
prefA: 2
prefB: 3
prefA - prefB = 2 - 3 = -1 => we prefer B but we should prefer A as the tighter satisfying fit.

So we have to return prefB - prefA to work correctly, but only if both are greater than cpuCnt. Otherwise, wouldn't,for instance, this work incorrectly:
cpuCnt: 2
prefA: 2
prefB: 1
prefB - prefA = -1 => we prefer B, but we should prefer A, since B has insufficient CPUs.

marquiz: Argh, yeah you're right 🙈 Depending on the precondition we choose either the bigger or the smaller set, like

```golang

		if prefA >= cpuCnt && prefB >= cpuCnt {
			// Both sets can satisfy the req -> choose the smaller (tighter fit)
			return prefA - prefB
		} else if prefA >= cpuCnt || prefB >= cpuCnt {
			// Only one set is suitable -> we need to pick it (the bigger)
			return prefB - prefA
		}
```

Maybe this could be documented more explicitly in the code comment to avoid future confusion (at least from my part) (?)

klihub: Hmm... but the `return prefB - prefA` vs. `return prefA - prefB` should be the other way around above...

klihub: Updated along those lines and with comments to that effect.

marquiz (approved):
Thanks @klihub for the tireless work on this one. Now even I can understand how it works 😸 At the same time wondering when we reach the point that nobody understands how the cpuallocator actually works...

LGTM

pkg/cpuallocator/allocator.go
marquiz: Ah yes, I had those wrong :/ Anyway, thanks for the update, easier to read for me, now

marquiz:
@askervin PTAL. When this is merged let's rebase #282 and dive into it

askervin (approved):
LGTM

PR #296, opened 2024-03-20
e2e: pull busybox from quay.io instead of dockerhub.

klihub:
Pull busybox from quay.io/prometheus to avoid dockerhub rate limiting pull errors causing false test failures.


marquiz (approved):
Dockerhub 😤

PR #298, opened 2024-03-21
scripts/testing: add a simple e2e-runner script.

klihub:
Add a simple script which can be used to periodically run e2e tests, for instance from a cron job. The script simply clones or fetches a remote/branch, runs a given set of tests, then collects and publishes the results in a directory.

### DIFF
--- a/scripts/testing/e2e-runner
+++ b/scripts/testing/e2e-runner
@@ -1,0 +1,393 @@
+#!/bin/bash
+
+set -o pipefail
+
+#
+# Default configuration, overridable on the command line.
+#
+
+# RESULT_ROOT:   directory for publishing results
+# RESULT_NAME:   subdirectory for publishing results of this test run
+# SKIP_WORKTREE: skip preparing worktree, assume it is already done
+# E2E_TESTS:     e2e tests to run
+
+# REMOTE_REPO: remote repository to clone
+REMOTE_REPO="${REMOTE_REPO:-https://github.com/containers/nri-plugins}"
+# CLONED_REPO: local location to clone to
+CLONED_REPO="${CLONED_REPO:-/opt/e2e-test/nri-plugins/nri-plugins}"
+# TEST_BRANCH: branch to clone and test
+TEST_BRANCH="${TEST_BRANCH:-main}"
+# RESULT_ROOT: directory for publishing each run's results under
+RESULT_ROOT="${RESULT_ROOT:-/opt/e2e-test/nri-plugins/results}"
+# RESULT_NAME: name of new test run
+RESULT_NAME="${RESULT_NAME:-$(date +%Y-%m-%d-%H-%M)}"
+# PROXIES_SRC: bash-sourcable file to set up proxies in the environment
+PROXIES_SRC=""
+# SKIP_WORKTREE: if set to non-empty, assume worktree already exists
+SKIP_WORKTREE="${SKIP_WORKTREE:-}"
+# E2E_TESTS: end-to-end tests to run instead of full test set
+E2E_TESTS="${E2E_TESTS:-}"
+
+# This script.
+SELF="$0"
+TEST_SUITE_DIR=policies.test-suite
+GIT_SHA1="git.sha1"
+GIT_DESC="git.describe"
+
+
+#
+# Derived configuration, set up after command line parsing
+#
+# RESULT_DIR: directory for publishing results of this test run
+# WORKTREE:   git worktree for this test run
+# LOG:        log file for this test run
+
+RESULT_DIR=""
+LOG=""
+WORKTREE=""
+
+usage() {
+    local exit_code="${exit_code:-0}"
+
+    echo "usage: ${1:-$SELF} [options] [<e2e-tests>]"
+    echo
+    echo "The possible options are:"
+    echo "  --source-proxies <file>   source proxy environment variables from file"
+    echo "  --remote|r <repo>         remote repository to clone/fetch from"
+    echo "  --branch|-b <branch>      name of the remote branch to test"
+    echo "  --local|-l <dir>          local directory to clone remote respository into"
+    echo "  --results|--out|-o <dir>  top-level directory for publishing results"
+    echo "  --name|-n <name>          subdirectory for results of this run"
+    echo "  --skip-worktree           skip worktree preparation, assume already done"
+    echo "  --trace|-t|-x)            run with set -x"
+    echo "  --help                    show this help"
+
+    exit $exit_code
+}
+
+info() {
+    echo 1>&2 "$@"
+}
+
+error() {
+    echo 1>&2 "error: $@"
+}
+
+fatal() {
+    echo 1>&2 "fatal error: $@"
+    exit 1
+}
+
+must-cd() {
+    if ! cd "$1"; then
+        fatal "failed to change directory to $1"
+    fi
+}
+
+must-mkdir() {
+    local dir="$1"; shift
+    if ! mkdir -p "$dir"; then
+        fatal "failed to create ${*:+$* }directory $dir"
+    fi
+}
+
+must-mv() {
+    if ! mv $*; then
+        fatal "mv $* failed"
+    fi
+}
+
+source_proxies() {
+    local var
+
+    if [ -z "$PROXIES_SRC" ]; then
+        info "no proxies to set up..."
+        return 0
+    fi
+
+    info "setting up proxies..."
+    if ! source "$PROXIES_SRC"; then
+        fatal "failed to source proxies from $src"
+    fi
+
+    for var in http_proxy https_proxy no_proxy HTTP_PROXY HTTPS_PROXY NO_PROXY; do
+        if [ -n "${!var}" ]; then
+            info "exporting proxy envrionment variable $var (with value ${!var})..."
+            export $var
+        fi
+    done
+}
+
+prepare_worktree() {
+    local worktree_dir="$(dirname $CLONED_REPO)/worktree"
+
+    WORKTREE="$worktree_dir/$RESULT_NAME"
+    case "$SKIP_WORKTREE" in
+        1|true|yes|y)
+            WORKTREE="$(realpath $WORKTREE)"
+            must-cd "$WORKTREE"
+            return 0
+            ;;
+    esac
+
+    if [ ! -e "$CLONED_REPO" ]; then
+        if ! git clone "$REMOTE_REPO" "$LOCAL_REPO"; then
+            fatal "failed to clone $REMOTE_REPO to $CLONED_REPO"
+        fi
+        must-cd "$LOCAL_REPO"
+    else
+        must-cd "$CLONED_REPO"
+        if ! git remote -v | grep origin | grep -q "$REMOTE_REPO"; then
+            fatal "existing clone $CLONED_REPO has other origin than $REMOTE_REPO"
+        fi
+        if ! git fetch origin; then
+            fatal "failed to fetch $REMOTE_REPO"
+        fi
+    fi
+
+    WORKTREE="$(realpath $WORKTREE)"
+    info "setting up git worktree $WORKTREE..."
+    if ! git worktree add "$WORKTREE" "origin/$TEST_BRANCH"; then
+        fatal "failed to add git worktree $WORKTREE"
+    fi
+
+    if [ -x "$WORKTREE/scripts/testing/e2e-runner" ]; then
+        exec "$WORKTREE/scripts/testing/e2e-runner" \
+             --skip-worktree \
+             ${PROXIES_SRC:+--source-proxies }${PROXIES_SRC:-} \
+             --remote "$REMOTE_REPO" \
+             --branch "$TEST_BRANCH" \
+             --local "$CLONED_REPO" \
+             --results "$RESULT_ROOT" \
+             --name "$RESULT_NAME" \
+             "$E2E_TESTS"
+    else
+        must-cd "$WORKTREE"
+    fi
+}
+
+prepare_result_dir() {
+    local dir="$RESULT_DIR"
+    local log="$dir/${SELF##*/}.log"
+
+    info "creating result dir $dir..."
+    must-mkdir "$dir" "result"
+
+    if ! touch "$log"; then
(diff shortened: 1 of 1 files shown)


scripts/testing/e2e-runner
```diff
@@ -0,0 +1,348 @@
...
+            fi
+            return 0
+            ;;
+    esac
+
+    if [ ! -e "$CLONED_REPO" ]; then
+        if ! mkdir -p $CLONED_REPO; then
+            fatal "failed to create directory $CLONED_REPO"
+        fi
+        cd $CLONED_REPO
```
askervin, line 111: probably good to check exit status of `cd` just like mkdir and git clone, too, and bail out in case of error. if someone else has created /tmp/repo and this user has no right to change into it, the script will continue and clone the repo in some surprising place.

the same applies to other places, too. should we have `cd-or-die()` or `must-cd()` or something to keep this short?

possibly also good idea double-quote $CLONED_REPO and others in commands, as at least shellcheck would complain about them.

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,348 @@
...
+    fi
+}
+
+cleanup_test_vms() {
+    local src
+
+    info "cleaning up test VMs..."
+    cd $WORKTREE
+    for src in test/e2e/n[0-9]*c[0-9]*-c*; do
+        ([ -d "$src" ] || exit 0; cd $src; vagrant destroy --force)
```
askervin, line 188: let's not destroy if `cd` fails... parenthesis could be simplified into
```
(cd "$src" && vagrant destroy --false)
```

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,348 @@
...
+    done
+}
+
+collect_results() {
+    local src dst
+
+    info "collecting test results..."
+    cd $WORKTREE
+    for src in test/e2e/n[0-9]*c[0-9]*-c*; do
+        dst=$RESULT_DIR/${i#test/e2e/}
```
askervin, line 198:
```
${i#test/e2e/} -> ${src#test/e2e/}
```

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,348 @@
...
+        if ! mkdir -p $dst; then
+            fatal "failed to create result directory $dst"
+        fi
+        mv $src/"$TEST_SUITE_DIR" $dst
```
askervin, line 202: maybe an error message or bail out, if cannot move (that is write) into $dst?

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,348 @@
...
+        esac
+done
+
+E2E_TESTS="$*"
+RESULT_DIR="$RESULT_ROOT/$RESULT_NAME"
+
+source_proxies
+prepare_worktree
+prepare_result_dir
+run_e2e_tests
```
askervin, line 342: should we try to cleanup vms also before run_e2e_tests to make sure we have a clean start every time?

askervin (commented):
Great to have this running soon! Thanks @klihub. A couple of comments.

scripts/testing/e2e-runner
klihub: Updated. Introduced/used `must-{cd,mkdir,mv}`, and quoted variable instances in most places.

scripts/testing/e2e-runner
klihub: Updated.

scripts/testing/e2e-runner
klihub: Fixed.

scripts/testing/e2e-runner
klihub: Added `must-mv` and switched to using it.

askervin (approved):
LGTM!

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,357 @@
...
+    fi
+}
+
+cleanup_test_vms() {
+    local src
+
+    info "cleaning up test VMs..."
+    must-cd "$WORKTREE"
+    for src in test/e2e/n[0-9]*c[0-9]*-c*; do
+        (cd "$src" && vagrant destroy --force)
```
fmuyassarov, line 199: Do we also need delete the directory that is created as a result of running e2e test? IINW `vagrant destroy` doesn't remove that directory.

fmuyassarov (commented):
Looks good, left two questions below.

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,357 @@
...
+# RESULT_ROOT:   directory for publishing results
+# RESULT_NAME:   subdirectory for publishing results of this test run
+# SKIP_WORKTREE: skip preparing worktree, assume it is already done
+# E2E_TESTS:     e2e tests to run
+
+REMOTE_REPO="${REMOTE_REPO:-https://github.com/containers/nri-plugins}"
+CLONED_REPO="${CLONED_REPO:-/opt/e2e-test/nri-plugins/nri-plugins}"
+TEST_BRANCH="${TEST_BRANCH:-main}"
+RESULT_ROOT="${RESULT_ROOT:-/opt/e2e-test/nri-plugins/results}"
+RESULT_NAME="${RESULT_NAME:-$(date +%Y-%m-%d-%H-%M)}"
```
fmuyassarov, line 18: I'm wondering if it would be beneficial to have the latest commit's short SHA-1 added either here or somewhere else ?

scripts/testing/e2e-runner
klihub: Yes, but we do that later in `cleanup_worktree` using a `git clean -fdx && git worktree remove ...` combo.

scripts/testing/e2e-runner
klihub: Yes definitely, and it is one of the things I now try to add by generating a more reasonable/useful summary/index for the results. It is visible in the logs, but I too would like to explicitly spell out the SHA1 in the generated summary.

klihub: Updated the script to extract and store both report in the generated index the both the SHA1 and the result of `git describe` of the worktree, and to show these in the generated top-level index.

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+    fi
+
+    for var in ${!http*} ${!HTTP*} ${!no_*} ${!NO_*}; do
+        case $var in
+            *_proxy|*_PROXY)
+                info "exporting proxy envrionment variable $var (with value ${!var})..."
+                export $var
+                ;;
+        esac
+    done
```
marquiz, line 112: This nice bash-fu just caught my untrained eye 😅 How about spelling out the exact variable names (and making it easier to read, for me at least). And ignoring variables like `no_PROXY` :)

```suggestion
    for var in http_proxy https_proxy no_proxy HTTP_PROXY HTTPS_PROXY NO_PROXY; do
        if [ -n "${!var}" ]; then
                info "exporting proxy envrionment variable $var (with value ${!var})..."
                export $var
        fi
    done
```

Of course it's not 100% the same as in my suggestion empty variables are not exported

...you can disregard this comment if you don't like it

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+        must-cd "$CLONED_REPO"
+        if ! git remote -v | grep origin | grep -q "$REMOTE_REPO"; then
+            fatal "existing clone $CLONED_REPO has other origin than $REMOTE_REPO"
+        fi
+        if ! git fetch origin; then
+            fatal "failed to fetch $REMOTE_REPO"
+        fi
+    fi
+
+    must-mkdir "$worktree_dir" "worktree root"
```
marquiz, line 143: Is there a reason we're pre-mkdir'ing the directories (git clone and worktree) and not let git do it?

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+        must-mv "$src/$TEST_SUITE_DIR" "$dst"
+    done
+}
+
+cleanup_worktree() {
+    info "cleaning up worktree $WORKTREE..."
+    must-cd "$WORKTREE"
+    if ! git clean -fdx; then
+        fatal "failed to git clean worktree $WORKTREE"
+    fi
```
marquiz, line 233: Why clean it as we're removing it anyway? Just do `git worktree remove -f` below?

```suggestion
```

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+            fi
+            if [ "$result" = "FAIL" ]; then
+                failed=true
+            fi
+            echo "  $result $test"
+        done | sort >> "$summary"
+    done
+}
+
+generate_index() {
```
marquiz, line 274: Just a thought that shouldn't ideally the lower-level test runner script create this one? No need to change it for this pr...

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+    local exit_code="${exit_code:-0}"
+
+    echo "usage: ${1:-$SELF} [options] [<e2e-tests>]"
+    echo
+    echo "The possible options are:"
+    echo "  --source-proxies <file>  source proxy environment variables from file"
+    echo "  --remote|r <repo>        remote repository to clone/fetch from"
+    echo "  --branch|-b <branch>     name of the remote branch to test"
+    echo "  --local|-l <dir>         local directory to clone remote respository into"
+    echo "  --results|--out <dir>    top-level directory for publishing results"
```
marquiz, line 52: nit: `-o` not documented

```suggestion
    echo "  --results|--out <dir>|-o    top-level directory for publishing results"
```

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+    echo "  --name|-n) <name>        subdirectory for results of this run"
+    echo "  --skip-worktree)         skip worktree preparation, assume already done"
+    echo "  --help)                  show this help"
```
marquiz, line 55: Drop the parenthesis

```suggestion
    echo "  --name|-n <name>         subdirectory for results of this run"
    echo "  --skip-worktree          skip worktree preparation, assume already done"
    echo "  --help                   show this help"
```

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+        --name|-n)
+            RESULT_NAME="$2"
+            shift 2
+            ;;
+        --skip-worktree)
+            SKIP_WORKTREE=1
+            shift
+            ;;
+
+        --trace|-t|-x)
```
marquiz, line 369: This is not documented in command help

scripts/testing/e2e-runner
klihub: Well, IMO ideally neither this nor the lower-level one should create the index. It should be created by the entity which is serving out the test results over HTTP[S]. But to get us started, I hacked it in now here...

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
...
+
+#
+# Default configuration, overridable on the command line.
+#
+# REMOTE_REPO:   remote repository to clone
+# CLONED_REPO:   local location to clone to
+# RESULT_ROOT:   directory for publishing results
+# RESULT_NAME:   subdirectory for publishing results of this test run
+# SKIP_WORKTREE: skip preparing worktree, assume it is already done
+# E2E_TESTS:     e2e tests to run
```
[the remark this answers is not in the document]
klihub: Updated.

scripts/testing/e2e-runner
klihub: The only reason is to have a better understanding and clearer error messages in case we don't have the necessary permissions to create the directories. But I'm fine with letting `git clone` and `git worktree add` fail with whatever messages they see fit, too.

scripts/testing/e2e-runner
klihub: I'm fine with updating it like that, although I think it is not exactly the same (worktree remove --force will also nuke a locked directory while git clean -fdx is at least not documented to do the same) but I don't think that matters here.

scripts/testing/e2e-runner
klihub: Added.

scripts/testing/e2e-runner
klihub: Added.

scripts/testing/e2e-runner
klihub: But if either one is creating, it should be this one. The lower-lever test runner has nothing to do with HTTP.

scripts/testing/e2e-runner
klihub: Yeah, probably better that way, since we know all the variables of interest. Updated.

scripts/testing/e2e-runner
```diff
@@ -0,0 +1,396 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
klihub: Added.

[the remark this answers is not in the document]
klihub: Actually, now that I tested a few runs with a version updated with `set -eo pipefail`, this feels like a really bad idea. This script should survive as much as possible and try to collect any results it can get from a test run, instead of the opposite. I've tried to patch up all the failures I've been running into, but I'm seriously considering reverting this change...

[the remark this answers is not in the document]
klihub: Okay, with the latest fixes, it now seems to successfully run till completion and is able to update the index.

scripts/testing/e2e-runner
marquiz: What locking are you referring to? Locking worktrees or smth else?

marquiz (approved):
This is cool

@klihub feel free to drop the `set -e` plus related `|| :` if you want. I didn't realize `-e` might not be that good idea here. I'm fine with either approach

LGTM

klihub:
> This is cool
> 
> @klihub feel free to drop the `set -e` plus related `|| :` if you want. I didn't realize `-e` might not be that good idea here. I'm fine with either approach
> 
> LGTM

Okay, I dropped those. It feels safer to make changes in the future without breaking stuff.

PR #300, opened 2024-03-27
go.mod: align go version in go.mod with project's semantic version

fmuyassarov:
In the Go 1.21 release, support was introduced to enforce stricter version control within go.mod files, enabling the specification of patch versions. This commit synchronizes the Go version specified in go.mod with the semantic version used across other files (dockefile, etc) for the sake of consistency.

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -3,8 +3,8 @@ FROM sphinxdoc/sphinx:5.3.0
 RUN apt-get update && apt-get install -y wget git
 
 # Note: Any golang version that can 'go list -m -f {{.Variable}}' is fine...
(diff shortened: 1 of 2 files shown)


marquiz (approved):
I don't have strong opinions on this. We can merge this if we want. The downside is just one more place to remember to update when bumping the (patch) version of Golang

fmuyassarov:
> I don't have strong opinions on this. We can merge this if we want. The downside is just one more place to remember to update when bumping the (patch) version of Golang
> 
> EDIT: I think what I wanted to ask is does this some benefits over some potential issues?

This is not trying to fix anything. It is just trying to follow the same version. There are some benefits but not for our use case/project.

PR #301, opened 2024-03-28
.github: fix unstable chart version and registry path.

klihub:
Add latest fixes which I managed to mis-push and thus omit from the merged version. These fix the Helm chart version to have a leading 'v', add a trailing '/helm-charts' to the registry push path and update the documentation accordingly.


marquiz (approved):
LGTM

PR #308, opened 2024-03-28
.github: fix release check for Helm Chart packaging.

klihub:
Setting an evironment variable and using that same variable to trigger conditional jobs with 'if:' does not work as expected. Use a dedicated job to decide if we are dealing with a release tag and propagate that decision to the output of the job. Then use the output to trigger either one of a dependent release or unstable job.

### DIFF
--- a/.github/workflows/package-helm.yaml
+++ b/.github/workflows/package-helm.yaml
@@ -10,7 +10,6 @@ on:
 
 env:
   CHARTS_DIR: deployment/helm/
-  IS_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') && 'true' || 'false' }}
   UNSTABLE_CHARTS: unstable-helm-charts
   REGISTRY: ghcr.io
   REGISTRY_USER: ${{ github.repository_owner }}
@@ -18,7 +17,7 @@ env:
 
 jobs:
   release:
-    if: ${{ github.env.IS_RELEASE == 'true' }}
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
     permissions:
       contents: write
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


askervin (approved):
LGTM

marquiz (commented):
Hmm, this starts to look a bit messy.

Can't we just use the check (`${{ startsWith(...`) directly in the `if:` field of the jobs? The other one would check for `refs/tags` and the other one for `refs/heads`.  Or,  split this into two separate workflows (`release-helm-charts` and `unstable-helm-charts`) the other triggering from tags and the other from branches.

klihub:
> Hmm, this starts to look a bit messy.
> 
> Can't we just use the check (`${{ startsWith(...`) directly in the `if:` field of the jobs? The other one would check for `refs/tags` and the other one for `refs/heads`. Or, split this into two separate workflows (`release-helm-charts` and `unstable-helm-charts`) the other triggering from tags and the other from branches.

We have now the 'interim' tag for the next release cycle. It is unannotated but a tag nonetheless and it matches the current trigger. If we can change the trigger to not match the interim tag then we can try to switch back to a simpler if-tag-then-a-release logic. Although I think we can't check for that once and then propagate the result as environment variable to both alternative branches like I tried to do because IIUC the environment variable in that if-expression gets evaluated once in the beginning. But we could go with two `startsWith()` expressions. I just wish there was built-in support for `ifMatchesRegexp()` expressions in workflows...

klihub:
@marquiz Thanks. Rebased on #311, removed the non-working 'evaluate condition once, pass result through environment' attempt and replaced it with the direct evaluation of the condition where it is needed.

askervin (approved):
LGTM

PR #312, opened 2024-04-09
helm: flip podPriorityClassNodeCritical to true.

klihub:
Set `podPriorityClassNodeCritical` to true by default, adding plugins to the `system-node-critical` priority class. This should mitigate the potential risk of a plugin getting evicted under heavy system load. It should also ensure that during autoscaling enough new nodes are brought up to leave room for plugins on each new node.


fmuyassarov (approved):
LGTM

marquiz (approved):
LGTM

PR #315, opened 2024-04-16
.github: fetch tags for image build.

klihub:
If we build images from a shallow clone, `git describe` isn't able to determine a descriptive version derived from the latest tag. Therefore fetch full history with all tags for image builds.


marquiz (approved):
LGTM

PR #317, opened 2024-04-22
e2e: enable running commands in Qemu monitor and hotpluggin memory

askervin:
Launch Qemu always with monitor server listening to a unix socket file in the VM directory. Add test script functions vm-monitor, vm-mem-hw, vm-mem-hotplug, and vm-mem-hotremove.


askervin:
This PR paves way for testing topologies where memory can be hotplugged to the system.

Tested with a topology with nodes that have
1. cpu + memory
2. cpu-only
3. cpu + hotpluggable memory (memory appears in correct node)
4. no-cpu + hotpluggable memory (node is not shown by kernel, memory added to it appears into node0)
5. no-cpu + memory

Don't know if the behavior in case (4) is due to qemu parameters or kernel handling of numa nodes that don't have memory or cpus at boot. Rebooting did not change the issue, all hotplugged memory in this case remained mapped to node0.

klihub (approved):
LGTM.

fmuyassarov (approved):
LGTM

PR #318, opened 2024-05-17
config-manager: allow configuring NRI timeouts.

klihub:
Allow setting NRI plugin registration and request timeouts in the CRI-O or containerd configuration files. Require neither or both timeouts to be given. If both are given, verify that registration timeout is larger than the request timeout.

### DIFF
--- a/cmd/config-manager/main.go
+++ b/cmd/config-manager/main.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"flag"
 	"fmt"
 	"os"
 	"time"
@@ -43,7 +44,26 @@ var (
 	log = logrus.StandardLogger()
 )
 
+type nriConfig struct {
+	registrationTimeout string
+	requestTimeout      string
+}
+
 func main() {
+	var cfg nriConfig
+
+	flag.StringVar(&cfg.requestTimeout,
+		"nri-plugin-request-timeout", "", "NRI plugin request timeout to patch, as time.Duration")
+	flag.StringVar(&cfg.registrationTimeout,
+		"nri-plugin-registration-timeout", "", "NRI plugin registration timeout to patch, as time.Duration")
+
+	flag.Parse()
+
+	err := cfg.check()
+	if err != nil {
+		log.Fatalf("invalid NRI configuration requested: %v", err)
+	}
+
 	unit, conn, err := detectRuntime()
 	if err != nil {
 		log.Fatalf("failed to autodetect container runtime: %v", err)
@@ -52,9 +72,9 @@ func main() {
 
 	switch unit {
 	case containerdUnit:
-		err = enableNriForContainerd()
+		err = configureNriForContainerd(&cfg)
 	case crioUnit:
-		err = enableNriForCrio()
+		err = configureNriForCrio(&cfg)
 	default:
 		log.Fatalf("unknown container runtime %q", unit)
 	}
@@ -83,14 +103,14 @@ func main() {
 	log.Println("enabled NRI for", unit)
 }
 
-func enableNriForContainerd() error {
-	log.Infof("enabling NRI in containerd configuration...")
+func configureNriForContainerd(cfg *nriConfig) error {
+	log.Infof("configuring NRI for containerd...")
 	tomlMap, err := readConfig(containerdConfigFile)
 	if err != nil {
 		return fmt.Errorf("error reading TOML file: %w", err)
 	}
 
-	updatedTomlMap := updateContainerdConfig(tomlMap)
+	updatedTomlMap := updateContainerdConfig(tomlMap, cfg)
 
 	err = writeToContainerdConfig(containerdConfigFile, updatedTomlMap)
 	if err != nil {
@@ -99,8 +119,8 @@ func enableNriForContainerd() error {
 	return nil
 }
 
-func enableNriForCrio() error {
-	log.Infof("enabling NRI in CRI-O configuration...")
+func configureNriForCrio(cfg *nriConfig) error {
+	log.Infof("configuring NRI for CRI-O...")
 	f, err := os.Create(crioConfigFile)
 	if err != nil {
 		return fmt.Errorf("error creating a drop-in file for CRI-O: %w", err)
@@ -111,6 +131,12 @@ func enableNriForCrio() error {
 	if err != nil {
 		return fmt.Errorf("error writing a drop-in file for CRI-O: %w", err)
 	}
+
+	err = cfg.writeCrioConfig(f)
+	if err != nil {
+		return fmt.Errorf("error writing NRI configuration for CRI-O: %w", err)
+	}
+
 	return nil
 }
 
@@ -149,7 +175,7 @@ func readConfig(file string) (map[string]interface{}, error) {
 	return tomlMap, nil
 }
 
-func updateContainerdConfig(config map[string]interface{}) map[string]interface{} {
+func updateContainerdConfig(config map[string]interface{}, cfg *nriConfig) map[string]interface{} {
 	plugins, exists := config["plugins"].(map[string]interface{})
 	if !exists {
 		log.Println("top level plugins section not found, adding it to enable NRI...")
@@ -165,7 +191,71 @@ func updateContainerdConfig(config map[string]interface{}) map[string]interface{
 	}
 
 	nri["disable"] = false
+
+	cfg.updateContainerdConfig(config)
+
 	return config
+}
+
+func (cfg *nriConfig) check() error {
+	switch {
+	case cfg.registrationTimeout == "" && cfg.requestTimeout == "":
+		return nil
+	case cfg.registrationTimeout != "" && cfg.requestTimeout == "":
+		return fmt.Errorf("NRI plugin registration timeout set without request timeout")
+	case cfg.registrationTimeout == "" && cfg.requestTimeout != "":
+		return fmt.Errorf("NRI plugin request timeout set without registration timeout")
+	}
+
+	register, err := time.ParseDuration(cfg.registrationTimeout)
+	if err != nil {
+		return fmt.Errorf("invalid plugin registration timeout: %w", err)
+	}
+	request, err := time.ParseDuration(cfg.requestTimeout)
+	if err != nil {
+		return fmt.Errorf("invalid plugin request timeout: %w", err)
+	}
+
+	if register <= request {
+		return fmt.Errorf("NRI plugin registration timeout (%s) must be > request timeout (%s)",
+			register, request)
+	}
+
+	return nil
+}
+
+func (cfg *nriConfig) writeCrioConfig(f *os.File) error {
+	const (
+		registrationTimeout = "nri_plugin_registration_timeout"
+		requestTimeout      = "nri_plugin_request_timeout"
+	)
+	if cfg.registrationTimeout != "" {
+		key, value := registrationTimeout, "\""+cfg.registrationTimeout+"\""
+		if _, err := f.WriteString(key + " = " + value + "\n"); err != nil {
+			return err
+		}
+	}
+	if cfg.requestTimeout != "" {
+		key, value := requestTimeout, "\""+cfg.requestTimeout+"\""
+		if _, err := f.WriteString(key + " = " + value + "\n"); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (cfg *nriConfig) updateContainerdConfig(tomlCfg map[string]interface{}) {
+	const (
+		registrationTimeout = "plugin_registration_timeout"
+		requestTimeout      = "plugin_request_timeout"
+	)
+
+	if cfg.registrationTimeout != "" {
+		tomlCfg[registrationTimeout] = cfg.registrationTimeout
+	}
+	if cfg.requestTimeout != "" {
+		tomlCfg[requestTimeout] = cfg.requestTimeout
+	}
 }
 
 func detectRuntime() (string, *dbus.Conn, error) {
--- a/config/crd/bases/config.nri_nriplugindeployments.yaml
+++ b/config/crd/bases/config.nri_nriplugindeployments.yaml
@@ -69,8 +69,23 @@ spec:
                     nri:
                       type: object
                       properties:
-                        patchRuntimeConfig:
-                          type: boolean
+                        runtime:
+                          type: object
+                          properties:
+                            patchConfig:
+                              type: boolean
+                            config:
+                              type: object
+                              required:
+                              - pluginRegistrationTimeout
+                              - pluginRequestTimeout
+                              properties:
+                                pluginRegistrationTimeout:
+                                  type: string
+                                  pattern: "^(([5-9])|([1-2][0-9])|(30))s$"
+                                pluginRequestTimeout:
+                                  type: string
+                                  pattern: "^(([2-9])|([1-2][0-9])|(30))s$"
                     image:
                       type: object
                       description: image defines Plugin DeamonSet image name and tag
--- a/deployment/helm/balloons/templates/daemonset.yaml
+++ b/deployment/helm/balloons/templates/daemonset.yaml
@@ -28,9 +28,16 @@ spec:
     {{- with .Values.nodeSelector }}
         {{- toYaml . | nindent 8 }}
     {{- end }}
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if (not (or (eq .Values.nri.runtime.config nil) (eq .Values.nri.runtime.config.pluginRegistrationTimeout ""))) }}
+        args:
+          - -nri-plugin-registration-timeout
+          - {{ .Values.nri.runtime.config.pluginRegistrationTimeout }}
+          - -nri-plugin-request-timeout
+          - {{ .Values.nri.runtime.config.pluginRequestTimeout }}
+        {{- end }}
         image: {{ .Values.initContainerImage.name }}:{{ .Values.initContainerImage.tag | default .Chart.AppVersion }}
         imagePullPolicy: {{ .Values.initContainerImage.pullPolicy }}
         volumeMounts:
@@ -55,7 +62,7 @@ spec:
             - -metrics-interval
             - 5s
             - --nri-plugin-index
-            - "{{ .Values.nri.pluginIndex | int | printf "%02d"  }}"
+            - "{{ .Values.nri.plugin.index | int | printf "%02d"  }}"
             {{- if .Values.configGroupLabel }}
             - --config-group-label
             - {{ .Values.configGroupLabel }}
@@ -116,7 +123,7 @@ spec:
         hostPath:
           path: /var/run/nri
           type: DirectoryOrCreate
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       - name: containerd-config
         hostPath:
           path: /etc/containerd/
--- a/deployment/helm/memory-qos/README.md
+++ b/deployment/helm/memory-qos/README.md
@@ -17,13 +17,13 @@ parameters: QoS class and direct memory annotations.
       [these](https://github.com/containerd/containerd/blob/main/docs/NRI.md#enabling-nri-support-in-containerd)
       detailed instructions. You can optionally enable the NRI in containerd
       using the Helm chart during the chart installation simply by setting the
-      `nri.patchRuntimeConfig` parameter. For instance,
+      `nri.runtime.patchConfig` parameter. For instance,
 
       ```sh
-      helm install my-memory-qos nri-plugins/nri-memory-qos --set nri.patchRuntimeConfig=true --namespace kube-system
+      helm install my-memory-qos nri-plugins/nri-memory-qos --set nri.runtime.patchConfig=true --namespace kube-system
       ```
 
-      Enabling `nri.patchRuntimeConfig` creates an init container to turn on
+      Enabling `nri.runtime.patchConfig` creates an init container to turn on
       NRI feature in containerd and only after that proceed the plugin
       installation.
 
@@ -34,10 +34,10 @@ parameters: QoS class and direct memory annotations.
       [these](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md#crionri-table)
       detailed instructions.  You can optionally enable the NRI in CRI-O using
       the Helm chart during the chart installation simply by setting the
-      `nri.patchRuntimeConfig` parameter. For instance,
+      `nri.runtime.patchConfig` parameter. For instance,
 
       ```sh
-      helm install my-memory-qos nri-plugins/nri-memory-qos --namespace kube-system --set nri.patchRuntimeConfig=true
+      helm install my-memory-qos nri-plugins/nri-memory-qos --namespace kube-system --set nri.runtime.patchConfig=true
       ```
 
 ## Installing the Chart
@@ -57,14 +57,17 @@ values.yaml file and provide it using the `-f` flag. For example:
 
 ```sh
 # Install the memory-qos plugin with custom values provided using the --set option
-helm install my-memory-qos nri-plugins/nri-memory-qos --namespace kube-system --set nri.patchRuntimeConfig=true
+helm install my-memory-qos nri-plugins/nri-memory-qos --namespace kube-system --set nri.runtime.patchConfig=true
 ```
 
 ```sh
 # Install the memory-qos plugin with custom values specified in a custom values.yaml file
 cat <<EOF > myPath/values.yaml
 nri:
-  patchRuntimeConfig: true
+  runtime:
+    patchConfig: true
+  plugin:
+    index: 10
 
 tolerations:
 - key: "node-role.kubernetes.io/control-plane"
@@ -95,8 +98,10 @@ customize with their own values, along with the default values.
 | `image.pullPolicy`       | Always                                                                                                                        | image pull policy                                    |
 | `resources.cpu`          | 10m                                                                                                                           | cpu resources for the Pod                            |
 | `resources.memory`       | 100Mi                                                                                                                         | memory qouta for the Pod                         |
-| `nri.patchRuntimeConfig` | false                                                                                                                         | enable NRI in containerd or CRI-O                    |
-| `nri.pluginIndex`        | 40                                                                                                                            | NRI plugin index to register with                    |
+| `nri.runtime.config.pluginRegistrationTimeout` | ""                                                                                                      | set NRI plugin registration timeout in NRI config of containerd or CRI-O |
+| `nri.runtime.config.pluginRequestTimeout`      | ""                                                                                                      | set NRI plugin request timeout in NRI config of containerd or CRI-O |
+| `nri.runtime.patchConfig` | false                                                                                                                        | patch NRI configuration in containerd or CRI-O       |
+| `nri.plugin.index`        | 90                                                                                                                           | NRI plugin index to register with            
 | `initImage.name`         | [ghcr.io/containers/nri-plugins/config-manager](https://ghcr.io/containers/nri-plugins/config-manager)                        | init container image name                            |
 | `initImage.tag`          | unstable                                                                                                                      | init container image tag                             |
 | `initImage.pullPolicy`   | Always                                                                                                                        | init container image pull policy                     |
--- a/deployment/helm/memory-qos/templates/daemonset.yaml
+++ b/deployment/helm/memory-qos/templates/daemonset.yaml
@@ -27,9 +27,16 @@ spec:
     {{- with .Values.nodeSelector }}
         {{- toYaml . | nindent 8 }}
     {{- end }}
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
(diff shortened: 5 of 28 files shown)


cmd/config-manager/main.go
```diff
@@ -43,7 +44,24 @@ var (
...
+
+	flag.StringVar(&cfg.requestTimeout,
+		"nri-plugin-request-timeout", "", "configure NRI plugin request timeout")
+	flag.StringVar(&cfg.registrationTimeout,
+		"nri-plugin-registration-timeout", "", "configure NRI plugin registration timeout")
+
+	err := cfg.check()
+	if err != nil {
+		log.Fatal("invalid NRI configuration requested: %v", err)
+	}
```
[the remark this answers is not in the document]
klihub: Yes, that's what I was doing with `log.Fatal`, but it would have rendered the message incorrectly. Fixed to use `Fatalf` instead.

cmd/config-manager/main.go
```diff
@@ -43,7 +44,26 @@ var (
...
+	requestTimeout      string
+}
+
 func main() {
+	var cfg nriConfig
+
+	flag.StringVar(&cfg.requestTimeout,
+		"nri-plugin-request-timeout", "", "configure NRI plugin request timeout")
+	flag.StringVar(&cfg.registrationTimeout,
+		"nri-plugin-registration-timeout", "", "configure NRI plugin registration timeout")
```
fmuyassarov, line 58: this is a nit, feel free to just ignore it. Perhaps If it was me, I would have dropped the `nri` prefix to make those args shorter

fmuyassarov, line 58: instead of string, can we perhaps use `flag.Int()`? One obvious benefit is that we can check if the timeout value is valid.

cmd/config-manager/main.go
```diff
@@ -165,9 +191,73 @@ func updateContainerdConfig(config map[string]interface{}) map[string]interface{
...
+		}
+	}
+	return nil
+}
+
+func (cfg *nriConfig) updateContainerdConfig(tomlCfg map[string]interface{}) {
+	const (
+		registrationTimeout = "plugin_registration_timeout"
+		requestTimeout      = "plugin_request_timeout"
+	)
```
fmuyassarov, line 251: can we make these as global constants and reuse them in whichever methods we need them?

deployment/helm/memory-qos/README.md
```diff
@@ -95,8 +98,10 @@ customize with their own values, along with the default values.
...
-| `nri.pluginIndex`        | 40                                                                                                                            | NRI plugin index to register with                    |
+| `nri.runtime.config.pluginRegistrationTimeout` | ""                                                                                                      | set NRI plugin registration timeout in NRI config of containerd or CRI-O |
+| `nri.runtime.config.pluginRequestTimeout`      | ""                                                                                                      | set NRI plugin request timeout in NRI config of containerd or CRI-O |
+| `nri.runtime.patchConfig` | false                                                                                                                        | patch NRI configuration in containerd or CRI-O       |
+| `nri.plugin.index`        | 90                                                                                                                           | NRI plugin index to register with
```
fmuyassarov, line 104: I think one more layer that you added (`.runtime`) makes sense, and it indicates that we are touching runtime related bits.

deployment/helm/memory-qos/templates/daemonset.yaml
```diff
@@ -27,9 +27,14 @@ spec:
...
         {{- toYaml . | nindent 8 }}
     {{- end }}
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if .Values.nri.runtime.registrationTimeout }}
+        args:
+          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.registrationTimeout }}
+          - -nri-plugin-request-timeout {{ .Values.nri.runtime.requestTimeout }}
```
fmuyassarov, line 36:
```suggestion
          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.config.registrationTimeout }}
          - -nri-plugin-request-timeout {{ .Values.nri.runtime.config.requestTimeout }}
```

deployment/helm/memory-qos/values.yaml
```diff
@@ -13,8 +13,12 @@ resources:
...
 
 nri:
-  patchRuntimeConfig: false
-  pluginIndex: 40
+  plugin:
+    index: 90
+  runtime:
+    patchConfig: false
+#    registrationTimeout: 5s
+#    requestTimeout: 4s
```
fmuyassarov, line 21:
```suggestion
#    config:
#       registrationTimeout: 5s
#       requestTimeout: 4s
```
I'm not sure which one should be the final you preferred. But I noticed that some places are missing `.config` level or the vise versa, some places have unnecessary `.config`

deployment/helm/memtierd/templates/daemonset.yaml
```diff
@@ -28,9 +28,14 @@ spec:
...
     {{- end }}
       hostPID: true
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if .Values.nri.runtime.registrationTimeout }}
+        args:
+          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.registrationTimeout }}
+          - -nri-plugin-request-timeout {{ .Values.nri.runtime.requestTimeout }}
```
fmuyassarov, line 37:
```suggestion
          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.config.registrationTimeout }}
          - -nri-plugin-request-timeout {{ .Values.nri.runtime.config.requestTimeout }}
```

deployment/helm/memtierd/values.yaml
```diff
@@ -15,8 +15,12 @@ resources:
...
 
 nri:
-  patchRuntimeConfig: false
-  pluginIndex: 45
+  plugin:
+    index: 90
+  runtime:
+    patchConfig: false
+#    registrationTimeout: 5s
+#    requestTimeout: 4s
```
fmuyassarov, line 23: same comment here.

deployment/helm/sgx-epc/templates/daemonset.yaml
```diff
@@ -28,9 +28,14 @@ spec:
...
         {{- toYaml . | nindent 8 }}
     {{- end }}
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if .Values.nri.runtime.registrationTimeout }}
+        args:
+          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.registrationTimeout }}
+          - -nri-plugin-request-timeout {{ .Values.nri.runtime.requestTimeout }}
```
fmuyassarov, line 37:
```suggestion
          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.config.registrationTimeout }}
          - -nri-plugin-request-timeout {{ .Values.nri.runtime.config.requestTimeout }}
```

deployment/helm/sgx-epc/values.yaml
```diff
@@ -13,8 +13,12 @@ resources:
...
 
 nri:
-  patchRuntimeConfig: false
-  pluginIndex: 40
+  plugin:
+    index: 90
+  runtime:
+    patchConfig: false
+#    registrationTimeout: 5s
+#    requestTimeout: 4s
```
fmuyassarov, line 21: same here

deployment/helm/template/templates/daemonset.yaml
```diff
@@ -21,9 +21,14 @@ spec:
...
       nodeSelector:
         kubernetes.io/os: "linux"
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if .Values.nri.runtime.registrationTimeout }}
+        args:
+          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.registrationTimeout }}
+          - -nri-plugin-request-timeout {{ .Values.nri.runtime.requestTimeout }}
```
fmuyassarov, line 30:
```suggestion
          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.config.registrationTimeout }}
          - -nri-plugin-request-timeout {{ .Values.nri.runtime.config.requestTimeout }}
```

deployment/helm/template/values.yaml
```diff
@@ -36,8 +36,12 @@ resources:
...
 
 nri:
-  patchRuntimeConfig: false
-  pluginIndex: 90
+  plugin:
+    index: 90
+  runtime:
+    patchConfig: false
+#    registrationTimeout: 5s
+#    requestTimeout: 4s
```
fmuyassarov, line 44: same here

deployment/helm/topology-aware/templates/daemonset.yaml
```diff
@@ -28,9 +28,14 @@ spec:
...
         {{- toYaml . | nindent 8 }}
     {{- end }}
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if .Values.nri.runtime.registrationTimeout }}
+        args:
+          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.registrationTimeout }}
+          - -nri-plugin-request-timeout {{ .Values.nri.runtime.requestTimeout }}
```
fmuyassarov, line 37:
```suggestion
          - -nri-plugin-registration-timeout {{ .Values.nri.runtime.config.registrationTimeout }}
          - -nri-plugin-request-timeout {{ .Values.nri.runtime.config.requestTimeout }}
```

deployment/helm/topology-aware/values.yaml
```diff
@@ -36,8 +36,12 @@ resources:
...
 
 nri:
-  patchRuntimeConfig: false
-  pluginIndex: 90
+  plugin:
+    index: 90
+  runtime:
+    patchConfig: false
+#    registrationTimeout: 5s
+#    requestTimeout: 4s
```
fmuyassarov, line 44: same here

deployment/operator/README.md
```diff
@@ -39,7 +39,12 @@ spec:
   state: present
   values:
     nri:
-      patchRuntimeConfig: true
+      plugin:
+        index: 90
+      runtime:
+        patchConfig: true
+#       registrationTimeout: 5s
+#       requestTimeout: 4s
```
fmuyassarov, line 47: same here

fmuyassarov (commented):
This looks already good. I did a quick review and will do it again. Left some minor comments or nits.

fmuyassarov:
> @fmuyassarov Looking at the updated docs, one question immediately came to my mind related to the updated helm charts. Considering the new timeout values in `Values.nri.config`, would it be better/more consistent to use `.Values.nri.config.patch: true` instead of the current `.Values.nri.patchConfig` ?

Yes, makes sense. That way we are more consistent indeed. I would also rather rename word patch to something more descriptive but at the same time I don't have a good suggestion to offer :)

cmd/config-manager/main.go
klihub: It is a valid point and I was considering that, too. Why I eventually chose these command line options is that these are literally the same as the corresponding command line options for CRI-O (in containerd you don't have fine-grain configuration control using the command line).

klihub: Again I wanted to use the exact same semantics as for the corresponding command line options in CRI-O. So these are actually `time.Duration`s.

cmd/config-manager/main.go
klihub: No, because they are different for CRI-O and containerd.

cmd/config-manager/main.go
klihub: And we do check now if the timeout value is valid in `nriConfig.check()` using `time.ParseDuration()`.

deployment/helm/memory-qos/README.md
klihub: Yes, I screwed up the value naming in docs wrt. the actual Helm charts. Fixed.

deployment/helm/memory-qos/templates/daemonset.yaml
klihub: Fixed.

deployment/helm/template/templates/daemonset.yaml
klihub: Fixed.

deployment/helm/topology-aware/templates/daemonset.yaml
klihub: Fixed.

deployment/operator/README.md
klihub: Fixed.

klihub:
> UPDATE: A slight problem (if we can really call it a problem) is that adding .config. makes it more deeper which might be something that users don't like. But that might be not be true really. Just thinking out loud.

Yes, I had the reaction when updating the daemonset files with templating conditionals. I still decided to stick with it for consistency.

deployment/helm/template/values.yaml
klihub: Fixed.

cmd/config-manager/main.go
```diff
@@ -165,9 +191,73 @@ func updateContainerdConfig(config map[string]interface{}) map[string]interface{
...
+
+	return nil
+}
+
+func (cfg *nriConfig) writeCrioConfig(f *os.File) error {
+	const (
+		registrationTimeout = "nri_plugin_registration_timeout"
+		requestTimeout      = "nri_plugin_request_timeout"
+	)
+	if cfg.registrationTimeout != "" {
```
fmuyassarov, line 232: Technically it is possible to put even zero right now, since we don't have validation that would block passing 0 value. Zero value could be a problematic and should not be allowed I would assume. Should we perhaps check weather value is non zero before entering the condition ?

deployment/helm/balloons/templates/daemonset.yaml
```diff
@@ -28,9 +28,16 @@ spec:
     {{- with .Values.nodeSelector }}
         {{- toYaml . | nindent 8 }}
     {{- end }}
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if (not (or (eq .Values.nri.runtime.config nil) (eq .Values.nri.runtime.config.pluginRegistrationTimeout ""))) }}
+        args:
```
fmuyassarov, line 35: Same here, we should not probably allow setting 0 value. And also, do we really need validation for the timeouts here? Because we have the proper validation later in the chain when those args are passed as flags into the go code. But if you want to fail early, then perhaps we should keep it here.

cmd/config-manager/main.go
klihub: Yes, we probably should do some further checking. We already check that the registration timeout is > than the request timeout. Maybe you were right in your earlier comment and we should not expect timeouts in time.ParseDuration-accepted formats but instead restrict it much more strictly. For instance allow only seconds to be given and document this, then don't accept units in the setting, only integers and use a minimum and a maximum in the verification schema to limit the timeouts to some reasonable range, for instance 1 sec <= requestTimeout <= 5 sec and 5 sec <= registrationTimeout <= 15 sec...

Would that sound reasonable ? I really would like to get some basic value verification in the schema, so that instead of creating the pod of the deployment failing, creating the deployment would fail before trying to schedule any pods. It is much more user friendly than having to use kubectl get * to dig out what might be the reason of the pods failing.

deployment/helm/balloons/templates/daemonset.yaml
klihub: We do need this. Otherwise you cannot patch the configuration just to enable NRI without giving both timeouts as well, which we don't want. Also, we don't want to attempt working around that by giving default settings to the timeouts which we believe to be the right defaults just because at the time of this PR to config-manager they happened to be the defaults in NRI...

klihub: Maybe we could use an explicit `"default"` to indicate we want to use a default value (IOW leave it untouched in the configuration), but then we can't use integers for the timeout so we can't enforce a simple range check using schema minimum and maximum. We'd need to use a `pattern` with a regexp like `^((default)|(2|3|4|5|6|7|8|9|(10)|(11)|(12)|(13)|(14)|(15))$`... for requestTimeout and a corresponding one for registrationTimeout.

cmd/config-manager/main.go
```diff
@@ -165,9 +191,73 @@ func updateContainerdConfig(config map[string]interface{}) map[string]interface{
...
+	return nil
+}
+
+func (cfg *nriConfig) updateContainerdConfig(tomlCfg map[string]interface{}) {
+	const (
+		registrationTimeout = "plugin_registration_timeout"
+		requestTimeout      = "plugin_request_timeout"
+	)
+
+	if cfg.registrationTimeout != "" {
```
askervin, line 253: If a slow NRI plugin is installed first with registrationTimeout 30s, and then a fast plugin with registrationTimeout 2s, we currently replace needed 30s timeout with 2s. Therefore I'm tempted to suggest always using max(currentRegistrationTimeout, newRegistrationTimeout).

An obvious downside of this suggestion is that if a timeout is too long, one cannot be reset to original unless we would have some kind of "force reset" bit telling this should be done.

I'd still prefer playing safe with max(). Philosophically, NRI plugins contain logic that is likely to be crucial for expected (correct) behavior of the whole node. Therefore, if one has to choose whether to give too much or too little time for a plugin to register, it's safer to give too much. If plugin needs only a little, we lose nothing. But we can loose a lot by living on the edge where plugins may be kicked out just because of fluctuation in node load.

deployment/helm/balloons/templates/daemonset.yaml
fmuyassarov: > Otherwise you cannot patch the configuration just to enable NRI without giving both timeouts as well, which we don't want.

I think we can, if you had something like this:
```yaml
{{- if .Values.nri.runtime.patchConfigg }}
initContainers:
- name: patch-runtime
  image: {{ .Values.initContainerImage.name }}:{{ .Values.initContainerImage.tag | default .Chart.AppVersion }}
  imagePullPolicy: {{ .Values.initContainerImage.pullPolicy }}
  volumeMounts:
  - name: containerd-config
    mountPath: /etc/containerd
  - name: crio-config
    mountPath: /etc/crio/crio.conf.d
  - name: dbus-socket
    mountPath: /var/run/dbus/system_bus_socket
  securityContext:
    privileged: true
  args:
  {{- if .Values.nri.runtime.config.pluginRegistrationTimeout }}
    - -nri-plugin-registration-timeout
    - {{ .Values.nri.runtime.config.pluginRegistrationTimeout }}
  {{- end }}
  {{- if .Values.nri.runtime.config.pluginRequestTimeout }}
    - -nri-plugin-request-timeout
    - {{ .Values.nri.runtime.config.pluginRequestTimeout }}
  {{- end }}
{{- end }}
```

config/crd/bases/config.nri_nriplugindeployments.yaml
```diff
@@ -69,8 +69,11 @@ spec:
                     nri:
                       type: object
                       properties:
-                        patchRuntimeConfig:
-                          type: boolean
+                        runtime:
+                          type: object
+                          properties:
+                            patchConfig:
+                              type: boolean
```
fmuyassarov, line 76: I bet we need to update the CRD here to include timeouts if we want consumers of the operator to be able to tune the timeouts. 
And if we do update the CRD here, we should also update the [README](https://github.com/containers/nri-plugins/blob/main/deployment/operator/README.md).

cmd/config-manager/main.go
klihub: > If a slow NRI plugin is installed first with registrationTimeout 30s, and then a fast plugin with registrationTimeout 2s, we currently replace needed 30s timeout with 2s. Therefore I'm tempted to suggest always using max(currentRegistrationTimeout, newRegistrationTimeout).

I have a few reservations with that.

1. I think we should be careful not to give the impression that the primary means to manage the NRI-related bits in the runtime configuration would be injecting config-manager as an init container.
2. The semantics of these timeouts patched in via the config-manager should not be considered 'this plugin wants a registration timeout X'. The semantics are statedly 'I want the runtime to be reconfigured to enable NRI, and reconfigure the global common NRI timeouts for plugin registration and request handling'. We try to emphasize this by calling the timeouts on the user-visible interface (Helm chart values)  `nri.runtime.config.plugin{Registration,Request}Timeout` and not `nri.plugin.{registration,request}Timeout`.

On the practical front, there are shortcomings in the current config-manager implementation that would make it a bit laborous to add. Currently config-manager only reads (and parses to toml) the full configuration of containerd, but not CRI-O.

> 
> An obvious downside of this suggestion is that if a timeout is too long, one cannot be reset to original unless we would have some kind of "force reset" bit telling this should be done.
> 
> I'd still prefer playing safe with max(). Philosophically, NRI plugins contain logic that is likely to be crucial for expected (correct) behavior of the whole node. Therefore, if one has to choose whether to give too much or too little time for a plugin to register, it's safer to give too much. If plugin needs only a little, we lose nothing. But we can loose a lot by living on the edge where plugins may be kicked out just because of fluctuation in node load.

Again, I've always considered patching the configuration using config-manager a stop-gap solution to lower that bar for entry. To allow folks to deploy and play around with NRI plugins in their clusters with relative ease, even though the runtimes default to NRI being off. @kad ?

config/crd/bases/config.nri_nriplugindeployments.yaml
klihub: It's updated in the next commit...

fmuyassarov: Sure, I will run the operator and let you know.

fmuyassarov: @klihub here is what I got
```
fmuyassarov@fmuyassa-MOBL1:/tmp/nri-plugins$ make -C deployment/operator deploy IMAGE=feruzjon/nri-plugins-operator:b4fde25552e46840d9c88da7f0fed989e6bb33dc
make: Entering directory '/tmp/nri-plugins/deployment/operator'
mkdir -p config/crd/bases
cp ../../config/crd/bases/* config/crd/bases
cp ../../sample-configs/balloons-config.yaml config/samples
cp ../../sample-configs/template-config.yaml config/samples
cp ../../sample-configs/topologyaware-config.yaml config/samples
cp ../../sample-configs/noderesourcetopology.yaml config/samples
cd config/manager && /usr/local/bin/kustomize edit set image controller=feruzjon/nri-plugins-operator:b4fde25552e46840d9c88da7f0fed989e6bb33dc
/usr/local/bin/kustomize build config/default | kubectl apply -f -
Error: accumulating resources: accumulation err='accumulating resources from '../crd': '/tmp/nri-plugins/deployment/operator/config/crd' must resolve to a file': recursed accumulation of path '/tmp/nri-plugins/deployment/operator/config/crd': accumulating resources: accumulating resources from 'bases/config.nri_nriplugindeployments.yaml': MalformedYAMLError: yaml: line 82: mapping values are not allowed in this context in File: bases/config.nri_nriplugindeployments.yaml
error: no objects passed to apply
make: *** [Makefile:186: deploy] Error 1
make: Leaving directory '/tmp/nri-plugins/deployment/operator'
```

deployment/helm/balloons/templates/daemonset.yaml
klihub: @fmuyassarov I ended up using the values JSON schema for limiting the timeouts and for ensuring that both (or neither) are given. The registration timeout is now limited to `5-30s` and the request timeout is `2-30s`.

That leaves us with the slightly user-unfriendly runtime check of `requestTimeout < registrationTimeout`, but I don't think the JSON schema would be expressive enough to allow expressing such a constraint, so this is the best I can do.

cmd/config-manager/main.go
klihub: @fmuyassarov I ended up using the values JSON schema for limiting the timeouts and for ensuring that both (or neither) are given. The registration timeout is now limited to `5-30s` and the request timeout is `2-30s`.

That leaves us with the slightly user-unfriendly runtime check of `requestTimeout < registrationTimeout`, but I don't think the JSON schema would be expressive enough to allow expressing such a constraint, so this is the best I can do.

deployment/helm/balloons/templates/daemonset.yaml
klihub: > > Otherwise you cannot patch the configuration just to enable NRI without giving both timeouts as well, which we don't want.
> 
> I think we can, if you had something like this:
> 
> ```yaml
> {{- if .Values.nri.runtime.patchConfigg }}
> initContainers:
> - name: patch-runtime
>   image: {{ .Values.initContainerImage.name }}:{{ .Values.initContainerImage.tag | default .Chart.AppVersion }}
>   imagePullPolicy: {{ .Values.initContainerImage.pullPolicy }}
>   volumeMounts:
>   - name: containerd-config
>     mountPath: /etc/containerd
>   - name: crio-config
>     mountPath: /etc/crio/crio.conf.d
>   - name: dbus-socket
>     mountPath: /var/run/dbus/system_bus_socket
>   securityContext:
>     privileged: true
>   args:
>   {{- if .Values.nri.runtime.config.pluginRegistrationTimeout }}
>     - -nri-plugin-registration-timeout
>     - {{ .Values.nri.runtime.config.pluginRegistrationTimeout }}
>   {{- end }}
>   {{- if .Values.nri.runtime.config.pluginRequestTimeout }}
>     - -nri-plugin-request-timeout
>     - {{ .Values.nri.runtime.config.pluginRequestTimeout }}
>   {{- end }}
> {{- end }}
> ```

That would allow you to give just one of the timeouts which is something I don't want to allow.

config/crd/bases/config.nri_nriplugindeployments.yaml
klihub: (I think I) Fixed (it).

cmd/config-manager/main.go
kad: I don't have strong preferences here. In theory, we shouldn't have situation when two instances of helm chart with different plugins are trying to (re-)configure runtime. The second one should detect that if NRI is already enabled/configured then just exit. Not sure with error code or with 0... both options have their own + and -.

deployment/helm/topology-aware/templates/daemonset.yaml
```diff
@@ -28,9 +28,16 @@ spec:
...
-      {{- if .Values.nri.patchRuntimeConfig }}
+      {{- if .Values.nri.runtime.patchConfig }}
       initContainers:
       - name: patch-runtime
+        {{- if (not (or (eq .Values.nri.runtime.config nil) (eq .Values.nri.runtime.config.pluginRegistrationTimeout ""))) }}
+        args:
+          - -nri-plugin-registration-timeout
+          - {{ .Values.nri.runtime.config.pluginRegistrationTimeout }}
+          - -nri-plugin-request-timeout
+          - {{ .Values.nri.runtime.config.pluginRequestTimeout }}
```
marquiz, line 39: Hmm, if only one of those is defined we end up in trouble(?)

config/crd/bases/config.nri_nriplugindeployments.yaml
fmuyassarov: @klihub one more issue to be fixed but we are close
```diff
diff --git a/config/crd/bases/config.nri_nriplugindeployments.yaml b/config/crd/bases/config.nri_nriplugindeployments.yaml
index 38f4134..4d0ff56 100644
--- a/config/crd/bases/config.nri_nriplugindeployments.yaml
+++ b/config/crd/bases/config.nri_nriplugindeployments.yaml
@@ -74,19 +74,18 @@ spec:
                           properties:
                             patchConfig:
                               type: boolean
-                          config:
-                            type: object
-                            properties:
+                            config:
+                              properties:
+                                pluginRegistrationTimeout:
+                                  type: string
+                                  pattern: "^(([5-9])|([1-2][0-9])|(30))s$"
+                                pluginRequestTimeout:
+                                  type: string
+                                  pattern: "^(([2-9])|([1-2][0-9])|(30))s$"
+                              type: object
                               required:
-                                - pluginRegistrationTimeout
-                                - pluginRequestTimeout
-                              pluginRegistrationTimeout:
-                                type: string
-                                pattern: "^(([5-9])|([1-2][0-9])|(30))s$"
-                              pluginRequestTimeout:
-                                type: string
-                                pattern: "^(([2-9])|([1-2][0-9])|(30))s$"
-
+                              - pluginRegistrationTimeout
+                              - pluginRequestTimeout
                     image:
                       type: object
                       description: image defines Plugin DeamonSet image name and tag

```

klihub: Argh, thanks for spotting that !

deployment/helm/topology-aware/templates/daemonset.yaml
klihub: Helm shouldn't accept that. The values JSON schema effectively allows either both or neither to be specified.

fmuyassarov (commented):
lgtm

fmuyassarov (approved):

deployment/helm/topology-aware/templates/daemonset.yaml
marquiz: K, then in the template it should use `and` instead of `or`(?)

PR #321, opened 2024-05-23
operator: fix formatting and missing fields in the operator crd

fmuyassarov:
Fix formatting and missing fields in the operator CRD. Otherwise, currently you can't create a CR to tune operator parameters like `patchConfig` or give the `plugin.index`.

### DIFF
--- a/config/crd/bases/config.nri_nriplugindeployments.yaml
+++ b/config/crd/bases/config.nri_nriplugindeployments.yaml
@@ -14,1001 +14,1050 @@ spec:
     - npd
   scope: Namespaced
   versions:
-    - name: v1alpha1
-      schema:
-        openAPIV3Schema:
-          description: NriPluginDeployment is the Schema for the nriplugindeployments API
-          properties:
-            apiVersion:
-              description: 'APIVersion defines the versioned schema of this representation
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: NriPluginDeployment is the Schema for the nriplugindeployments
+          API
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
               of an object. Servers should convert recognized schemas to the latest
               internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
-              type: string
-            kind:
-              description: 'Kind is a string value representing the REST resource this
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
               object represents. Servers may infer this from the endpoint the client
               submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
-              type: string
-            metadata:
-              type: object
-            spec:
-              description: Spec defines the intended state of NriPluginDeployment
-              properties:
-                pluginName:
-                  type: string
-                  x-kubernetes-validations:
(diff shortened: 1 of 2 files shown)


config/crd/bases/config.nri_nriplugindeployments.yaml
```diff
@@ -14,1001 +14,1050 @@ spec:
...
+                          description: Operator represents a key's relationship to
+                            the value. Valid operators are Exists and Equal. Defaults
+                            to Equal. Exists is equivalent to wildcard for value,
+                            so that a pod can tolerate all taints of a particular
+                            category.
+                          enum:
+                          - Exists
+                          - Equal
+                          type: string
+                        "tolerationS\teconds":
```
klihub, line 1033: That tab in `"tolerationS\teconds"` looks like a typo to me.

fmuyassarov: I don't know how it ended up to be like that we have it in main as well https://github.com/containers/nri-plugins/blob/1f45c70f6aa07e6be68c6c18f744b5c4bc48d0cc/config/crd/bases/config.nri_nriplugindeployments.yaml#L984. I will fix it though.

klihub: Yeah, weird. It was already in the original commit for the operator. It has been there since day 1 of the operator, already present in the original commit 4ab020681 like that. None of us spotted it then...

klihub (approved):
LGTM

deployment/operator/README.md
```diff
@@ -40,12 +40,12 @@ spec:
...
       plugin:
-        index: 90
+        index: 90                       # optional
       runtime:
-        patchConfig: true
-#       config:
-#         pluginRegistrationTimeout: 5s
-#         pluginRequestTimeout: 2s
+        patchConfig: true               # optional
+      config:
```
klihub, line 46: I think this is tabulated incorrectly. The value key for that config is `values.nri.runtime.config`, not `values.nri.config`. So it should be at the same level as `patchConfig`, non ?

deployment/operator/README.md
```diff
@@ -40,12 +40,12 @@ spec:
...
+        pluginRegistrationTimeout: 5s   # optional
+        pluginRequestTimeout: 2s        # optional
```
klihub, line 48: I would change `optional` to `either neither or both of the timeouts should be given and patchConfig needs to be true for updating timeouts`.

fmuyassarov: I removed the optional comment and added a note in about ^ below.

deployment/operator/README.md
```diff
@@ -40,12 +40,12 @@ spec:
...
       plugin:
-        index: 90
+        index: 90                       # optional
       runtime:
-        patchConfig: true
-#       config:
-#         pluginRegistrationTimeout: 5s
-#         pluginRequestTimeout: 2s
+        patchConfig: true               # optional
+        config:
```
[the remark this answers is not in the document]
fmuyassarov: done

klihub (approved):
LGTM.

Also, sorry and thank you for putting up with my unnecessarily multi-part nit-b*tching.

askervin (approved):
LGTM.

The only thing that I was a bit suspicious about were value ranges of pluginRegistrationTimeout (5-30s) and pluginRequestTimeouts (2-30s). But as we have discussed earlier, it's not our purpose to use helm as the primary runtime config editor :D, but just a handy helper that covers most simple and common changes. So I think this is good.

PR #322, opened 2024-05-23
e2e: use github API to list and get latest release.

klihub:
Our method of using the github.com/$repo/releases/latest to query and determine the latest release of containerd, CRI-O or kubernetes is flawed. The semantics of that 'latest' is calendar time of release, so it will pick wrong versions if a new tagged maintenance release is the newest by calendar time. Therefore, use the official github API instead to list all tagged releases not marked as a prerelease, and pick the latest one by tag.


klihub:
Unfortunately from behind a proxy with enough users, this gets easily rate-limited without authentication. So it's leprosy vs. black death, I guess. Maybe we need to go back to listing releases page by page and parsing HTML.

PR #323, opened 2024-05-23
e2e: fetch and version-sort github releases to get latest release.

klihub:
Our method of querying github.com/$repo/releases/latest to determine the latest release of containerd, CRI-O or kubernetes is unasafe. The semantics of 'latest' is up to the package maintainer. Since errors happen, we can end up picking and trying to use a too old version. So instead, list releases and version-sort them to find the latest one.

Ideally we would use the official github API to list all tagged releases not marked as a prerelease, and pick the latest one by tag. However, without authenticating that API is severly rate limited and fails often from behind busy proxies. So instead, fetch pages of releases, parse them for release URLs, sort them by version and pick the

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -23,27 +23,47 @@ export cni_plugin=${cni_plugin:-cilium}
 export cni_release=${cni_release:-latest}
 TOPOLOGY_DIR=${TOPOLOGY_DIR:=e2e}
 
-K8S_REPO="https://github.com/kubernetes/kubernetes"
+GH_K8S_REPO="kubernetes/kubernetes"
 export k8s_release=${k8s_release:-"latest"}
 export k8s_version=""
 
 source "$LIB_DIR"/vm.bash
 
+list-github-releases () {
+    local repo="$1" max="${max_releases:-20}" page=1 tags="" more="" sep=""
+    local total=0 cnt=-1
+
+    while [ "$cnt" != 0 -a "$total" -lt "$max" ]; do
+        echo -n "fetching releases for $repo (page $page)..." 1>&2
+        if more=$(wget -q "https://github.com/$repo/releases?page=$page" -O- | \
+                      grep -E '.*/releases/tag/v[0-9]+.[0-9]+.[0-9]+"' | \
+                      sed -E 's|.*/releases/tag/(v[0-9]+.[0-9]+.[0-9]+).*|\1|g')
+        then
+            tags="$tags$sep$more"
+            sep=" "
+            cnt="$(echo $more | wc -w)"
+            total=$((total+cnt))
+            echo " got $cnt more, $total in total..." 1>&2
+            page=$((page+1))
+        else
+            echo "failed to determine latest github release for $repo"
+            return 1
+        fi
+    done
+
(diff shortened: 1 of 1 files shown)


klihub:
BTW: There are probably plenty of more low hanging fruits to fix/improve on this front. For instance, I just noticed that the scripts determine the latest runtime and K8s release (thus accessing and fetching plenty of data from github) for every single test we run. Since these cannot really change once the test VM has been provisioned it would be better to determine the versions once then store them under the test VM directory in case they are needed for further, version-specific test scripting logic.

klihub:
> BTW: There are probably plenty of more low hanging fruits to fix/improve on this front. For instance, I just noticed that the scripts determine the latest runtime and K8s release (thus accessing and fetching plenty of data from github) for every single test we run. Since these cannot really change once the test VM has been provisioned it would be better to determine the versions once then store them under the test VM directory in case they are needed for further, version-specific test scripting logic.

This is now addressed in #324, which is stacked on top of this PR.

test/e2e/run.sh
```diff
@@ -23,27 +23,47 @@ export cni_plugin=${cni_plugin:-cilium}
...
+        if more=$(wget -q "https://github.com/$repo/releases?page=$page" -O- | \
+                      grep -E '.*/releases/tag/v[0-9]+.[0-9]+.[0-9]+"' | \
+                      sed -E 's|.*/releases/tag/(v[0-9]+.[0-9]+.[0-9]+).*|\1|g')
+        then
+            tags="$tags$sep$more"
+            sep=" "
+            cnt="$(echo $more | wc -w)"
+            total=$((total+cnt))
+            echo " got $cnt more, $total in total..." 1>&2
+            let page=$page+1
```
askervin, line 47: Style-nit: on line 45 (and elsewhere in the code) arithmetic evaluation and variable assignment happens with var=$(( ... )), so maybe 47 could be written similarly without using `let`.

test/e2e/run.sh
```diff
@@ -23,27 +23,47 @@ export cni_plugin=${cni_plugin:-cilium}
...
 latest-github-release () {
-    local repo="$1"
-    local latest=$(wget -q -O- $repo/releases/latest | grep '/tree/v[0-9.]*' |
-                       sed -E 's:.*/tree/v([0-9]*\.[0-9]*)(\.[0-9]*).*:\1\2:g' | head -1)
-    if [ -z "$latest" ]; then
-        echo "failed to determine latest release of $repo"
-        exit 1
-    fi
-    echo "$latest"
+    list-github-releases "$1" | head -1
```
askervin, line 58: Given that `latest-github-release` is digging out only the latest release, would it suffice if it calls `list-github-releases` with max_releases=1 or max_releases=2?

askervin (commented):
Commit message: s/unasafe/unsafe.

test/e2e/run.sh
klihub: Unfortunately no. Generally speaking this whole thing is probabilistic. We don't know how many releases are listed per page (depends on how much collective metadata is associated with each release), and there is no guarantee that the latest release would be among the ones listed on the first page. So we're now a bit conservative and rather list too many than too few.

askervin (approved):
LGTM, Thanks @klihub!

PR #324, opened 2024-05-23
e2e: cache and restore latest github versions

klihub:
Note: this PR is stacked on top of #323.

Add support for caching and restoring variables per test VM. Once the `latest` github version of CRI-O, containerd or kubernetes has been resolved to a real version tag, save it in the new vm-specific variable cache. Always try restoring these versions from cache before turning to discover them from github. This should significantly reduce how much we punish/bang on github for the full set of our e2e tests.


askervin (approved):
LGTM

askervin:
Thanks @klihub!

PR #325, opened 2024-05-30
resmgr: don't log about NRT CR updates when object isn't created

fmuyassarov:
You can see the following lines of logging in the balloons plugin for example:
```
I: [         resource-manager         ] updating topology zones...
I: [               agent              ] updating node resource topology CR
```
Currently it is only [TA plugin](https://github.com/containers/nri-plugins/blob/82131f985fbe5d496215c23cd0a198acfed75bcd/cmd/plugins/topology-aware/policy/topology-aware-policy.go#L260) that actually fetches the zone data, while other plugins like balloons doesn't fetch the data even though it has a [method](https://github.com/containers/nri-plugins/blob/82131f985fbe5d496215c23cd0a198acfed75bcd/cmd/plugins/balloons/policy/balloons-policy.go#L291) for that. As such, with this change, we now check if there is a zone data and only if yes, we allow calling methods & functions for updating the NRT CR.

### DIFF
--- a/pkg/agent/node-resource-topology.go
+++ b/pkg/agent/node-resource-topology.go
@@ -67,24 +67,13 @@ func (a *Agent) updateNrtCR(policy string, zones []*policyapi.TopologyZone) erro
 		}
 	}
 
-	// delete existing CR if we got no data from policy
 	// XXX TODO Deletion should be handled differently:
 	//   1. add expiration timestamp to nrt.NodeResourceTopology
 	//   2. GC CRs that are past their expiration time (for instance by NFD)
 	//   3. make sure we refresh our CR (either here or preferably/easier
 	//      by triggering in resmgr an updateTopologyZones() during longer
 	//      periods of inactivity)
-	if len(zones) == 0 {
-		if cr != nil {
-			err := cli.Delete(ctx, a.nodeName, metav1.DeleteOptions{})
-			if err != nil && !errors.IsNotFound(err) {
-				return fmt.Errorf("failed to delete node resource topology CR: %w", err)
-			}
-		}
-		return nil
-	}
-
-	// otherwise update CR if one exists
+	// update CR if one exists
 	if cr != nil {
 		cr.Attributes = nrt.AttributeList{
 			nrt.AttributeInfo{
--- a/pkg/resmgr/resource-manager.go
+++ b/pkg/resmgr/resource-manager.go
@@ -271,10 +271,11 @@ func (m *resmgr) startControllers() error {
 
 // updateTopologyZones updates the 'topology zone' CRDs.
 func (m *resmgr) updateTopologyZones() {
-	m.Info("updating topology zones...")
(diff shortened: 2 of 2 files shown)


pkg/agent/node-resource-topology.go
```diff
@@ -67,24 +67,13 @@ func (a *Agent) updateNrtCR(policy string, zones []*policyapi.TopologyZone) erro
...
-	if len(zones) == 0 {
-		if cr != nil {
-			err := cli.Delete(ctx, a.nodeName, metav1.DeleteOptions{})
-			if err != nil && !errors.IsNotFound(err) {
-				return fmt.Errorf("failed to delete node resource topology CR: %w", err)
-			}
-		}
-		return nil
-	}
-
```
fmuyassarov, line 86: I know that by this, we are not covering the deletion of CR when there is no zone data. But I think we won't be even creating the CR in the first place if we don't call `UpdateNrtCR()` unnecessarily (when zone data doesn't exist).

pkg/resmgr/resource-manager.go
```diff
@@ -271,10 +271,12 @@ func (m *resmgr) startControllers() error {
...
-	m.Info("updating topology zones...")
-	err := m.agent.UpdateNrtCR(m.policy.ActivePolicy(), m.policy.GetTopologyZones())
-	if err != nil {
-		m.Error("failed to update topology zones: %v", err)
+	if len(m.policy.GetTopologyZones()) != 0 {
+		m.Info("updating topology zones...")
+		err := m.agent.UpdateNrtCR(m.policy.ActivePolicy(), m.policy.GetTopologyZones())
+		if err != nil {
+			m.Error("failed to update topology zones: %v", err)
+		}
```
klihub, line 279: nit: I would update this to query topology zone once, check if we got data and if we do, pass the already queried data to the agent for publishing. IOW, something like
```
  if zones := m.policy.GetTopologyZones(); len(zones) != 0 {
    m.Info("updating topology zones...")
    if err := m.agent.UpdateNrtCR(m.policy.ActivePolicy(), zones);  err != nil {
      m.Error("failed to update topology zones: %v" ,err)
    }
  }
```

pkg/agent/node-resource-topology.go
klihub: Yes, probably not something to worry about.

klihub (changes requested):
I have a small nit to avoid querying and generating the same topology zone data twice, once to find out if there is data at all, then a second time if there was data, to pass it on to the agent. Let's generate it just once. Otherwise LGTM.

pkg/resmgr/resource-manager.go
fmuyassarov: fixed.

klihub (approved):

pfl (approved):
LGTM

askervin (approved):
LGTM

PR #326, opened 2024-05-30
fixes: fix topology hint tests, hint generation for non-empty sysroot.

klihub:
This PR fixes topology hint units tests, test data, and hint generation for non-empty sysroots. In particular this PR has the following changes:
- clean sysroot upon setting
- return resolved device paths without sysroot prefix (as expected by NewTopologyHints)
- don't run topology tests parallel (tests set/clear sysroot which is a global shared setting)
- add missing test data for /dev/null


pkg/topology/topology.go
```diff
@@ -300,6 +300,12 @@ func findSysFsDevice(devType string, major, minor int64) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to get realpath for %s: %w", devPath, err)
 	}
+	if sysRoot != "" {
+		realDevPath = strings.TrimPrefix(realDevPath, sysRoot)
+		if realDevPath != "" && realDevPath[0] != '/' {
+			realDevPath = "/" + realDevPath
+		}
```
fmuyassarov, line 307: just curious, why do we need to add a slash ? to indicate it as path?

fmuyassarov (commented):
@klihub this looks good to me. I have a quick question in line. Also, your https://github.com/containers/nri-plugins/pull/326/commits/2024ea90071d72717c81e03d3a544478ef06d754 commit is doing something else it seems? There is a tarball added which might be added accidentally ...?

klihub:
> @klihub this looks good to me. I have a quick question in line. Also, your [2024ea9](https://github.com/containers/nri-plugins/commit/2024ea90071d72717c81e03d3a544478ef06d754) commit is doing something else it seems? There is a tarball added which might be added accidentally ...?

That's the 2nd bullet point above. It adds missing data for /dev/null. One of the test cases tries to resolve /dev/null to a sysfs path, but our sample sysfs test data lacks /dev/null information. That test case only worked by accident, due to another bug. Test cases were run in parallel, and each test case which sets up the package global non-standard sysfs path also clears/resets it on tetscase teardown. Since tests were run in parallel, the first and fastest test almost always had time to reset the sysfs path, and the /dev/null resolution test was resolving /dev/null using the hosts sysfs, instead of our sample test data. Once that bug was fixed, the /dev/null test started understandably failing. Adding the missing data for that test case then fixed this and it started working for real, not just by accident.

fmuyassarov:
> @fmuyassarov So this PR consists of 3 separate fixes to the topology package in 3 separate commits. I tried to describe each with a one-liner bullet point in the PR description [above](https://github.com/containers/nri-plugins/pull/326#issue-2325921797).

Yes I know. I just didn't realize that the tarball was the test data.

fmuyassarov (approved):
LGTM

pkg/topology/topology.go
```diff
@@ -300,6 +300,12 @@ func findSysFsDevice(devType string, major, minor int64) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to get realpath for %s: %w", devPath, err)
 	}
+	if sysRoot != "" {
+		realDevPath = strings.TrimPrefix(realDevPath, sysRoot)
```
kad, line 304: should we better to check that if the `realDevPath` prefixed with `sysRoot`, instead of just checking if `sysRoot` is not empty?

kad: As a side note, in `SetSysRoot` we can use `filepath.Clean()` to make sure that the value is correct, so operations with those paths will be more predictable. e.g. in line 181.

askervin:
Validated in real environment with multiple PCI devices. Works great. Thank you, @klihub. I think Draft status can be removed.

pkg/topology/topology.go
klihub: Updated to check prefix against cleaned sysroot.

askervin: Path returned by `filepath.Clean()` is promised not to end with "/" except for in the case of the root directory. Therefore, if user calls `SetSysRoot("/")`, the latest change seems to make it possible that `findSysFsDevice()` returns paths that do not start with "/". I'm not sure if this will cause issues, and whether or not it is possible to call `SetSysRoot("/")`, but at least it is more possible than it was before.

askervin (commented):
A question on "always absolute path" promise from findSysFsDevice().

pkg/topology/topology.go
klihub: Updated to 1) only use implicit "/" as sysroot and 2) always resolve relative sysroot to absolute one, panicking if this fails. Added a few basic test cases for `SetSysRoot()`.

askervin (approved):
Thanks @klihub, LGTM.

PR #327, opened 2024-06-07
Add NRI plugin for CDI device injection

elezar:
This PR adds a plugin for CDI injection.

The values of defined pod annotations are interpreted as a comma-separated list of CDI devices. The local CDI cache is read and the corresponding container adjusments are generated.

Some items to do:
- [ ] Make the CDI folders configurable

In the longer-term, it should also be possible to restrict the injection of devices to containers in a particular namespace / based on properties of a namespace.


klihub:
@elezar @askervin I rolled a proof-of-concept/initial implementation of CDI device injection in NRI, directly support by the protocol and supporting libraries. I also updated containerd 1.7 and main/2.0 and cri-o main to allow CDI injections. The NRI bits also update the test/sample device-injector plugin (in the NRI repo itself) to allow injecting CDI devices by annotation. This is what I used for testing against all of the updated containerd and cri-o trees.

Here are the related working trees/branches:

 - https://github.com/klihub/nri/tree/devel/native-cdi-injection
 - https://github.com/klihub/containerd/tree/devel/release-1.7/nri-cdi-injection
 - https://github.com/klihub/containerd/tree/devel/release-2.0/nri-cdi-injection
 - https://github.com/klihub/cri-o/tree/devel/nri-cdi-injection
 
Based on this, it looks technically feasible, fairly straightforward, and not very intrusive to implement. Next we should decide if we want to push this or not. It has pros but maybe semantically it is a bit of an outlier compared to the rest of our features.

cmd/plugins/cdi-device-injector/cdi-device-injector.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, nil, fmt.Errorf("failed to parse CDI Device annotations: %w", err)
+	}
+
+	if len(cdiDevices) == 0 {
+		return nil, nil, nil
+	}
+
+	var allowedCDIDevices []string
+	for _, cdiDevice := range cdiDevices {
+		match, _ := filepath.Match(p.allowedCDIDevicePattern, cdiDevice)
```
klihub, line 80: I think it would be better to use `path.Match()` for globbing instead of `filepath.Match()`.

klihub:
@elezar Is this still relevant for you ? Or should this be closed now as obsolete since we have DRA adminAccess ?

PR #328, opened 2024-06-07
pkg/{cgroups,topology}: move linux-specific functions

elezar:
This change moves linux-specific functions to `_linux.go` source files and adds stub implementations for other platforms.

### DIFF
--- a/pkg/cgroups/cgroupid_linux.go
+++ b/pkg/cgroups/cgroupid_linux.go
@@ -1,0 +1,16 @@
+package cgroups
+
+import (
+	"encoding/binary"
+
+	"golang.org/x/sys/unix"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/cgroups/cgroupid_linux.go
```diff
@@ -0,0 +1,12 @@
+package cgroups
+
+import "encoding/binary"
```
askervin, line 3: missing import "golang.org/x/sys/unix".

elezar: Yes, you're right. I was building locally on a `darwin` machine. Let me do so in a container too.

klihub (approved):
LGTM

askervin (approved):
LGTM

PR #329, opened 2024-06-11
topology-aware: don't ignore HBM memory nodes without close CPUs.

klihub:
Treat HBM memory nodes without close CPUs similarly to PMEM nodes, IOW assign them to the closest DRAM node as extra capacity. Also, for now don't ever allocate HBM memory implicitly to containers.


askervin (approved):
LGTM. Thanks @klihub, this is very nice.

PR #330, opened 2024-06-11
topology-aware: log memory type preference as such.

klihub:
Don't log effective memory type preference as cold start preference.


askervin (approved):
LGTM

PR #331, opened 2024-06-17
topology-aware: add support for hide-hyperthreads annotation.

askervin:
Containers with effective hide-hyperthreads annotation are allowed to run only on the first CPU hyperthread of all physical CPU cores allocated to its pool.

### DIFF
--- a/cmd/plugins/topology-aware/policy/pod-preferences.go
+++ b/cmd/plugins/topology-aware/policy/pod-preferences.go
@@ -44,6 +44,8 @@ const (
 	keyReservedCPUsPreference = "prefer-reserved-cpus"
 	// annotation key for CPU Priority preference
 	keyCpuPriorityPreference = "prefer-cpu-priority"
+	// annotation key for hiding hyperthreads from allocated CPU sets
+	keyHideHyperthreads = "hide-hyperthreads"
 
 	// effective annotation key for isolated CPU preference
 	preferIsolatedCPUsKey = keyIsolationPreference + "." + kubernetes.ResmgrKeyNamespace
@@ -57,6 +59,8 @@ const (
 	preferReservedCPUsKey = keyReservedCPUsPreference + "." + kubernetes.ResmgrKeyNamespace
 	// effective annotation key for CPU priority preference
 	preferCpuPriorityKey = keyCpuPriorityPreference + "." + kubernetes.ResmgrKeyNamespace
+	// effective annotation key for hiding hyperthreads
+	hideHyperthreadsKey = keyHideHyperthreads + "." + kubernetes.ResmgrKeyNamespace
 )
 
 // cpuClass is a type of CPU to allocate
@@ -185,6 +189,20 @@ func cpuPrioPreference(pod cache.Pod, container cache.Container, fallback cpuPri
 
 	log.Debug("%s: explicit CPU priority preference %q", container.PrettyName(), prio)
 	return prio
+}
+
+// hideHyperthreadsPreference returns if a container should run using
+// only single hyperthread from each physical core.
+func hideHyperthreadsPreference(pod cache.Pod, container cache.Container) bool {
+	value, ok := container.GetEffectiveAnnotation(hideHyperthreadsKey)
+	if !ok {
+		return false
+	}
+	hide, err := strconv.ParseBool(value)
+	if err != nil {
+		return false
+	}
+	return hide
 }
 
 // memoryTypePreference returns what type of memory should be allocated for the container.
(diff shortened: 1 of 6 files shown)


pkg/sysfs/system.go
```diff
@@ -104,6 +104,7 @@ type System interface {
 	CoreKindCPUs(CoreKind) cpuset.CPUSet
 	CoreKinds() []CoreKind
 	AllThreadsForCPUs(cpuset.CPUSet) cpuset.CPUSet
+	SingleThreadPerCore(cpuset.CPUSet) cpuset.CPUSet
```
klihub, line 107: nitpick: although sysfs is already far from being anywhere consistent wrt. naming, I'd still prefer either misnaming this to `SingleThreadForCPUs` or renaming the existing `AllThreadsForCPUs` to `AllThreads[For|Per]Cores`.

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -588,6 +589,25 @@ func (p *policy) allocatePool(container cache.Container, poolHint string) (Grant
 	return grant, nil
 }
 
+// allowAllocatedCPUs pins container's CPUs according to what has been
+// allocated for it, taking into account if the container should run
+// with hyperthreads hidden.
+func (p *policy) allowAllocatedCPUs(container cache.Container, allocated cpuset.CPUSet, info string) {
```
klihub, line 595: nitpick: I'd probably call this set* since it is essentially a filter+SetCpusetCpus()... maybe call it `setAllocatedCpus` or `setPreferredCpusetCpus`.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -44,6 +44,8 @@ const (
 	keyReservedCPUsPreference = "prefer-reserved-cpus"
 	// annotation key for CPU Priority preference
 	keyCpuPriorityPreference = "prefer-cpu-priority"
+	// annotation key for hiding hyperthreads from allocated CPU sets
+	keyHideHyperthreads = "hide-hyperthreads"
```
klihub, line 48: Do/don't we have something like this in the balloons policy, and if we do, is this key consistent with the annotation used there ?

klihub (commented):
Looks good to me. I have a few nits and a question related to the end-user visible annotation name.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -57,6 +59,9 @@ const (
 	preferReservedCPUsKey = keyReservedCPUsPreference + "." + kubernetes.ResmgrKeyNamespace
 	// effective annotation key for CPU priority preference
 	preferCpuPriorityKey = keyCpuPriorityPreference + "." + kubernetes.ResmgrKeyNamespace
+	// effective annotation key for hiding hyperthreads
+	hideHyperthreadsKey = keyHideHyperthreads + "." + kubernetes.ResmgrKeyNamespace
+
```
klihub, line 64: Whitespace/formatting error, flagged by CI as well.

pkg/sysfs/system.go
askervin: Renamed to SingleThreadForCPUs.

cmd/plugins/topology-aware/policy/pod-preferences.go
askervin: Balloons is still in my feature branch and it does not support annotation at all. Now that we have topology-aware implementation in place, I know how to rewrite the balloons part so that it'll fit in the picture. (Add support for the same annotation and start using the SingleThread filter from system.go instead of having it in cputree.go)

askervin: Couldn't really find a good place for a common annotation. Still thinking if we should have this annotation parser in the cache package, similarly to preserve-* annotations. Or will I use only GetEffectiveAnnotation from there and parse the boolean in balloons and topology-aware separately - which is the direction that this PR is now trending to. What do you think @klihub? I can still move it to the cache interface, but on the other hand, I would not like to add there too many annotations that may not make sense in every policy.

cmd/plugins/topology-aware/policy/pod-preferences.go
askervin: Fixed, thanks.

askervin:
Ready for re-review and fine for merging in my perspective with one open question/reconsideration.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: I think putting it to `pkg/resmgr/cache` implies that we expect all policies to implement it, so we should be careful with that. Although is a simple boolean preference which is semantically simple enough that it might fit that bill, I have no strong opinions and am fine with either way.

klihub (approved):
LGTM

PR #332, opened 2024-06-18
libmem: implement policy-agnostic memory allocation/accounting.

klihub:
This PR is work in progress.

This patch series 
- implements a  policy-agnostic library for memory allocation/accounting
- replaces existing memory accounting in the `topology-aware` policy with libmem
- plugs in libmem-based memory accounting and allocation to the `balloons` policy

The PR is (partially) stacked on top of #329.

TODO for bringing this out of draft state:
- [ ] get #329 merged then rebase
- [ ] add missing `ensureNormalMemory()` during initial zone selection
- [ ] add a description of how the allocator works/describe the basic idea 
- [ ] comment/document key builtin logic (zone expansion, zone OOM-resolution)
- [ ] clean up debug-logging to be actually helpful 
- [ ] make sure memory type preference is consistently either a preference or a requirement (not it is not)

Questions:
- [ ] should we have more granular prioritization than the current QoS class mapped to `Inertia`
- [ ] should we add an interface for overriding builtin zone expansion and OOM-resolution
- [ ] are the builtin defaults now sane
- [ ] should we allow for both strict and preferred memory type selection


cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1417,11 +1428,93 @@ func (p *balloons) pinCpuMem(c cache.Container, cpus cpuset.CPUSet, mems idset.I
 	if p.bpoptions.PinMemory == nil || *p.bpoptions.PinMemory {
 		if c.PreserveMemoryResources() {
 			log.Debug("  - preserving %s pinning to memory %q", c.PrettyName, c.GetCpusetMems())
+			preserveMems, err := parseIDSet(c.GetCpusetMems())
+			if err != nil {
+				log.Error("failed to parse CpusetMems: %v", err)
+			} else {
+				zone := p.allocMem(c, preserveMems, true)
+				log.Debug("  - allocated preserved memory %s", c.PrettyName, zone)
+				c.SetCpusetMems(zone.MemsetString())
```
askervin, line 1437: When preservation is required by outside this policy, I think we should never touch CpusetMems, even if it would screw up our bookkeeping somehow. Instead of SetCpusetMems it would be good to log.Error if zone is different from original pinning?

askervin (commented):
....starting review. I could not yet follow through the whole path if it is or not possible that PreserveContainer requests can return zones different from originals. But if they do... for instance if a single zone of size 32 GB is the only zone that is allowed for two preserved-memory containers, each requesting 32 GB of RAM, then I'd still assume that whoever did this preservation must know what he is doing. If he prefers oom kill of his containers rather than letting them access memory out of the zone, then so be it, we won't touch pinning.

cmd/plugins/balloons/policy/balloons-policy.go
klihub: That is slightly different and called in libmem a `reservation` as opposed to a container/allocation with `Preserved` inertia/memory. libmem never moves reservations around. Preservations it can move around as a last resort, but libmem guarantees that if a Preserved allocation is moved around, the new memory set is always a superset of the old one. IOW, a preserved allocation never loses any memory node it once had access to.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1462,13 +1466,91 @@ func (p *balloons) pinCpuMem(c cache.Container, cpus cpuset.CPUSet, mems idset.I
...
+			}
 		} else {
 			log.Debug("  - pinning %s to memory %s", c.PrettyName(), mems)
-			c.SetCpusetMems(mems.String())
+			zone := p.allocMem(c, mems, false)
+			log.Debug("  - allocated memory %s", c.PrettyName, zone)
```
askervin, line 1480: mismatching formats and parameters

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1462,13 +1466,91 @@ func (p *balloons) pinCpuMem(c cache.Container, cpus cpuset.CPUSet, mems idset.I
...
+				c.GetID(),
+				c.PrettyName(),
+				string(c.GetQOSClass()),
+				amount,
+				nodes,
+			)
+		}
+		zone, updates, err = p.memAllocator.Allocate(req)
+	} else {
+		zone, updates, err = p.memAllocator.Realloc(c.GetID(), nodes, 0)
```
askervin, line 1515: A call to Realloc(_, _, **0**) can never change allocation, as memory types mask contains no allowed types. Is the idea to hand control (without types?) to custom expanders? Didn't quite understand this...

klihub: No... I think I changed the semantics somewhere at some point and the idea was that `libmem.Realloc(id, nodes, 0)` would be equivalent to `libmem(id, nodes, libmem.ZoneType(nodes))`... but either something fell through the cracks somewhere or I changed my mind for some reason. I'll take a look at which one it was...

klihub: Update with those semantics and added a test case for it. But I kept those as WiP for now. Need to think a bit more about whether these semantics make the most sense, before squashing those commits.

cmd/plugins/balloons/policy/balloons-policy.go
klihub: Fixed.

pkg/resmgr/lib/memory/allocator.go
```diff
@@ -0,0 +1,771 @@
...
+
+		a.ForeachNode(nodes, func(n *Node) bool {
+			n.ForeachDistance(func(dist int, distNodes NodeMask) bool {
+				if dist > minDist {
+					return false
+				}
+
+				extra := distNodes & a.masks.nodes.byTypes[t.Mask()] &^ nodes
+				if extra == 0 {
+					return dist < minDist // only continue if next set closer than minDist
```
askervin, line 514: Not fully sure about this, so this is more of a question: is it possible that a later call into this for-callback would come with dist == minDist and with `extra != 0`? That is, it would still expand to some valid extra nodes if we would allow continuing the loop. If this is the case, maybe we should actually `return dist <= minDist`...

askervin (commented):
Minor notes... I'll continue still looking at this tomorrow, as I'd wish to take it in as soon as possible.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: The only alternative semantics that I could come up with would be returning an error when trying to reallocate memory with all nodes masked out. In that case `libmem.Realloc(id, nodes, -1)` could be used for reallocating on any possible node.

I could imagine the mask "accidentally" ending up to 0 due to calculating intersections of memory preferences (from annotations and balloontype configurations, for instance), but even in that case the responsibility of finding and handling the issue would be most natural on the caller side of Realloc. And therefore I do not see it necessary to get an error from Realloc, if called with 0.

That is, my gut feeling is almost numb. Without any thinking my personal initial choice might have been -1=any and 0=error, but I like your implementation better, because 0=any scales nicely to more than 64 nodes case. :D

pkg/resmgr/lib/memory/allocator.go
```diff
@@ -0,0 +1,771 @@
...
+func (a *Allocator) Allocate(req *Request) (NodeMask, map[string]NodeMask, error) {
+	log.Debug("allocate memory for %s", req)
+	return a.allocate(req)
+}
+
+// Realloc updates an existing allocation with the given extra affinity
+// and memory types. Therefore Realloc is semantically always expansive.
+// You cannot remove memory nodes or types from an allocation by Realloc.
+// Realloc returns the updated nodes for the allocation, together with
+// any updates necessary to other existing allocations for fulfillig the
```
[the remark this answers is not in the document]
klihub: Ah, thanks ! On a related, once we merge it, #356 should help preventing these kinds of typos in the future...

pkg/resmgr/lib/memory/allocator.go
klihub: Hmm... I'm not entirely sure why that would be better... but let me describe this in a bit more detail and then you can tell me whether it would make sense (and why) in the light of that if that is still the case.

The innermost `n.ForeachDistance()` iterates sets of nodes (NodeMasks) at the same distance by increasing distance from `n`. So if `dist == minDist`, it is guaranteed that we are right now considering the last possible set of nodes at minDist or closer distance, so there should be no reason to consider the next closest set of nodes, but... If `extra == 0` and `types` is a TypeMask of more than one type, then we might do other iterations (of the outermost loop) for another type `t`, and during that iteration we might pick up nodes of other type from the very same node set. In this light, is the original/suggestion still valid ?

I think regarding this implementation of expand one relevant question is: would it make more sense to ditch the current outermost loop, IOW type-by-type iteration, and simply iterate `nodes` by distance and neighbor set, since we `a.masks.nodes.byTypes` is set up so that you can look up nodes by any type mask, not just ones with a single bit set...

pkg/resmgr/lib/memory/allocator.go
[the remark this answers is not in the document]
klihub: Fixed.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1462,13 +1466,91 @@ func (p *balloons) pinCpuMem(c cache.Container, cpus cpuset.CPUSet, mems idset.I
...
+		amount  = getMemoryLimit(c)
+		nodes   = libmem.NewNodeMask(mems.Members()...)
+		req     *libmem.Request
+		zone    libmem.NodeMask
+		updates map[string]libmem.NodeMask
+		err     error
+	)
+
+	if _, ok := p.memAllocator.AssignedZone(c.GetID()); !ok {
+		if c.PreserveMemoryResources() {
```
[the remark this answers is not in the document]
klihub: Yes, definitely a thinko... Updated the signature to pass in the fact, but then failed to update the implementation to use it.

askervin:
@klihub, I'm testing this PR with the topmost (e2e) commit in https://github.com/askervin/nri-plugins/commits/5Wj_libmem_debug/

It introduces a topology with DRAM, HBM and PMEM memories detected. (I also considered adding memory-less CPU-only nodes, but didn't do it yet. Let me know if you'd like to include that as well to the topology...)

Launching
```
.../test/e2e$ ./run_tests.sh policies.test-suite/balloons/n6-hbm-cxl/test01-memory-types/ 
```
produces "insufficient available memory" prints in the balloons log:

```
E: [           policy           ] allocMem: falling back to nodes{1}, failed to allocate memory for kube-system/kube-scheduler-n6-hbm-cxl-generic-fedora37-containerd/kube-scheduler: libmem: insufficient available memory: failed to reallocate, can't find new DRAM nodes
```

Nevertheless pinned memory zones look good to me.

klihub:
> @klihub, I'm testing this PR with the topmost (e2e) commit in https://github.com/askervin/nri-plugins/commits/5Wj_libmem_debug/
> 
> It introduces a topology with DRAM, HBM and PMEM memories detected. (I also considered adding memory-less CPU-only nodes, but didn't do it yet. Let me know if you'd like to include that as well to the topology...)
> 
> Launching
> 
> ```
> .../test/e2e$ ./run_tests.sh policies.test-suite/balloons/n6-hbm-cxl/test01-memory-types/ 
> ```
> 
> produces "insufficient available memory" prints in the balloons log:
> 
> ```
> E: [           policy           ] allocMem: falling back to nodes{1}, failed to allocate memory for kube-system/kube-scheduler-n6-hbm-cxl-generic-fedora37-containerd/kube-scheduler: libmem: insufficient available memory: failed to reallocate, can't find new DRAM nodes
> ```
> 
> Nevertheless pinned memory zones look good to me.

I'll pull in those tests and see if I can dig out from the logs/debugging what goes wrong...

klihub:
> > Nevertheless pinned memory zones look good to me.
> 
> That's just because the fallback code paths for failure are not totally insane... but that's nothing to write home about.
> 
> I'll pull in those tests and see if I can dig out from the logs/debugging what goes wrong...

Okay, so I think this is a typical the left Dr. Jekyll does not know/remember how the right Mr. Hyde intended the `Realloc()` API to be used. When updating pinning we now always call `Realloc()` from balloons regardless of whether anything (in CPU pinning) has changed. But `Realloc()` does not really expect this to happen and always tries to force expansion... which can fails if we we run out of suitable nodes. Updating `Realloc()` to accomodate for such calls gets rid of the errors.

BTW, the test cases should use MEMLIM instead of MEMREQ... MEMREQ still can't be recovered with acceptable granularity (mapped to a range of something like [-2000, 2000] which is lacking way too many bits).

pkg/resmgr/lib/memory/allocator.go
klihub: @askervin I updated `defaultExpand()` along those lines described above.

pkg/resmgr/lib/memory/types.go
```diff
@@ -0,0 +1,318 @@
...
+		m |= (1 << TypeForSysfs(st))
+	}
+	return m & TypeMaskAll
+}
+
+// ParseTypeMask parses the given string into a TypeMask.
+func ParseTypeMask(str string) (TypeMask, error) {
+	m := TypeMask(0)
+	for _, s := range strings.Split(str, ",") {
+		t, err := ParseType(s)
```
askervin, line 182: Minor details... Do you think ParseTypeMask could allow whitespace around the mask string and commas? I have stumbled into this a couple of time in our current `memory-type` annotation parser, as it's quite natural to type annotations with spaces: `memory-types...: hbm, dram`.

I'd like to use this parser in refactoring the annotation code and making it work in both balloons and topology-aware.

pkg/resmgr/lib/memory/types.go
```diff
@@ -0,0 +1,318 @@
...
+	"strings"
+
+	"github.com/containers/nri-plugins/pkg/sysfs"
+)
+
+// Type represents known types of memory.
+type Type int
+
+const (
+	TypeDRAM Type = iota // ordinary DRAM
```
askervin, line 29: As some hardware topologies may have nodes with CPUs and without memory, should we prepare for that by having `TypeNoMem` (or something like that), too? Or do we want to handle them as DRAM or <dontcare> nodes with zero memory capacity?

klihub: Currently we're handling them as DRAM nodes with 0 capacity.

askervin (approved):
LGTM.

Thank you for this huge work, @klihub!

I think we should take it in and start developing on top of it.

klihub:
> LGTM.
> 
> Thank you for this huge work, @klihub!
> 
> I think we should take it in and start developing on top of it.

Yeah, I tend to think so, too. Then we'll need to fix bugs and polish the turd further as we go. I'm just unable to focus enough to get this into any better shape now.

PR #334, opened 2024-06-18
build: allow overriding set of unit test to run.

klihub:
Let the set of golang unit test be overridden using the TEST_PKGS Makefile variable. It defaults to '.' which will cause all tests to be run.


askervin (approved):
LGTM

PR #336, opened 2024-06-20
topology-aware: relax NUMA node topology checks.

klihub:
Relax hardware topology checks to allow multiple dies sharing a NUMA node, omitting die nodes from the pool tree if necessary on such hardware.


askervin (approved):
LGTM

PR #337, opened 2024-06-20
sysfs,cpuallocator: fix CPU cluster discovery. 

klihub:
This patch series fixes a few problems related to CPU cluster discovery. In particular
- on systems without SMT/hyperthreading use physical clusters as the logical ones, ensuring that we always have logical clusters if we have any
- during CPU cluster discovery, use the correct die ID instead of the incorrect package ID when querying per package CPU clusters


askervin (approved):
LGTM

PR #339, opened 2024-06-21
sysfs: survive NUMA nodes without memory.

klihub:
Don't fail for CPU-only NUMA nodes. Treat them as DRAM nodes with 0 bytes of memory attached, but exclude them when calculating average memory size for memory type approximation.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -1183,6 +1183,15 @@ func (sys *system) discoverNodes() error {
 		return fmt.Errorf("failed to parse nodes with memory (%q): %v",
 			memoryNodeIDs, err)
 	}
+	onlineNodeIDs, err := readSysfsEntry(sysNodesPath, "online", nil)
+	if err != nil {
+		return fmt.Errorf("failed to discover online nodes: %v", err)
+	}
+	onlineNodes, err := cpuset.Parse(onlineNodeIDs)
+	if err != nil {
+		return fmt.Errorf("failed to parse online nodes (%q): %v",
+			onlineNodeIDs, err)
+	}
 
 	cpuNodesSlice := []int{}
 	for id, node := range sys.nodes {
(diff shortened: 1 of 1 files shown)


pkg/sysfs/system.go
```diff
@@ -1183,6 +1183,15 @@ func (sys *system) discoverNodes() error {
 		return fmt.Errorf("failed to parse nodes with memory (%q): %v",
 			memoryNodeIDs, err)
 	}
+	onlineNodeIDs, err := readSysfsEntry(sysNodesPath, "online", nil)
```
askervin, line 1186: There seems to be a corner case: if there is a memoryless NUMA node but its CPUs are offlined, Linux still shows the node in the "online" list (tested with linux 6.5.6-300.fc39.x86_64). As a result, this logic includes the node in the noMemNodes list, as through that into dramNodes list, too.

On line 1212 we calculate dramNodes as "memory nodes with CPUs + nodes without memory". I'd suggest to simplify this as "nodes with CPUs", and fill it from the "has_cpu" list. As a result we would avoid getting nodes without (online) CPUs and memory included in dramNodes and thereby avoid accidentally pinning workloads without cpu/memory requests into a node that has neither.

pkg/sysfs/system.go
```diff
@@ -1224,7 +1234,7 @@ func (sys *system) discoverNodes() error {
 				dramTotal += info.MemTotal
 			}
 		}
-		dramAvg = dramTotal / uint64(len(dramNodeIds))
+		dramAvg = dramTotal / uint64(len(dramNodeIds)-noMemNodes.Size())
```
askervin, line 1237: ...need to be cautious with division by zero here. If there is a risk, let's error out with "unsupported memory configuration" rather than runtime error.

pkg/sysfs/system.go
klihub: Since you want to filter out noMemNodes without CPUs, I think it would be much more straightforward to just change line 1212 from the 
current intersection+union `dramNodes := memoryNodes.Intersection(cpuNodes).Union(noMemNodes)`
to first union then intersect `dramNodes := memoryNodes.Union(noMemNodes).Intersection(cpuNodes)`

WDYT ?

pkg/sysfs/system.go
klihub: Added an explicit check for a 0 divisor and bailout with an error if it happens.

pkg/sysfs/system.go
askervin: Isn't this the same as `dramNodes := cpuNodes`?

klihub: Hmm... yes, I think it should be.

askervin (approved):
LGTM

fmuyassarov (approved):
LGTM

pfl (approved):

PR #340, opened 2024-06-21
sysfs: allow non-uniform thread count.

klihub:
Replace ThreadCount() with {Min,Max}ThreadCount() to better match hybrid core architectures.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -102,7 +102,8 @@ type System interface {
 	SocketCount() int
 	CPUCount() int
 	NUMANodeCount() int
-	ThreadCount() int
+	MinThreadCount() int
+	MaxThreadCount() int
 	CPUSet() cpuset.CPUSet
 	Package(id idset.ID) CPUPackage
 	Node(id idset.ID) Node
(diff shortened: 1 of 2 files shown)


askervin (approved):
LGTM

pkg/sysfs/system.go
```diff
@@ -913,15 +922,17 @@ func (sys *system) discoverCPU(path string) error {
...
-	if cpu.threads.Size() > sys.threads {
-		sys.threads = cpu.threads.Size()
-	}
-
 	sys.cpus[cpu.id] = cpu
 
+	if threadCnt := cpu.threads.Size(); threadCnt > 0 {
+		if threadCnt < sys.minThreads {
+			sys.minThreads = threadCnt
+		}
```
pfl, line 930: If sys.minThreads is initialized to 1, then only a threadCnt of zero or less than zero will update its value. If the minimal threadCnt over all cpu.threads.Size() is for example 2, is it still ok to have min.Threads as 1?

klihub: Thanks for spotting that. Updated.

pfl (approved):

fmuyassarov (approved):
LGTM

PR #342, opened 2024-06-24
e2e: fix broken ansible playbook (ubuntu)

fmuyassarov:
Fix broken Ansible playbook (Ubuntu host).

### DIFF
--- a/test/e2e/playbook/provision.yaml
+++ b/test/e2e/playbook/provision.yaml
@@ -52,17 +52,34 @@
         - swapoff --all
       when: ansible_swaptotal_mb > 0
 
-    - name: Download public signing key for Kubernetes
-      ansible.builtin.apt_key:
-        url: https://pkgs.k8s.io/core:/stable:/v{{ k8s_version }}/deb/Release.key
-        state: present
+    - name: Add Kubernetes APT repository to sources.list.d
+      ansible.builtin.copy:
+        content: "deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v{{ k8s_version }}/deb/ /"
+        dest: /etc/apt/sources.list.d/kubernetes.list
+        mode: '0644'
       when: ansible_facts['distribution'] == "Ubuntu"
 
-    - name: Add apt repository for Kubernetes
-      ansible.builtin.apt_repository:
-        repo: "https://pkgs.k8s.io/core:/stable:/v{{ k8s_version }}/deb/ /"
-        state: present
-        filename: /etc/apt/sources.list.d/kubernetes.list
+    - name: Download Kubernetes APT key
+      ansible.builtin.get_url:
+        url: https://pkgs.k8s.io/core:/stable:/v{{ k8s_version }}/deb/Release.key
+        dest: /tmp/kubernetes-release.key
+        mode: '0644'
+      when: ansible_facts['distribution'] == "Ubuntu"
+
+    - name: Convert the Kubernetes APT key to keyring format
+      ansible.builtin.command:
+        cmd: gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg /tmp/kubernetes-release.key
+      when: ansible_facts['distribution'] == "Ubuntu"
+
+    - name: Clean up temporary key file
+      ansible.builtin.file:
+        path: /tmp/kubernetes-release.key
+        state: absent
+      when: ansible_facts['distribution'] == "Ubuntu"
+
+    - name: Update APT package index
+      ansible.builtin.apt:
+        update_cache: yes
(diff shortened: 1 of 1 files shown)


fmuyassarov:
Based on this, I guess that we don't run Ubuntu based e2e on our daily jobs @klihub . If that's true, perhaps we should think weather it would make sense to add it too or would it be too expensive to maintain it.

klihub:
> Based on this, I guess that we don't run Ubuntu based e2e on our daily jobs @klihub . If that's true, perhaps we should think weather it would make sense to add it too or would it be too expensive to maintain it.

If you don't specify a distro, the e2e tests default to `fedora`. In our nightly runs we don't override it. Whether it would be too expensive to run/maintain tests also on Ubuntu depends on whether we consider it useful enough to put in the extra effort to keep it up to date. Our e2e tests are functional tests so I don't see that running them on Ubuntu as well would contribute much to the original purpose per se. That said, if we want to keep our e2e test set runnable with a simple additional `distro=ubuntu-x.y` command line argument, then it probably would make sense to run some of the tests on ubuntu, to make sure the ubuntu-specific bits of bootstrapping the test env. et al. work.

test/e2e/playbook/provision.yaml
```diff
@@ -85,7 +130,6 @@
           - kubeadm
           - kubectl
           - systemd-container
-          - helm
```
klihub, line 88: I think this breaks our tests (on the default fedora VM). It stops installing Helm explicitly, so unless something pulls it in as a dependency (which I think won't happen) we'll end up without helm in the VM.

Could you update this PR to add the helm repo to the sources.list.d and then pull it in among the common packages ? We're fine with doing that on fedora test VMs, so it should be fine on Ubuntu as well.

Here are the [necessary details](https://helm.sh/docs/intro/install/#from-apt-debianubuntu) for doing it.

klihub: Ok, so the changes unconditionally install from release tarballs. So I think this should be fine after all.

klihub (approved):
For future reference, ideally these should have been two commits, one fixing the K8s repo on Ubuntu, and another switching unconditionally to installing Helm from release tarballs and pointing this out in the commit message for hasty reviewers like myself.

fmuyassarov:
> For future reference, ideally these should have been two commits, one fixing the K8s repo on Ubuntu, and another switching unconditionally to installing Helm from release tarballs and pointing this out in the commit message for hasty reviewers like myself.

Done.

klihub:
@askervin: PTAL

fmuyassarov:
Do you think we can merge the PR?

askervin:
```
$ distro=generic/ubuntu2204 ./run_tests.sh policies.test-suite/balloons/n4c16/test01-basic-placement/
...
TASK [Download public signing key for Kubernetes] ******************************
[WARNING]: Module remote_tmp /root/.ansible/tmp did not exist and was created
with a mode of 0700, this may cause issues when running as another user. To
avoid this, create the remote_tmp dir with the correct permissions manually
fatal: [n4c16-generic-ubuntu2204-containerd]: FAILED! => {"changed": false, "msg": "Failed to download key at
https://pkgs.k8s.io/core:/stable:/v1.30/deb/Release.key: Request failed: <urlopen error [Errno -3]
Temporary failure in name resolution>"]
```

askervin:
Does the "Download public signing key for Kubernetes" work for you?  And is it still needed now that this patch adds steps "Download Kubernetes APT key" and "Convert the Kubernetes APT key to keyring format"?

fmuyassarov:
> Does the "Download public signing key for Kubernetes" work for you? And is it still needed now that this patch adds steps "Download Kubernetes APT key" and "Convert the Kubernetes APT key to keyring format"?

@askervin Can you try once more please with the latest state?

askervin:
Found the root cause for the error. It was in my environment. I didn't provide DNS server address, but that seems to be required when using Ubuntu in a network environment where access to global DNS servers is restricted.

**dns_nameserver=DNS-IP-ADDRESS** distro=generic/ubuntu2204 ./run_tests.sh policies.test-suite/balloons/n4c16/test01-basic-placement/

test/e2e/playbook/provision.yaml
```diff
@@ -74,6 +91,29 @@
...
+      ansible.builtin.command:
+        cmd: /tmp/get_helm.sh
+      environment:
+        VERIFY_CHECKSUM: "false"
+
+    - name: Clean up the Helm install script
+      ansible.builtin.file:
+        path: /tmp/get_helm.sh
+        state: absent
+
```
askervin, line 116: There is trailing whitespace (spaces before end-of-line) in line 116. Would you remove it, and we are good to go with this patch.

fmuyassarov: fixed

askervin (approved):
LGTM, Thank you @fmuyassarov!

PR #343, opened 2024-06-26
cpuallocator: implement clustered allocation based on cache groups.

klihub:
This patch series adds support for clustered CPU allocation based on cache groups to the cpuallocator package. In particular with these patches in place, the cpuallocator now
- discovers cache groups on instantiation
- tries to satisfy allocation requests using one or more cache groups
- uses explicit cluster info based allocation only on hybrid core architectures

Allocation always prefers using fully idle cache groups first, fragmenting idle cache groups by partial allocation second, and only resort to using fragmented groups when the allocation cannot be satisfied in any other way. Partial cache group allocations are done in a hyperthread aware fashion, taking full physical cores whenever possible.

The allocator uses the last useful level of cache for clustering. IOW it tries to find the last cache which provides non-trivial clustering, one that provides different grouping than packages, dies, or hyperthreading, and uses that for grouping CPUs. The current implementation is currently somewhat simplistic as it expects all CPUs to provide identical cache grouping and therefore picks a single cache level for grouping. This is fine on most architectures, but it might result in suboptimal clustering with hybrid cores. If necessary this limitation can be addressed in the future.

### DIFF
--- a/pkg/cpuallocator/allocator.go
+++ b/pkg/cpuallocator/allocator.go
@@ -93,6 +93,7 @@ type topologyCache struct {
 
 	cpuPriorities cpuPriorities // CPU priority mapping
 	clusters      []*cpuCluster // CPU clusters
+	cacheGroups   []*cacheGroup // CPU cache groups
 }
 
 type cpuPriorities [NumCPUPriorities]cpuset.CPUSet
@@ -103,6 +104,15 @@ type cpuCluster struct {
 	cluster idset.ID
 	cpus    cpuset.CPUSet
 	kind    sysfs.CoreKind
+}
+
+type cacheGroup struct {
+	id   int
+	pkg  idset.ID
+	die  idset.ID
+	node idset.ID
+	cpus cpuset.CPUSet
+	kind sysfs.CoreKind
 }
 
 // IDFilter helps filtering Ids.
@@ -706,6 +716,7 @@ func newTopologyCache(sys sysfs.System) topologyCache {
 	}
 
 	c.discoverCPUClusters(sys)
+	c.discoverCacheGroups(sys)
 	c.discoverCPUPriorities(sys)
 
 	return c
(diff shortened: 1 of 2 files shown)


pkg/cpuallocator/allocator.go
```diff
@@ -389,6 +402,526 @@ func (a *allocatorHelper) takeIdleClusters() {
...
+		return
+	}
+
+	//
+	// take full idle cache groups, splitting up the last one if necessary
+	//
+
+	log.Debug("trying to take idle cache groups...")
+	for i, g := range sorter.prefer {
+		if cnt == 0 {
```
askervin, line 727: `cnt <= 0` would look safer than `cnt == 0` here and on line 757.

pkg/cpuallocator/allocator.go
```diff
@@ -389,6 +402,526 @@ func (a *allocatorHelper) takeIdleClusters() {
...
+
+	var (
+		groupsBySize = map[int][]*cacheGroup{}
+		totalByIndex = make([]int, 0, len(sorter.usable))
+		totalCPUs    = 0
+	)
+
+	for i := 0; i <= len(sorter.usable)-1; i++ {
+		g := sorter.usable[i]
+		cset := sorter.cpus[g]
```
askervin, line 783: Question: is it possible that due to CPU offlining or fully used cache groups or something else it could be that cset.Size() == 0?

If so, maybe these groups should be skipped over here to avoid the risk of division-by-zero on line 828.

pkg/cpuallocator/allocator.go
```diff
@@ -389,6 +402,526 @@ func (a *allocatorHelper) takeIdleClusters() {
...
+				pkgIdle := s.preferPkgCPUCount(idle.pkg)
+
+				if gA.pkg == pkgIdle && gB.pkg != pkgIdle {
+					return -1
+				}
+				if gA.pkg != pkgIdle && gB.pkg == pkgIdle {
+					return 1
+				}
+				if gA.pkg == pkgIdle && gB.pkg == pkgIdle {
+					if gA.die == dieIdle && gB.die != pkgIdle {
```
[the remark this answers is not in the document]
klihub: Ouch, should be .die != dieIdle. Thanks for spotting these.

pkg/cpuallocator/allocator.go
klihub: Updated.

pkg/cpuallocator/allocator.go
klihub: But I can add a `grpSize > 0` safeguard check there...

pkg/sysfs/system.go
```diff
@@ -1096,6 +1097,25 @@ func (c *cpu) GetLastLevelCaches() []*Cache {
...
+	if len(c.caches) < 1 {
+		return c.ThreadCPUSet()
+	}
+
+	cpus := cpuset.New()
+
+	for _, cch := range c.caches {
+		if cch.level == n {
+			cpus = cpus.Union(CPUSetFromIDSet(cch.cpus))
+		} else if cch.level > n {
```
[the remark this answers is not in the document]
klihub: No, `cpu.caches` are sorted, but they are sorted by index, not by level. This also guarantees that they are sorted by level as well. I can add a `if cch.level > level` breakout condition to terminate the loop early.

It is also true that `cpu.caches` are sorted because they are discovered from sysfs by globbing and sysfs returns them by lexiographic index order. So if there ever were more than 10 cache levels (I seriously doubt this would ever happen), then this breaks as lexiographic order != index order. I can add an explicit sort for them during discovery.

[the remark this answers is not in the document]
klihub: Added explicit sorting of caches by index.

pkg/cpuallocator/allocator.go
klihub: Added.

askervin (approved):
Looks Great. Thanks @klihub!

PR #344, opened 2024-06-26
 balloons: add support for isolated cpus.

fmuyassarov:
The isolcpus boot parameter allows you to isolate specific CPUs from
the general SMP balancing and scheduler algorithms. Currently, when a
balloon picks up CPUs, it does not differentiate between isolated and
non-isolated CPUs. This change improves the balloon's logic to:
1. Distinguish between isolated and non-isolated CPUs.
2. Allow users to preferably select isolated CPUs, though this is not guaranteed.
3. Avoid using isolated CPUs unless specifically requested by the user via CR.

### DIFF
--- a/docs/resource-policy/policy/balloons.md
+++ b/docs/resource-policy/policy/balloons.md
@@ -168,6 +168,14 @@ Balloons policy parameters:
     preferring exclusive CPUs, as long as there are enough free
     CPUs. The default is `false`: prefer filling and inflating
     existing balloons over creating new ones.
+  - `preferIsolCpus`: if `true`,prefer system isolated CPUs (refer to
+    kernel command line parameter "isolcpus") for this balloon. Warning:
+    if there are not enough isolated CPUs in the system for balloons that
+    prefer them, balloons may include normal CPUs, too. This kind of
+    mixed-CPU balloons require special attention when implementing
+    programs that run on them. Therefore it is recommended to limit the
+    number of balloon CPUs (see maxCPUs) and allocate CPUs upfront (see
+    minBalloons, minCPUs) when using preferIsolCpus. The default is `false`.
   - `shareIdleCPUsInSame`: Whenever the number of or sizes of balloons
     change, idle CPUs (that do not belong to any balloon) are reshared
     as extra CPUs to containers in balloons with this option. The value
--- a/test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/balloons-isolcpus.cfg
+++ b/test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/balloons-isolcpus.cfg
@@ -1,0 +1,13 @@
+config:
+  balloonTypes:
+    - name: isolcpus
+      minCPUs: 1
+      minBalloons: 1
+      preferIsolCpus: true
+      namespaces:
+      - isolcpus
+  log:
+    debug:
+      - policy
+    klog:
+      skip_headers: true
--- a/test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
+++ b/test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
@@ -1,0 +1,65 @@
+reboot-node() {
+    vm-reboot
+    timeout=600 host-wait-vm-ssh-server $OUTPUT_DIR
+}
+
+restart-kubelet() {
+    vm-command "systemctl restart kubelet"
+    sleep 5
+    vm-wait-process --timeout 120 kube-apiserver
+    vm-command "cilium status --wait --wait-duration=120s --interactive=false"
+}
+
+cleanup-pods() {
+    vm-command "kubectl delete pod --all --now"
+    vm-command "kubectl delete namespace $ns --now"
+}
+
+cleanup() {
+    cleanup-pods
+    if [[ "$distro" == *fedora* ]]; then
+        fedora-set-kernel-cmdline ""
+    else
+        ubuntu-set-kernel-cmdline ""
+    fi
+    reboot-node
+    vm-command "grep -v isolcpus /proc/cmdline"
+    if [ $? -ne 0 ]; then
+        error "failed to unset isolcpus kernel commandline parameter"
+    fi
+    restart-kubelet
+    return 0
+}
+
+ns=isolcpus
+cleanup-pods
+
+vm-command "grep isolcpus=0,1 /proc/cmdline" || {
+    if [[ "$distro" == *fedora* ]]; then
+        fedora-set-kernel-cmdline "isolcpus=0,1"
+    else
+        ubuntu-set-kernel-cmdline "isolcpus=0,1"
+    fi
+    reboot-node
+    vm-command "grep isolcpus=0,1 /proc/cmdline" || {
+        error "failed to set isolcpus kernel commandline parameter"
+    }
+    restart-kubelet
+}
+
+helm-terminate
+helm_config=${TEST_DIR}/balloons-isolcpus.cfg helm-launch balloons
+vm-command "kubectl create namespace $ns"
+
+# pod0: should run on non-isolated CPUs
+CONTCOUNT=2 namespace="default" create balloons-busybox
+report allowed
+verify "set.union(cpus['pod0c0'], cpus['pod0c1']).isdisjoint({'cpu00', 'cpu01'})"
+
+# pod1: runs on system isolated CPUs
+CONTCOUNT=2 namespace="$ns" create balloons-busybox
+report allowed
+verify "cpus['pod1c0'] == {'cpu00', 'cpu01'}" 
+verify "cpus['pod1c1'] == {'cpu00', 'cpu01'}"
+
+cleanup
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -915,6 +915,62 @@ $py_assertion
     return 0
 }
 
+host-wait-vm-ssh-server() {
+    local _vagrantdir="$1"
+    local _timeout=${timeout:-10}
+    local _waited
+
+    if [ -z "$_vagrantdir" ]; then
+        echo 1>&2 "host-wait-vm-ssh-server: missing vagrant directory"
+        return 1
+    fi
+
+    _waited=0; while (( $_waited < $_timeout )); do
+        if [ ! -f $_vagrantdir/.ssh-config ]; then
+            sleep 1
+        else
+            $SSH -o ConnectTimeout=1 node true
+            if [ $? == 0 ]; then
+                return 0
+            fi
+        fi
+        _waited=$((_waited + 1))
+    done
+
+    echo 1>&2 "host-wait-vm-ssh-server: timeout waiting for $_vagrantdir ssh server"
+    return 1
+}
+
+fedora-set-kernel-cmdline() {
+    local e2e_defaults="$*"
+    vm-command "mkdir -p /etc/default; touch /etc/default/grub; sed -i '/e2e:fedora-set-kernel-cmdline/d' /etc/default/grub"
+    vm-command "echo 'GRUB_CMDLINE_LINUX_DEFAULT=\"\${GRUB_CMDLINE_LINUX_DEFAULT} ${e2e_defaults}\" # by e2e:fedora-set-kernel-cmdline' >> /etc/default/grub" || {
+        command-error "writing new command line parameters failed"
+    }
+    vm-command "grub2-mkconfig -o /boot/grub2/grub.cfg" || {
+        command-error "updating grub failed"
+    }
+}
+
+ubuntu-set-kernel-cmdline() {
+    local e2e_defaults="$*"
+    vm-command "echo 'GRUB_CMDLINE_LINUX_DEFAULT=\"\${GRUB_CMDLINE_LINUX_DEFAULT} ${e2e_defaults}\"' > /etc/default/grub.d/60-e2e-defaults.cfg" || {
+        command-error "writing new command line parameters failed"
+    }
+    vm-command "update-grub" || {
+        command-error "updating grub failed"
+    }
+}
+
+vm-reboot() { # script API
+    # Usage: vm-reboot
+    #
+    # Reboots the virtual machine and waits that the ssh server starts
+    # responding again.
+    vm-command "reboot"
+    sleep 5
+}
+
 # Defaults to use in case the test case does not define these values.
 yaml_in_defaults="CPU=1 MEM=100M ISO=true CPUREQ=1 CPULIM=2 MEMREQ=100M MEMLIM=200M CONTCOUNT=1"
 
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -53,6 +53,9 @@ const (
 	// virtDevReservedCpus is the name of a virtual device close to
 	// CPUs that are configured as ReservedResources.
 	virtDevReservedCpus = "reserved CPUs"
+	// virtDevIsolatedCpus is the name of a virtual device close to
+	// host isolated CPUs.
+	virtDevIsolatedCpus = "isolated CPUs"
 )
 
 // balloons contains configuration and runtime attributes of the balloons policy
@@ -558,6 +561,7 @@ func (p *balloons) newBalloon(blnDef *BalloonDef, confCpus bool) (*Balloon, erro
 		preferFarFromDevices:        blnDef.PreferFarFromDevices,
 		virtDevCpusets: map[string][]cpuset.CPUSet{
 			virtDevReservedCpus: {p.reserved},
+			virtDevIsolatedCpus: {p.options.System.Isolated()},
 		},
 	}
 	if blnDef.AllocatorTopologyBalancing != nil {
@@ -994,6 +998,9 @@ func (p *balloons) validateConfig(bpoptions *BalloonsOptions) error {
 					blnDef.Name, blnDef.MaxBalloons)
 			}
 		}
+		if blnDef.PreferIsolCpus && blnDef.ShareIdleCpusInSame != "" {
+			log.Warn("WARNING: using PreferIsolCpus with ShareIdleCpusInSame is highly discouraged")
+		}
 	}
 	return nil
 }
@@ -1030,6 +1037,7 @@ func (p *balloons) setConfig(bpoptions *BalloonsOptions) error {
 	if err = p.validateConfig(bpoptions); err != nil {
 		return balloonsError("invalid configuration: %w", err)
 	}
+	p.fillCloseToDevices(bpoptions.BalloonDefs)
 	p.fillFarFromDevices(bpoptions.BalloonDefs)
 
 	// Preparation and configuration validation is now done
@@ -1189,6 +1197,14 @@ func (p *balloons) fillBuiltinBalloonDefs(bpoptions *BalloonsOptions) (*BalloonD
 	return reservedBalloonDef, defaultBalloonDef, nil
 }
 
+func (p *balloons) fillCloseToDevices(blnDefs []*BalloonDef) {
+	for _, blnDef := range blnDefs {
+		if blnDef.PreferIsolCpus {
+			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevIsolatedCpus)
+		}
+	}
+}
+
 // fillFarFromDevices adds BalloonDefs implicit device anti-affinities
 // towards devices that other BalloonDefs prefer to be close to.
 func (p *balloons) fillFarFromDevices(blnDefs []*BalloonDef) {
@@ -1201,6 +1217,9 @@ func (p *balloons) fillFarFromDevices(blnDefs []*BalloonDef) {
 	// beginning of the list will be more effectively avoided than
 	// devices later in the list.
 	avoidDevs := []string{}
+	if p.options.System.Isolated().Size() != 0 {
+		avoidDevs = append(avoidDevs, virtDevIsolatedCpus)
+	}
 	for _, blnDef := range blnDefs {
 		for _, closeDev := range blnDef.PreferCloseToDevices {
 			if _, ok := devDefClose[closeDev]; !ok {
@@ -1361,6 +1380,7 @@ func (p *balloons) shareIdleCpus(addCpus, removeCpus cpuset.CPUSet) []*Balloon {
 			}
 		}
 	}
+	addCpus = addCpus.Difference(p.options.System.Isolated())
 	if addCpus.Size() > 0 {
 		for blnIdx, bln := range p.balloons {
 			topoLevel := bln.Def.ShareIdleCpusInSame
--- a/config/crd/bases/config.nri_balloonspolicies.yaml
+++ b/config/crd/bases/config.nri_balloonspolicies.yaml
@@ -195,6 +195,10 @@ spec:
                       items:
                         type: string
                       type: array
+                    preferIsolCpus:
+                      default: false
+                      description: 'preferIsolCpus: prefer kernel isolated cpus'
+                      type: boolean
                     preferNewBalloons:
                       description: |-
                         PreferNewBalloons: prefer creating new balloons over adding
--- a/deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
+++ b/deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
@@ -195,6 +195,10 @@ spec:
                       items:
                         type: string
                       type: array
+                    preferIsolCpus:
+                      default: false
+                      description: 'preferIsolCpus: prefer kernel isolated cpus'
+                      type: boolean
                     preferNewBalloons:
                       description: |-
                         PreferNewBalloons: prefer creating new balloons over adding
--- a/pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
+++ b/pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
@@ -216,6 +216,9 @@ type BalloonDef struct {
 	// TODO: PreferFarFromDevices is considered too untested for usage. Hence,
 	// for the time being we prevent its usage through CRDs.
 	PreferFarFromDevices []string `json:"-"`
+	// preferIsolCpus: prefer kernel isolated cpus
+	// +kubebuilder:default:=false
+	PreferIsolCpus bool `json:"preferIsolCpus,omitempty"`
 }
 
 // String stringifies a BalloonDef


test/e2e/run.sh
```diff
@@ -874,6 +874,34 @@ $py_assertion
...
+    # Usage: vm-set-kernel-cmdline E2E-DEFAULTS
+    #
+    # Adds/replaces E2E-DEFAULTS to kernel command line"
+    #
+    # Example:
+    #   vm-set-kernel-cmdline nr_cpus=4
+    #   vm-reboot
+    #   vm-command "cat /proc/cmdline"
+    local e2e_defaults="$1"
+    vm-command "echo 'GRUB_CMDLINE_LINUX_DEFAULT=\"\${GRUB_CMDLINE_LINUX_DEFAULT} ${e2e_defaults}\"' > /etc/default/grub.d/60-e2e-defaults.cfg" || {
```
klihub, line 887: Have you tested that this works on a fedora VM, too ?

test/e2e/run.sh
```diff
@@ -874,6 +874,34 @@ $py_assertion
...
+        command-error "writing new command line parameters failed"
+    }
+    vm-command "update-grub" || {
```
klihub, line 890: ditto for this

test/e2e/run.sh
fmuyassarov: No I tested only on Ubuntu. But I will verify on Fedora too.

fmuyassarov: well it doesn't, welcome to Fedora :). I have no idea how to do it on Fedora since I don't use it. But I will do some search and try to find a way.

klihub: On Fedora this is really easy using grubby.

test/e2e/run.sh
fmuyassarov: Fixed.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,33 @@
+helm-terminate
+helm_config=${TEST_DIR}/balloons-isolcpus.cfg helm-launch balloons
+
+vm-command "grep isolcpus=0,1 /proc/cmdline" || {
+    if [ "$distro" = "generic/fedora37" ]; then
```
klihub, line 5: Let's not assume it's fedora37 here, let's just check for fedora.

test/e2e/run.sh
```diff
@@ -874,6 +874,36 @@ $py_assertion
...
+    }
+}
+
+vm-reboot() { # script API
+    # Usage: vm-reboot
+    #
+    # Reboots the virtual machine and waits that the ssh server starts
+    # responding again.
+    vm-command "reboot"
+    sleep 5
```
askervin, line 904: The caller of `vm-reboot` expects in this test (and it's safe to assume in other tests, too) that the vm is up and running and ready for next commands immediately when this function returns. 

In cri-resource-manager we had a function where we (on the host) wait that vm responds. If not, it's better to error out with clear error message than let the test script continue and fail later when the root cause is harder to debug.

See:
https://github.com/intel/cri-resource-manager/blob/master/demo/lib/vm.bash#L622

(You don't need to copy vm-monitor-based forced reboot here, but waiting for ssh connection to be available is a good idea.)

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,33 @@
...
+    fi
+    vm-reboot
+    vm-command "grep isolcpus=0,1 /proc/cmdline" || {
+        error "failed to set isolcpus kernel commandline parameter"
+    }
+    vm-command "systemctl restart kubelet"
+    sleep 5
+    vm-wait-process --timeout 120 kube-apiserver
+}
+vm-command "kubectl create namespace isolcpus"
```
askervin, line 18: Two namespaces are created but not used. These should be cleaned up when exiting the test script.

I'd suggest changing the balloons configuration so that preferIsolCpus would be used in a balloon type that is not applied to all containers in the default namespace. (I'd suggest this because placing all default stuff to isolcpus sounds suspicious configuration, and I would not like to use a suspicious configuration even in an e2e test.)

If you change balloons-isolcpus.cfg so that only containers in the isolcpus namespace will be put into the isolcpus balloon, then there is use for this namespace and the configuration will be a better example, too.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,33 @@
...
+CONTCOUNT=2 namespace=default create balloons-busybox
+report allowed
+verify "cpus['pod0c0'] == {'cpu00', 'cpu01'}"
+
+# pod1: should run on non-isolated CPUs
+CONTCOUNT=1 namespace="kube-system" create balloons-busybox
+report allowed
+verify "cpus['pod1c0'] != {'cpu00', 'cpu01'}"
+
+cleanup
```
askervin, line 31: cleanup implementation is missing.

delete pods and namespaces you created and restore kernel parameters. otherwise other tests will be in trouble if they use the same vm.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
fmuyassarov: > Two namespaces are created but not used. These should be cleaned up when exiting the test script.
> 

Good catch. I will remove them.

> I'd suggest changing the balloons configuration so that preferIsolCpus would be used in a balloon type that is not applied to all containers in the default namespace. (I'd suggest this because placing all default stuff to isolcpus sounds suspicious configuration, and I would not like to use a suspicious configuration even in an e2e test.)
> 

Running pods in any namespaces than `kube-system`, `default` requires patching RBAC. For that reason I decided to deploy the workload in the `default` namespaces so that I don't have to introduce extra changes to RBAC.
If you think there is no difference between running the workload in `default` namespace vs `foo`, because we can just imagine that `default` namespace is a special namespace in this particular test that is meant to host only specific workload that requires isolated CPUs. Because, at the time of deploying workload on `default` namespace, there is no other Pod running on that namespace and as such I think it is safe to consider this namespace as special. But if there is any other reason to not pick the `default` namespace, I can extend the RBAC permissions.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/balloons-isolcpus.cfg
```diff
@@ -0,0 +1,17 @@
+config:
+  balloonTypes:
+    - name: isolcpus
+      minBalloons: 0
+      minCPUs: 1
+      minBalloons: 1
```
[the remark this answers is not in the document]
fmuyassarov: Good catch.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
fmuyassarov: Fixed.

test/e2e/run.sh
fmuyassarov: I have tried but it seems that the `host-wait-vm-ssh-server` function is broken. Do you prefer that function to be here or is it fine to have my own custom func?

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
klihub: > > Two namespaces are created but not used. These should be cleaned up when exiting the test script.
> 
> Good catch. I will remove them.
> 
> > I'd suggest changing the balloons configuration so that preferIsolCpus would be used in a balloon type that is not applied to all containers in the default namespace. (I'd suggest this because placing all default stuff to isolcpus sounds suspicious configuration, and I would not like to use a suspicious configuration even in an e2e test.)
> 
> Running pods in any namespaces than `kube-system`, `default` requires patching RBAC. For that reason I decided to deploy the workload in the `default` namespaces so that I don't have to introduce extra changes to RBAC.

I don't think this is true for our e2e test setup. There the test user is running with the same kubeconfig as the cluster admin who bootstrapped the cluster. We have several test cases that create pods in the `kube-system` namespace successfully without any extra/special authorization.

klihub (commented):
Once all review comments are addressed I'm happy with this as soon as @askervin finds it good to go in.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
fmuyassarov: Have you tried in any other namespace which not kube-system or default  ?

klihub: Unless the test shell library are broken and it does not really use namespaces in those cases...

klihub: Which does not seem to be the case... sampling a test run I got among other things this. You can see a non-default, non-kube-system namespace pod there (in ns 'three'):
```
[vagrant@n4c16-generic-fedora37-containerd ~]$ kubectl get pods -A
NAMESPACE     NAME                                                        READY   STATUS        RESTARTS      AGE
default       pod1                                                        2/2     Running       0             7s
default       pod2                                                        1/1     Running       0             5s
kube-system   cilium-hhxcw                                                1/1     Running       1 (97s ago)   27d
kube-system   cilium-operator-6df6cdb59b-bgdzm                            1/1     Running       1 (97s ago)   27d
kube-system   coredns-7db6d8ff4d-gwf6x                                    1/1     Running       1 (97s ago)   27d
kube-system   coredns-7db6d8ff4d-r2c6l                                    1/1     Running       1 (97s ago)   27d
kube-system   etcd-n4c16-generic-fedora37-containerd                      1/1     Running       1 (97s ago)   27d
kube-system   kube-apiserver-n4c16-generic-fedora37-containerd            1/1     Running       1 (97s ago)   27d
kube-system   kube-controller-manager-n4c16-generic-fedora37-containerd   1/1     Running       1 (97s ago)   27d
kube-system   kube-proxy-b6ms9                                            1/1     Running       1 (97s ago)   27d
kube-system   kube-scheduler-n4c16-generic-fedora37-containerd            1/1     Running       1 (97s ago)   27d
kube-system   nri-resource-policy-balloons-wgcmp                          1/1     Running       0             12s
kube-system   pod0                                                        2/2     Terminating   0             9s
three         pod3                                                        2/2     Running       0             3s
```

test/e2e/run.sh
askervin: Is there something else broken than `vm-monitor`-based hard reset? I think that hard reset should be dropped anyway. This function should be about waiting for the ssh server to respond and not really doing any other tricks but waiting. I was hoping that you could add the function ((possibly directly from cri-resource-manager and just ripping off the vm-monitor part).

If there is something wrong with it, then it needs to be implemented in some other way. We'll anyway need some way to wait for virtual machines to be again accessible through ssh after reboots. Something that continues pretty soon when ssh works again, but can wait for some time for the vm to boot. That is, something better than `sleep 5` or `sleep 5m` and just wishing that the vm woke up.

klihub: The reason why `host-wait-vm-ssh-server` can't be lifted over as such from CRI Resource Manager e2e tests is that we now use vagrant instead of govm for test VMs. One stupidly simplistic way to update a `host-wait-vm-ssh-server` for the way we now use vagrant would be to have something like this:

```
host-wait-vm-ssh-server() {
    local _timeout=${timeout:-10}
    local _waited

    _waited=0; while [[ $_waited < $_timeout ]]; do
        if [ ! -e $OUTPUT_DIR/vagrant/.ssh ]]; then
            sleep 1
        else
            ssh -o ConnectTimeout=1 -F $OUTPUT_DIR/vagrant/.ssh-config true
            if [ $? == 0 ]; then
                return 0
            fi
        fi
        _waited=$((_waited + 1))
    done
}
```

askervin: Tiny fix: instead of `while [[ ... ]]` use `while (( ... ))`, because `[[ 2 < 10 ]]` is false (due to lexicographic order) but `(( 2 < 10 ))` is true.

docs/resource-policy/policy/balloons.md
```diff
@@ -168,6 +168,11 @@ Balloons policy parameters:
     preferring exclusive CPUs, as long as there are enough free
     CPUs. The default is `false`: prefer filling and inflating
     existing balloons over creating new ones.
+  - `preferIsolCpus`: if `true`, prefer system isolated CPUs for this
+    balloon. If the workload increases and requires more isolated CPUs
+    than are available, the balloon will inflate by taking additional
+    non-isolated CPUs as well. The default is `false`: don't prefer
+    isoleted CPUs.
```
askervin, line 175: s/isoleted/isolated

askervin (commented):
Nit: spotted a typo in documentation.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,49 @@
+helm-terminate
+helm_config=${TEST_DIR}/balloons-isolcpus.cfg helm-launch balloons
```
klihub, line 2: It does not make sense to launch the plugins here. Usually you don't have isolcpus set on the kernel command line at this point, which you later check and try to fix, once the policy is up and running. It make more sense to first check the kernel command line, fix it if there is a need, then reboot, verify the updated command line, and only after that install the policy.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,49 @@
...
+
+vm-command "grep isolcpus=0,1 /proc/cmdline" || {
+    if [[ "$distro" == *fedora* ]]; then
+        fedora-set-kernel-cmdline "isolcpus=0,1"
+    else
+        ubuntu-set-kernel-cmdline "isolcpus=0,1"
+    fi
+    vm-reboot
+    sleep 5
```
klihub, line 11: This will be unreliable. It would be better to have something like

```
reboot-node() {
    vm-reboot
    sleep 5
    timeout=120 host-wait-vm-ssh-server $OUTPUT_DIR
}
```

and then call `reboot-node` to reboot the node and wait for it to come up and be ready for ssh commands.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,49 @@
...
+    vm-command "grep isolcpus=0,1 /proc/cmdline" || {
+        error "failed to set isolcpus kernel commandline parameter"
+    }
+    vm-command "systemctl restart kubelet"
+    sleep 5
+    vm-wait-process --timeout 120 kube-apiserver
```
klihub, line 17: This is not enough. After a node reboot, none of the control plane containers are running. Waiting for the API server is not enough. It will take a long time after it is running before the node is really fully bootstrapped again.
Since ATM we always use cilium as our CNI, you could use something like this:

```
restart-kubelet() {
    vm-command "systemctl restart kubelet"
    sleep 5
    vm-wait-process --timeout 120 kube-apiserver
    vm-command "cilium status --wait --interactive=false"
}
```
and then call `restart-kubelet` to restart the kubelet and wait for the node to become service ready.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# pod1: should run on non-isolated CPUs
+CONTCOUNT=1 namespace="kube-system" create balloons-busybox
+report allowed
+verify "cpus['pod1c0'] != {'cpu00', 'cpu01'}"
+
+cleanup() {
+    vm-command "kubectl delete pods --all --now"
+    if [[ "$distro" == *fedora* ]]; then
+        fedora-set-kernel-cmdline "isolcpus="
```
klihub, line 33: IMO it would be better to reset the isolcpu setting by `{fedora,ubuntu}-set-kernel-cmdline ""`. That should leave the node properly in a state without any `isolcpus=` kernel command line option.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,49 @@
...
+report allowed
+verify "cpus['pod0c0'] == {'cpu00', 'cpu01'}"
+
+# pod1: should run on non-isolated CPUs
+CONTCOUNT=1 namespace="kube-system" create balloons-busybox
+report allowed
+verify "cpus['pod1c0'] != {'cpu00', 'cpu01'}"
+
+cleanup() {
+    vm-command "kubectl delete pods --all --now"
```
klihub, line 31: This is not enough, on a successful test you have placed pods in the `kube-system` namespace and those should be cleaned up as well.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
klihub, line 33: If make the above changes, or something similar then you won't need this as `cleanup` will take care of this for you.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,49 @@
...
+        error "failed to unset isolcpus kernel commandline parameter"
+    }
+    vm-command "systemctl restart kubelet"
+    sleep 5
+    vm-wait-process --timeout 120 kube-apiserver
+    return 0
+}
+cleanup
+
+helm-terminate
```
klihub, line 49: I think we usually don't do a helm-terminate at the end of tests, so I wouldn't do this here either (so it is not included in the suggested `cleanup` above).

klihub (changes requested):
I have a number of new review comments and suggestions related to the e2e test framework additions and the test scriptlet itself.

test/e2e/run.sh
```diff
@@ -874,6 +874,37 @@ $py_assertion
...
+    local e2e_defaults="$*"
+    vm-command "mkdir -p /etc/default; touch /etc/default/grub; sed -i '/e2e:fedora-set-kernel-cmdline/d' /etc/default/grub"
+    vm-command "sed -i '/${e2e_defaults}/d' /etc/default/grub"
+    vm-command "echo 'GRUB_CMDLINE_LINUX_DEFAULT=\"\${GRUB_CMDLINE_LINUX_DEFAULT} ${e2e_defaults}\"' >> /etc/default/grub" || {
+        command-error "writing new command line parameters failed"
+    }
+    vm-command "grub2-mkconfig -o /boot/grub2/grub.cfg" || {
+        command-error "updating grub failed"
+    }
+}
```
klihub, line 887: This is lifted over incorrectly from CRI Resource Manager and therefore does not work properly/well. The original version injects a well-known comment for itself, so if necessary it later can properly re-edit the same default file without creating duplicate entries variable entries. It should be something like this instead:

```
fedora-set-kernel-cmdline() {
    local e2e_defaults="$*"
    vm-command "mkdir -p /etc/default; touch /etc/default/grub; sed -i '/e2e:fedora-set-kernel-cmdline/d' /etc/default/grub"
    vm-command "echo 'GRUB_CMDLINE_LINUX_DEFAULT=\"\${GRUB_CMDLINE_LINUX_DEFAULT} ${e2e_defaults}\" # by e2e:fedora-set-kernel-cmdline' >> /etc/default/grub" || {
        command-error "writing new command line parameters failed"
    }
    vm-command "grub2-mkconfig -o /boot/grub2/grub.cfg" || {
        command-error "updating grub failed"
    }
}
```

test/e2e/run.sh
```diff
@@ -915,6 +915,73 @@ $py_assertion
...
+    local e2e_defaults="$*"
+    vm-command "mkdir -p /etc/default; touch /etc/default/grub; sed -i '/e2e:fedora-set-kernel-cmdline/d' /etc/default/grub"
+    vm-command "echo 'GRUB_CMDLINE_LINUX_DEFAULT=\"\${GRUB_CMDLINE_LINUX_DEFAULT} ${e2e_defaults}\" # by e2e:fedora-set-kernel-cmdline' >> /etc/default/grub" || {
+        command-error "writing new command line parameters failed"
+    }
+    vm-command "grub2-mkconfig -o /boot/grub2/grub.cfg" || {
+        command-error "updating grub failed"
+    }
+}
+
```
klihub, line 965: fedora-set-kernel-cmdline is now defined twice, only this one staying in effect.

klihub:
AFAICT, the tests are written so that both containers you create end up handled by the same balloon definition, but since the first one exhausts the available isolated CPUs (0,1) the 2nd one gets non-isolated ones. To me it feels like we are not testing the right thing. Or actually we are testing an unwanted corner case.

It would feel more natural to create the 2nd container in some other namespace than isolcpus and that's why it then would get another balloon and therefore other than isolated CPUs.

Or do I read the diffs wrong?

fmuyassarov:
> AFAICT, the tests are written so that both containers you create end up handled by the same balloon definition, but since the first one exhausts the available isolated CPUs (0,1) the 2nd one gets non-isolated ones. To me it feels like we are not testing the right thing. Or actually we are testing an unwanted corner case.
> 
> It would feel more natural to create the 2nd container in some other namespace than isolcpus and that's why it then would get another balloon (or the default) and therefore other than isolated CPUs.
> 
> Or do I read the diffs wrong?

Yes that's right. I fixed accordingly.

klihub (approved):
LGTM.

@fmuyassarov And thanks for bearing with us and being persistent throughout the review !

docs/resource-policy/policy/balloons.md
```diff
@@ -168,6 +168,11 @@ Balloons policy parameters:
     preferring exclusive CPUs, as long as there are enough free
     CPUs. The default is `false`: prefer filling and inflating
     existing balloons over creating new ones.
+  - `preferIsolCpus`: if `true`, prefer system isolated CPUs for this
```
askervin, line 171: It's a bit inaccurate to say "the workload increases" or "the workload _requires_ more isolated CPUs". A workload can at most wish to be put in a balloon by annotation, and a balloon can only _prefer_ isolcpus. Furthermore, I'd like to provide a warning and guidance in this documentation, because otherwise there are chances that a reader does not understand that here be dragons...

I'd suggest rewording the preferIsolCpus doc:

if `true`, prefer system isolated CPUs (refer to kernel command line parameter "isolcpus") for this balloon. Warning: if there are not enough isolated CPUs in the system for balloons that prefer them, balloons may include normal CPUs, too. This kind of mixed-CPU balloons require special attention when implementing programs that run on them. Therefore it is recommended to limit the number of balloon CPUs (see maxCPUs) and allocate CPUs upfront (see minBalloons, minCPUs) when using preferIsolCpus. The default is `false`.

fmuyassarov: Fixed.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,59 @@
...
+helm_config=${TEST_DIR}/balloons-isolcpus.cfg helm-launch balloons
+vm-command "kubectl create namespace $ns"
+
+# pod0: runs on system isolated CPUs
+CONTCOUNT=2 namespace="$ns" create balloons-busybox
+report allowed
+verify "cpus['pod0c0'] == {'cpu00', 'cpu01'}"
+
+# pod1: should run on non-isolated CPUs
```
askervin, line 37: There's one essential thing that this test should test:
**when** new balloons are created (or existing balloons inflated) **and** there are isolcpus available in the system,
**is it possible that** an isolcpu could end up into a balloon that does not prefer them.

Currently both isolcpus seem to be already allocated into the "isolcpus" balloon (based on line 35), and therefore it cannot happen that isolcpus would be allocated into new balloons (or inflated existing balloons).

Would you create pods in a different order? First create containers that need most (but not all non-isolcpus) and verify that none of isolcpus was used for that. That is, first create containers that need many CPUs and verify something like: `verify "set.union(cpus['pod0c0'], cpus['pod0c1'], ...).isdisjoint({'cpu00', 'cpu01'})"`

fmuyassarov: Fixed.

klihub (approved):
LGTM.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,60 @@
...
+        error "failed to set isolcpus kernel commandline parameter"
+    }
+    restart-kubelet
+}
+
+helm-terminate
+helm_config=${TEST_DIR}/balloons-isolcpus.cfg helm-launch balloons
+vm-command "kubectl create namespace $ns"
+
+# pod1: should run on non-isolated CPUs
```
klihub, line 32: nit: this is not pod0, not pod1

askervin (changes requested):
After passed test pod0 is left behind:

```
vagrant@n4c16-generic-ubuntu2204-containerd:~$ kubectl get pods -A
NAMESPACE     NAME                                                          READY   STATUS    RESTARTS      AGE
default       pod0                                                          2/2     Running   2 (92s ago)   2m10s
...
```

@klihub already suggested changes to `cleanup()` so that nothing is left behind if the test passed.

fmuyassarov:
> After passed test pod0 is left behind:
> 
> ```
> vagrant@n4c16-generic-ubuntu2204-containerd:~$ kubectl get pods -A
> NAMESPACE     NAME                                                          READY   STATUS    RESTARTS      AGE
> default       pod0                                                          2/2     Running   2 (92s ago)   2m10s
> ...
> ```
> 
> @klihub already suggested changes to `cleanup()` so that nothing is left behind if the test passed.

@askervin @klihub this should be addressed now. PTAL.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,68 @@
...
+    else
+        ubuntu-set-kernel-cmdline ""
+    fi
+    reboot-node
+    vm-command "grep -v isolcpus /proc/cmdline"
+    if [ $? -ne 0 ]; then
+        error "failed to unset isolcpus kernel commandline parameter"
+    fi
+    return 0
+    restart-kubelet
```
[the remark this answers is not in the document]
fmuyassarov: Fixed.

test/e2e/policies.test-suite/balloons/n4c16/test22-isolcpus/code.var.sh
```diff
@@ -0,0 +1,68 @@
...
+
+cleanup-pods() {
+    vm-command "kubectl delete pod --all --now"
+    vm-command "kubectl -n $ns delete pod --all now"
+    vm-command "kubectl delete namespace $ns --now"
+}
+
+cleanup() {
+    cleanup-pods
+    vm-command "kubectl delete namespace $ns"
```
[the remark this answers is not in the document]
fmuyassarov: Good point. Fixed.

askervin (approved):
LGTM. Thanks for this new feature, @fmuyassarov!

klihub (approved):

PR #345, opened 2024-06-26
cpuallocator: don't filter based on single CoreKind.

klihub:
Only do CoreKind filtering in clustered allocation when we have multiple CoreKinds available, IOW on hybrid core architectures.


fmuyassarov (commented):
This looks good based on what title says. But would you mind to share why is this needed?

fmuyassarov (approved):
Thanks @klihub .

askervin (approved):
LGTM.

PR #350, opened 2024-07-05
fixes: fix host-wait-vm-ssh-server, improve vm-reboot.

klihub:
This PR fixes or improves a few functions in the e2e test script library, including 
- fix host-wait-vm-ssh-server being hasty
- reboot and wait for VM by vagrant.

Here's a more detailed description and reasoning for these changes:

ssh -o ConnectTimeout=$timeout's semantics is such that if connection establishment fails early for other reasons, eg. the host IP not being reachable, ssh exits immediately. Rework the waiting function with this in mind. Instead of counting iterations with the incorrect assumption that each iteration would take 1 second, calculate and use a deadline. Also, move the function to lib/host.bash, since that's where we used to keep all host-* script lib functions.

Don't use "vm-command reboot" for rebooting the VM. It is a bit unreliable and gets stuck too often. Use vagrant halt+ vagrant up instead. Moreover, once rebooted, wait for the VM to come up and start serving SSH connections, like the function comment suggests it would already do. Finally, move the the function to vm.bash, where we used to keep most of the vm-* script lib functions.

### DIFF
--- a/test/e2e/lib/host.bash
+++ b/test/e2e/lib/host.bash
@@ -10,3 +10,33 @@ host-command() {
     command-end ${PIPESTATUS[0]}
     return $COMMAND_STATUS
 }
+
+host-wait-vm-ssh-server() {
+    local _vagrantdir="$1"
+    local _deadline=${deadline:-}
+    local _once=1
+
+    if [ -z "$_vagrantdir" ]; then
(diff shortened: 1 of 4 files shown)


askervin (approved):
LGTM

PR #351, opened 2024-07-08
fix: clean up plugin at the beginning/end of tests.

klihub:
This PR fixes failing nightly runs. They fail because the latest introduced case does not clean up the plugin and the first topology-aware test case does start by properly cleaning up potential leftover plugin instances from earlier cases. This PR should fix both.


fmuyassarov (approved):
LGTM

askervin (approved):
LGTM

PR #352, opened 2024-07-10
balloons: add support for explicit CPUs

fmuyassarov:
Introduce a new parameter (`preferCpus`) in Balloons policy for giving preferred CPU sets on each balloon type. 
When combining `preferIsolCpus` with `preferCpus`, priority will be given to isolated CPUs over the user-specified set of CPUs if these are two different sets.

### DIFF
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -56,6 +56,9 @@ const (
 	// virtDevIsolatedCpus is the name of a virtual device close to
 	// host isolated CPUs.
 	virtDevIsolatedCpus = "isolated CPUs"
+	// virtDevPreferredCpus represents a virtual device that is
+	// associated with a user-preferred set of CPUs
+	virtDevPreferredCpus = "user preferred CPUs"
 )
 
 // balloons contains configuration and runtime attributes of the balloons policy
@@ -560,8 +563,9 @@ func (p *balloons) newBalloon(blnDef *BalloonDef, confCpus bool) (*Balloon, erro
 		preferCloseToDevices:        blnDef.PreferCloseToDevices,
 		preferFarFromDevices:        blnDef.PreferFarFromDevices,
 		virtDevCpusets: map[string][]cpuset.CPUSet{
-			virtDevReservedCpus: {p.reserved},
-			virtDevIsolatedCpus: {p.options.System.Isolated()},
+			virtDevReservedCpus:  {p.reserved},
+			virtDevIsolatedCpus:  {p.options.System.Isolated()},
+			virtDevPreferredCpus: {cpuset.MustParse(blnDef.PreferCpus)},
 		},
 	}
 	if blnDef.AllocatorTopologyBalancing != nil {
@@ -1202,6 +1206,9 @@ func (p *balloons) fillCloseToDevices(blnDefs []*BalloonDef) {
 		if blnDef.PreferIsolCpus {
 			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevIsolatedCpus)
 		}
+		if cpuset.MustParse(blnDef.PreferCpus).Size() != 0 {
+			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevPreferredCpus)
+		}
 	}
 }
 
@@ -1221,6 +1228,9 @@ func (p *balloons) fillFarFromDevices(blnDefs []*BalloonDef) {
 		avoidDevs = append(avoidDevs, virtDevIsolatedCpus)
 	}
 	for _, blnDef := range blnDefs {
+		if cpuset.MustParse(blnDef.PreferCpus).Size() != 0 {
+			avoidDevs = append(avoidDevs, virtDevPreferredCpus)
+		}
 		for _, closeDev := range blnDef.PreferCloseToDevices {
 			if _, ok := devDefClose[closeDev]; !ok {
 				avoidDevs = append(avoidDevs, closeDev)
--- a/docs/resource-policy/policy/balloons.md
+++ b/docs/resource-policy/policy/balloons.md
@@ -154,6 +154,10 @@ Balloons policy parameters:
     separate `cpu.classes` objects, see below.
   - `preferCloseToDevices`: prefer creating new balloons close to
     listed devices. List of strings
+  - `preferCpus`: prefer allocating CPUs from an explicitly defined
+    set of CPUs. Warning: when combining `preferIsolCpus` with `preferCpus`,
+    priority will be given to isolated CPUs over the user-specified
+    set of CPUs.
   - `preferSpreadingPods`: if `true`, containers of the same pod
     should be spread to different balloons of this type. The default
     is `false`: prefer placing containers of the same pod to the same
@@ -168,7 +172,7 @@ Balloons policy parameters:
     preferring exclusive CPUs, as long as there are enough free
     CPUs. The default is `false`: prefer filling and inflating
     existing balloons over creating new ones.
-  - `preferIsolCpus`: if `true`,prefer system isolated CPUs (refer to
+  - `preferIsolCpus`: if `true`, prefer system isolated CPUs (refer to
     kernel command line parameter "isolcpus") for this balloon. Warning:
     if there are not enough isolated CPUs in the system for balloons that
     prefer them, balloons may include normal CPUs, too. This kind of
--- a/config/crd/bases/config.nri_balloonspolicies.yaml
+++ b/config/crd/bases/config.nri_balloonspolicies.yaml
@@ -195,6 +195,9 @@ spec:
                       items:
                         type: string
                       type: array
+                    preferCpus:
+                      description: 'preferCpus: user-preferred set of CPUs.'
+                      type: string
                     preferIsolCpus:
                       default: false
                       description: 'preferIsolCpus: prefer kernel isolated cpus'
--- a/deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
+++ b/deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
@@ -195,6 +195,9 @@ spec:
                       items:
                         type: string
                       type: array
+                    preferCpus:
+                      description: 'preferCpus: user-preferred set of CPUs.'
+                      type: string
                     preferIsolCpus:
                       default: false
                       description: 'preferIsolCpus: prefer kernel isolated cpus'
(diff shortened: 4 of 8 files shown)


klihub (commented):
Is this a good idea or something we both want and need ? So far, we have been deliberate and careful to not allow setting explicitly CPUs by ID for workloads.

fmuyassarov:
> Is this a good idea or something we both want and need ? So far, we have been deliberate and careful to not allow setting explicitly CPUs by ID for workloads.

I was expecting this question :smile:
To be honest I don't know weather it is good or bad idea to allow users for setting CPUs explicitly. I just had this task on my list. Perhaps @askervin has something to add.

klihub:
> > Is this a good idea or something we both want and need ? So far, we have been deliberate and careful to not allow setting explicitly CPUs by ID for workloads.
> 
> I was expecting this question :smile:
> To be honest I don't know weather it is good or bad idea to allow users for setting CPUs explicitly. I just had this task on my list. Perhaps @askervin has something to add. 

So @askervin knows about this and is okay with this ? I'll wait for his review then. I'd rather not open this Pandora's box if we can avoid it in any way. Usually the user does not really want any specific CPU or cpuset by ID(s). Instead they want their workload to be topologically close to some memory controller or one or more PCI devices and that's why they ask for the ability to set specific cpusets for workloads. However, almost always it is better to provide a way for the user/workload to describe *what* they need ("this workload needs to be close to a PCI device of a particular vendor and class" or "this workload needs to be close to that other workload") and then let the policy in use imolement the necessary means of *how* that is achieved. So, UI-/UX-visible interfaces (such as workload annotations or policy configuration) we prefer to be usually described in terms of *what* is needed not in terms of *how* that could be achieved on some piece of hardware, especually since the latter might not be portable between cluster nodes. Of course, there are exceptions to every rule...

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1202,6 +1206,9 @@ func (p *balloons) fillCloseToDevices(blnDefs []*BalloonDef) {
 		if blnDef.PreferIsolCpus {
 			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevIsolatedCpus)
 		}
+		if cpuset.MustParse(blnDef.PreferCpus).Size() != 0 {
```
askervin, line 1209: There is no error handling for invalid preferCpus, and the value is parsed many times. The opposite would be better.

Would you change this so that you parse each preferCpus only once and store the value to an internal attribute of balloon type? If you'd do this as part of configuration validation, then you could give a nice error message on parse errors.

docs/resource-policy/policy/balloons.md
```diff
@@ -154,6 +154,10 @@ Balloons policy parameters:
     separate `cpu.classes` objects, see below.
   - `preferCloseToDevices`: prefer creating new balloons close to
     listed devices. List of strings
+  - `preferCpus`: prefer allocating CPUs from an explicitly defined
+    set of CPUs. Warning: when combining `preferIsolCpus` with `preferCpus`,
```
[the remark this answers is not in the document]
askervin: @fmuyassarov, you can see excellent content for the warning in @klihub's comment: https://github.com/containers/nri-plugins/pull/352#issuecomment-2220222755

askervin:
> > > Is this a good idea or something we both want and need ? So far, we have been deliberate and careful to not allow setting explicitly CPUs by ID for workloads.

Based on the requests and wishes seen so far regarding CPU affinity, I'm leaning towards handing the gun to the user with all the big fat warnings on it (many picked from your inputs above). Use this only in case of emergency. You will lose all the nice features. The warranty is void if this seal is broken. Not for production use.

The reason is that I'm sure that also in the future there will be special cases where balloons simply is not smart enough out-of-the-box. There could be a special process or a feature or what not running/watching/tracing/enabled exactly on certain core, for instance. Or it might be that (once again) balloons would not be smart enough to fulfill a requirement as the precedence of its preferences is not configurable or something. And because there is no easy alternative for handling such special cases, it would be sad if affinity of single difficult workload would prevent using the balloons policy (and naturally the topology aware policy, too). For this kind of purposes I'd like to enable users to have this parameter in their sleave if everything else fails and there is no time to wait until balloons has proper support for their request.

But to be honest, I'm not leaning towards this direction without hesitation. I see it as a risk if anyone builds a production system that uses this option. I see it also as a risk that we might not get a feature request if someone can hack around their problem by using this feature. These are big risks. Maybe we could add to the list of warnings: "never use this parameter without telling the resource policy developers what they should implement to this policy so that you could stop using this parameter".

If you see better, I think we can postpone this feature and let @fmuyassarov work on P/E cores, which is definitely worth supporting. And adding that support immediately removes one reason to use preferCpus. :D

cmd/plugins/balloons/policy/balloons-policy.go
fmuyassarov: `MustParse` doesn't return an error so I guess moving the caller to validateConfig won't change anything in terms of error handling. But, I think we can use `cpuset.Parse()` similarly to [this](https://github.com/containers/nri-plugins/blob/46d22a67f989bc0fefe679e532461429a13936d1/pkg/utils/cpuset/cpuset.go#L35)?

fmuyassarov: and I'm okay to move the calling the MustParse/Parse to `validateConfig()` .

fmuyassarov: But probably we should all agree first if this PR is something that can be beneficial and merged. Because otherwise I will waste my time trying to fix something that we don't want to take in.

askervin: > `MustParse` doesn't return an error so I guess moving the caller to validateConfig won't change anything in terms of error handling. But, I think we can use `cpuset.Parse()` similarly to [this](https://github.com/containers/nri-plugins/blob/46d22a67f989bc0fefe679e532461429a13936d1/pkg/utils/cpuset/cpuset.go#L35)?

Yes, exactly. We definitely do not want to crash on a typo in a config.

> But probably we should all agree first if this PR is something that can be beneficial and merged. Because otherwise I will waste my & your time trying to fix something that we don't want to take in.

I agree. I think this PR is a very good start for this feature, thank you for writing it! Let's continue if @klihub and @kad show green(ish) light for this. I think you can put this PR into the draft state.

Would you proceed with preferCoreType (or preferPCores/preferECores... feel free to decide what would be the most suitable parameter name) instead? You should be able to test it in our e2e vm by adding environment variables to balloons policy container: look for OVERRIDE_SYS_CORE_CPUS and OVERRIDE_SYS_ATOM_CPUS from the code. Those variables enable faking that there are P and E cores in the system.

klihub:
> > > > Is this a good idea or something we both want and need ? So far, we have been deliberate and careful to not allow setting explicitly CPUs by ID for workloads.
> 
> Based on the requests and wishes seen so far regarding CPU affinity, I'm leaning towards handing the gun to the user with all the big fat warnings on it (many picked from your inputs above). Use this only in case of emergency. You will lose all the nice features. The warranty is void if this seal is broken. Not for production use.
> 
> The reason is that I'm sure that also in the future there will be special cases where balloons simply is not smart enough out-of-the-box. There could be a special process or a feature or what not running/watching/tracing/enabled exactly on certain core, for instance. Or it might be that (once again) balloons would not be smart enough to fulfill a requirement as the precedence of its preferences is not configurable or something. And because there is no easy alternative for handling such special cases, it would be sad if affinity of single difficult workload would prevent using the balloons policy (and naturally the topology aware policy, too). For this kind of purposes I'd like to enable users to have this parameter in their sleave if everything else fails and there is no time to wait until balloons has proper support for their request.
> 
> But to be honest, I'm not leaning towards this direction without hesitation. I see it as a risk if anyone builds a production system that uses this option. I see it also as a risk that we might not get a feature request if someone can hack around their problem by using this feature. These are big risks. Maybe we could add to the list of warnings: "never use this parameter without telling the resource policy developers what they should implement to this policy so that you could stop using this parameter".
> 
> If you see better, I think we can postpone this feature and let @fmuyassarov work on P/E cores, which is definitely worth supporting. And adding that support immediately removes one reason to use preferCpus. :D

@askervin If you tell me that we need this, and that we can paint it with big enough "dontcha shoot yourself in the foot" warnings, then I trust your judgement and am fine with it, too.

kad:
Let's "sleep over" with that configuration option. I agree that giving exact cpuset to the user might be usable in some rare scenarios, as well, it would lead to many of corner cases: e.g. multiple balloons that have overlapping cpusets, or in some cpus being allocated to other ballons in normal policy way if this balloon can be shrink to 0, etc,etc...
What we can use as a potential "middle ground" is the option for balloon that would specify multiple cpusets in priority order, so normal cpu allocator would be using those more as hint rather than explicit list... but again, that might cause many other corner cases. So, let's get back to "drawing board" with this for near future?

klihub:
Let's mark this as a draft, while we sleep/sit on it.

NoamNakash:
> But to be honest, I'm not leaning towards this direction without hesitation. I see it as a risk if anyone builds a production system that uses this option. I see it also as a risk that we might not get a feature request if someone can hack around their problem by using this feature. These are big risks. Maybe we could add to the list of warnings: "never use this parameter without telling the resource policy developers what they should implement to this policy so that you could stop using this parameter".

Hi, I have a usecase, where we limit certain non-k8s pod workloads on our clusters (systemd services like kube-api, kube-scheduler and other "infra" services) to only use a certain CPU set. the goal is to have "infra" and "workload" completely separate using balloons, and I don't see a way to create that separation without this feature. Would this be something balloon policy is able to do? or do you have an idea on a feature that is less dangerous that would allow us to do this (I don't mind coding it, but I don't have any other idea on how to solve it)?

optionally, would an "excludeCPUs" option to list CPUs to exclude from balloon scheduling for host workloads that are not pods

tl;dr - we have cpu cores where we don't want workloads running (generally would prefer allowing certain workloads to run and disallowing others)

P.S
this is my first time commenting, please tell me if this comment doesn't fit here and where I should be asking this

fmuyassarov:
@NoamNakash 
Sorry for the radio silence, most of the people were and are still on vacation and thus there was no response. 
Thank you for sharing your use case. I'm still leaning towards having this option available but I can't decide on my own since this is not my project and will wait for others to come back from the vacation and decide what do we do wit this feature and potential user requests like yours.

askervin:
@noamnakash, sorry about the delay. After quite a few discussions we ended up with a decision not to implement this feature, at least not at this point. I will elaborate this in a separate comment.

For your particular use case, that is separating non-k8s processes on dedicated CPUs, there is an option called `availableResources`. It definitely was not easy to find, as it was not documented among other balloons options, so I added it in PR#355. This option is not a perfect match, as it lists CPUs available for the policy instead of CPUs not available to the policy.

Do you still think it works for you?

PR #354, opened 2024-08-13
balloons: add support for power efficient & high performance cores

fmuyassarov:
Introduce a new parameter `coreType` in Balloons policy for allowing users to express their preference of selecting P-high performance or power efficient cores.

### DIFF
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -39,7 +39,6 @@ const (
 	PolicyName = "balloons"
 	// PolicyDescription is a short description of this policy.
 	PolicyDescription = "Flexible pools with per-pool CPU parameters"
-
 	// balloonKey is a pod annotation key, the value is a pod balloon name.
 	balloonKey = "balloon." + PolicyName + "." + kubernetes.ResmgrKeyNamespace
 	// hideHyperthreadsKey is a pod annotation key for pod/container-specific hyperthread allowance.
@@ -56,6 +55,12 @@ const (
 	// virtDevIsolatedCpus is the name of a virtual device close to
 	// host isolated CPUs.
 	virtDevIsolatedCpus = "isolated CPUs"
+	// virtDevECores is the name of a virtual device close to
+	// power efficient cores.
+	virtDevECores = "efficient cores"
+	// virtDevPCores is the name of a virtual device close to
+	// high performance cores.
+	virtDevPCores = "performance cores"
 )
 
 // balloons contains configuration and runtime attributes of the balloons policy
@@ -562,6 +567,8 @@ func (p *balloons) newBalloon(blnDef *BalloonDef, confCpus bool) (*Balloon, erro
 		virtDevCpusets: map[string][]cpuset.CPUSet{
 			virtDevReservedCpus: {p.reserved},
 			virtDevIsolatedCpus: {p.options.System.Isolated()},
+			virtDevECores:       {p.cpuAllocator.GetCPUPriorities()[cpuallocator.PriorityLow]},
+			virtDevPCores:       {p.cpuAllocator.GetCPUPriorities()[cpuallocator.PriorityHigh]},
 		},
 	}
 	if blnDef.AllocatorTopologyBalancing != nil {
@@ -1201,6 +1208,12 @@ func (p *balloons) fillCloseToDevices(blnDefs []*BalloonDef) {
 	for _, blnDef := range blnDefs {
 		if blnDef.PreferIsolCpus {
 			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevIsolatedCpus)
+		}
+		if blnDef.PreferCoreType == "performance" {
+			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevPCores)
+		}
+		if blnDef.PreferCoreType == "efficient" {
+			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevECores)
 		}
 	}
 }
--- a/config/crd/bases/config.nri_balloonspolicies.yaml
+++ b/config/crd/bases/config.nri_balloonspolicies.yaml
@@ -195,6 +195,14 @@ spec:
                       items:
                         type: string
                       type: array
+                    preferCoreType:
+                      description: |-
+                        preferCoreType: prefer performance or efficient (P/E) CPU cores on
+                        hybrid architectures.
+                      enum:
+                      - efficient
+                      - performance
+                      type: string
                     preferIsolCpus:
                       default: false
                       description: 'preferIsolCpus: prefer kernel isolated cpus'
--- a/docs/resource-policy/policy/balloons.md
+++ b/docs/resource-policy/policy/balloons.md
@@ -154,6 +154,9 @@ Balloons policy parameters:
     separate `cpu.classes` objects, see below.
   - `preferCloseToDevices`: prefer creating new balloons close to
     listed devices. List of strings
+  - `preferCoreType`:  specifies preferences of the core type which
+    could be either power efficient (`efficient`) or high performance
+    (`performance`).
   - `preferSpreadingPods`: if `true`, containers of the same pod
     should be spread to different balloons of this type. The default
     is `false`: prefer placing containers of the same pod to the same
@@ -168,7 +171,7 @@ Balloons policy parameters:
     preferring exclusive CPUs, as long as there are enough free
     CPUs. The default is `false`: prefer filling and inflating
     existing balloons over creating new ones.
-  - `preferIsolCpus`: if `true`,prefer system isolated CPUs (refer to
+  - `preferIsolCpus`: if `true`, prefer system isolated CPUs (refer to
     kernel command line parameter "isolcpus") for this balloon. Warning:
     if there are not enough isolated CPUs in the system for balloons that
     prefer them, balloons may include normal CPUs, too. This kind of
--- a/pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
+++ b/pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
@@ -219,6 +219,11 @@ type BalloonDef struct {
 	// preferIsolCpus: prefer kernel isolated cpus
 	// +kubebuilder:default:=false
 	PreferIsolCpus bool `json:"preferIsolCpus,omitempty"`
+	// preferCoreType: prefer performance or efficient (P/E) CPU cores on
+	// hybrid architectures.
+	// +optional
+	// +kubebuilder:validation:Enum:=efficient;performance
+	PreferCoreType string `json:"preferCoreType,omitempty"`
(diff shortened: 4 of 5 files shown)


config/crd/bases/config.nri_balloonspolicies.yaml
```diff
@@ -195,6 +195,9 @@ spec:
                       items:
                         type: string
                       type: array
+                    preferECores:
+                      description: 'preferECores: prefer efficient cores (E-cores)'
+                      type: boolean
```
klihub, line 200: Couldn't handle this at the user-visible interfaces without explicit reference to P- and E-cores which IMO is a bit HW-/platform-specific. One potential way to do it would be to (re)use instead the CPU priorities `high` and `low`, which are already end-user visible through configuration (used by the existing `AllocatorPriority` option). BTW, this is how a similar preference configuration and corresponding annotations have been added to the topology-aware policy. `high` will prefer P-cores if available, and (together with `normal`) avoid E-cores if possible, while `low` will prefer E-cores if available to fulfill the allocation. This would also has the additional benefit that if you are not running on a hybrid-core system but instead on a system where there is a single kind of core but different cores are clocked differently, allocation will still try to do a reasonable job to a similar effect within the realm of possibilities and considering the different clocking (classes) of the cores.

fmuyassarov: Do you suggest to use AllocatorPriority with values Low, High ?

klihub: Well, I'm really asking whether we could somehow avoid leaking arguably hardware-specific terminology into the user-visible bits of the interface to the balloons policy, in this case through the configuration ?

So not necessarily directly AllocatorPriority with values low/high, but at least effectively the same approach codewise and similar terminology in the user-visible configuration. But I think using AllocatorPriority with low/high should have the same preference effect as the separate preferences you try to introduce here...

fmuyassarov: I'm fine to use `AllocatorPriority` and perhaps some alternatives I can think of would be `PerformanceMode`, `ExecutionProfile`, `ResourcePriority`, `AllocationWeight`...

pfl: > similar preference configuration and corresponding annotations have been added to the topology-aware policy

Consistent terminology between CPU selecting functionality would be a big plus, even/especially between different plugins. It avoids human error with less to remember when configuring.

fmuyassarov: @klihub perhaps we use `CoreOptimization` or `CPUUtilizationType` or `CoreUsageMode`.

fmuyassarov: One think to keep in mind though, when having a single key instead of two separate as it is currently on this PR, there won't be an option to opt in for both efficient and performance cores nor will there be a automatic prioritization of performance cores over efficient cores as it is the case right now.

fmuyassarov:
@klihub @askervin @kad PTAL. Updated according to the discussion we had yesterday.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -35,11 +35,12 @@ import (
 )
 
 const (
+	highPrio = cpuallocator.PriorityHigh
+	lowPrio  = cpuallocator.PriorityLow
```
klihub, line 39: It looks to me as the original const block was deliberately 'exported consts first, linebreak, unexported consts'. @askervin Was it so, or did it just happen to look like that ? If it was so, I'd propose to reorder these additions to keep the original idea intact, IOW move these to the original 2nd block and reinsert the removed empty line.

docs/resource-policy/policy/balloons.md
```diff
@@ -154,6 +154,9 @@ Balloons policy parameters:
     separate `cpu.classes` objects, see below.
   - `preferCloseToDevices`: prefer creating new balloons close to
     listed devices. List of strings
+  - `coreType`:  specifies preferences of the core type which could
+    be power either efficient (`efficient`) or high performance
```
klihub, line 158: typo/thinko: 'could be either power efficient ('`efficient`) or high performance'

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -219,6 +219,10 @@ type BalloonDef struct {
 	// preferIsolCpus: prefer kernel isolated cpus
 	// +kubebuilder:default:=false
 	PreferIsolCpus bool `json:"preferIsolCpus,omitempty"`
+	// coreType: Accepts either performance or efficienct
```
klihub, line 222: typo: 'efficienct' should be 'efficient'

klihub (changes requested):
I have one nit and spotted a few typos. Otherwise LGTM.

config/crd/bases/config.nri_balloonspolicies.yaml
klihub: But since it is now a string with no (other) default (than the zero-value empty string), you can implement prioritizing performance cores by default, if you think that is the right thing to do. 

About preferring both... that's IMO a bit vague unless you ask for exactly 2 CPUs. Otherwise if you ask for, let's say, 10 CPUs and prefer both what does that mean in practice: 1 power-efficient and 9 performance cores, 9 power-efficient and 1 performance core, any other of the possible combinations ?

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1202,6 +1209,12 @@ func (p *balloons) fillCloseToDevices(blnDefs []*BalloonDef) {
 		if blnDef.PreferIsolCpus {
 			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevIsolatedCpus)
 		}
+		if blnDef.CoreType == "performance" {
+			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevPCores)
+		}
+		if blnDef.CoreType == "efficient" {
```
klihub, line 1215: ubernit: maybe it would have been warranted to use consts for "performance" and "efficient", since they appear to be referenced/used multiple times in the code.

klihub (approved):

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1202,6 +1209,12 @@ func (p *balloons) fillCloseToDevices(blnDefs []*BalloonDef) {
 		if blnDef.PreferIsolCpus {
 			blnDef.PreferCloseToDevices = append(blnDef.PreferCloseToDevices, virtDevIsolatedCpus)
 		}
+		if blnDef.CoreType == "performance" {
```
pfl, line 1212: It is fine to append both virtDevIsolatedCpus and either of virtDevPCores or virtDevECores to the blnDef.PreferCloseToDevices list?

pfl (approved):

cmd/plugins/balloons/policy/balloons-policy.go
fmuyassarov: yes, because preference is given to isolated CPUs.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1221,6 +1234,12 @@ func (p *balloons) fillFarFromDevices(blnDefs []*BalloonDef) {
 		avoidDevs = append(avoidDevs, virtDevIsolatedCpus)
 	}
 	for _, blnDef := range blnDefs {
+		if blnDef.CoreType == "efficient" {
```
askervin, line 1237: fillCloseToDevices() has already appended virtDevECores and virtDevPCores into blndef.PreferCloseToDevices. The next loop after these `if`s goes through blndef.PreferCloseToDevices and adds all devices in there into the avoidDevs list (but at most once).

This change adds virtDevECores and virtDevPCores potentially many times into the avoidDevs list, and they will be added once more by the next loop.

This sounds like unwanted behavior. Did you have some specific case (a preference order?) in mind when adding these separately, or would it suffice to let the existing loop add them. In the latter case this change can be dropped.

docs/resource-policy/policy/balloons.md
```diff
@@ -154,6 +154,9 @@ Balloons policy parameters:
     separate `cpu.classes` objects, see below.
   - `preferCloseToDevices`: prefer creating new balloons close to
     listed devices. List of strings
+  - `coreType`:  specifies preferences of the core type which could
```
askervin, line 157: As this option is handled as a preference, that is, inability to honor the configuration choice is not an error, would you name it `preferCoreType`?

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -219,6 +219,10 @@ type BalloonDef struct {
 	// preferIsolCpus: prefer kernel isolated cpus
 	// +kubebuilder:default:=false
 	PreferIsolCpus bool `json:"preferIsolCpus,omitempty"`
+	// coreType: Accepts either performance or efficient
```
askervin, line 222: The documentation should state what it does, not just valid values. 

```
preferCoreType: prefer performance or efficient (P/E) CPU cores on hybrid architectures.
```

cmd/plugins/balloons/policy/balloons-policy.go
fmuyassarov: Thanks, fixed.

cmd/plugins/balloons/policy/balloons-policy.go
fmuyassarov: we have only one instance of "efficient" and "performance". So, I don't think we need a const here.

fmuyassarov:
@klihub @askervin PTAL. Addressed the last comments.

klihub:
> @klihub @askervin PTAL. Addressed the last comments.

@fmuyassarov Just to make sure I understand the latest changes: we have explicit positive preferences but no implied negative ones, right ? IOW: a core type preference is simply translated to a virtual device close to the right types of cores, but we don't try to generate/inject avoided virtual devices to avoid taking cores of non-preferred types. Is this correct ?

fmuyassarov:
> > @klihub @askervin PTAL. Addressed the last comments.
> 
> @fmuyassarov Just to make sure I understand the latest changes: we have explicit positive preferences but no implied negative ones, right ? 

That's right. Basically, try to use cores  of the preferred device as much as we can if not we continue to use non-preferred ones.

klihub:
> > > @klihub @askervin PTAL. Addressed the last comments.
> > 
> > 
> > @fmuyassarov Just to make sure I understand the latest changes: we have explicit positive preferences but no implied negative ones, right ?
> 
> That's right. Basically, try to use cores of the preferred device as much as we can if not we continue to use non-preferred ones.

Thanks ! I am happy with this. Let's wait for @askervin to check if his concerns have now all been addressed.

askervin (approved):
LGTM.

I think we can have the e2e test added in a separate PR.

klihub (approved):

PR #357, opened 2024-09-02
build: allow building a single plugin image.

klihub:
Allow skipping builds of operator and bundle image. This allows one to build just a single image, instead lf always building at least three.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,10 @@ PLUGINS ?= \
 BINARIES ?= \
 	config-manager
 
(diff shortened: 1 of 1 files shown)


askervin (approved):
LGTM! I was just about to write something very similar. But now I only cherry-picked this up and ran happily:
```
make PLUGINS=nri-resource-policy-topology-aware OTHER_IMAGE_TARGETS="" BINARIES="" DEBUG=1 images
```

Thank you for this developer experience improvement!

PR #358, opened 2024-09-17
resmgr: lifecycle overlap detection and workaround.

klihub:
We recently discovered a problem with the generated stream of container lifecycle events with some runtime versions. A side effect of this is that we get Create/Stop events for multiple container instances with seemingly overlapping lifecycle: the latter instance get created before the former one is stopped.

When undetected, such a false overlap might cause overcommit of resources, with both instances temporarily using the full resource set of the offending container. As a workaround, we now track containers also by fully qualified name ($namespace/pod/ctr) and internally generate an event for releasing the resources of the old instance whenever we notice that a creation event would cause a duplicate instance for the same name.

### DIFF
--- a/cmd/plugins/topology-aware/policy/topology-aware-policy.go
+++ b/cmd/plugins/topology-aware/policy/topology-aware-policy.go
@@ -134,6 +134,7 @@ func (p *policy) Start() error {
 	p.checkColdstartOff()
 
 	p.root.Dump("<post-start>")
+	p.checkAllocations("  <post-start>")
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


askervin (approved):
LGTM

cmd/plugins/topology-aware/policy/topology-aware-policy.go
```diff
@@ -148,19 +149,82 @@ func (p *policy) Sync(add []cache.Container, del []cache.Container) error {
...
+	}
+
+	for key, grants := range dup {
+		log.Error("%s DUPLICATE allocation entries for container %s", prefix, key)
+		for _, g := range grants {
+			log.Error("%s   %s (%s)", prefix, g, g.GetContainer().GetID())
+		}
+	}
+
+	log.Info("%s total CPU granted: %dm (%d exclusive + %dm shared), total memory granted: %s",
```
kad, line 212: given that chackAllocations() called quite often, would it be this log.Info() making it too chatty?

pfl (approved):
With the comments already on checkAllocations()

cmd/plugins/topology-aware/policy/topology-aware-policy.go
klihub: With policy-debugging off, there would be one such dump
- after startup
- after initial sync
- after reconfiguration
- after allocation requests
- after a release requests
- after resource update requests
- and one after an internal call to reallocateResources()

This last one is called when the configuration changes, so that clearly might be one too much. I can remove that. The others normally change the overall state of allocations, so it might be a good idea to calculate and dump the final state after those.

klihub: Ok, I updated it so that for `reallocateResources()` we neither dump the pools nor the total allocations by grants, since we later do both as a post-config action.

kad (approved):

askervin (approved):
LGTM

PR #361, opened 2024-09-19
e2e: bridge CNI setup fixes for Fedora/containerd. 

klihub:
Fix e2e tests runnning with bridge CNI plugin on fedora. In particular,

- fix non-default CNI plugin binary path on fedora
- configure bridge CNI plugin earlier during provisoning

These fixes should ensure that the node does not stay lingering in NotReady state, causing false test failures/positives.

### DIFF
--- a/test/e2e/playbook/provision.yaml
+++ b/test/e2e/playbook/provision.yaml
@@ -306,6 +306,23 @@
             regexp: ' *bin_dir *= *"/opt/cni/bin" *'
             line: 'bin_dir = "/usr/libexec/cni"'
 
+    - name: Configure bridge CNI plugin
+      when: cni_plugin == "bridge"
+      block:
+        - name: Copy CNI bridge plugin configuration
+          ansible.builtin.copy:
+            src: "{{ nri_resource_policy_src }}/test/e2e/files/10-bridge.conf.in"
+            dest: "/etc/cni/net.d/10-bridge.conf"
+            owner: root
+            group: root
+            mode: '0644'
+
+        - name: Update CNI bridge plugin configuration
+          ansible.builtin.replace:
+            path: /etc/cni/net.d/10-bridge.conf
+            regexp: '(CNI_NETWORK)'
+            replace: "{{ network }}"
+
     - name: Setup NRI
       ansible.builtin.file:
         path: "{{ item }}"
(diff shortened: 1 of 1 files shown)


askervin (commented):
I had an issue with a regexp with '(#CNI_NETWORK#)', but it worked fine with only 'CNI_NETWORK'. Would the latter work for you?

klihub:
> I had an issue with a regexp with '(#CNI_NETWORK#)', but it worked fine with only 'CNI_NETWORK'. Would the latter work for you?

Thanks for spotting that ! My tree was not up-to-date, missing your regexp pattern fix commit, plus instead of moving that block earlier it was accidentally copying it earlier. It would have not worked at all.

klihub:
> Let me put up an e2e test run to double-check that this committed and pushed version works...

Okay, it looks good. Spun up the VM and the cluster and successfully ran all tests.

test/e2e/playbook/provision.yaml
```diff
@@ -290,12 +290,37 @@
...
+            - mkdir -p /etc/containerd
+            - containerd config default > /etc/containerd/config.toml
+            - sed -i 's/^.*disabled_plugins *= *.*$/disabled_plugins = []/' /etc/containerd/config.toml
+            - sed -i 's/SystemdCgroup = false/SystemdCgroup = true/g' /etc/containerd/config.toml
+
+        - name: Update CNI plugin directory on Fedora
+          when: ansible_facts['distribution'] == "Fedora"
+          ansible.builtin.shell: "{{ item }}"
+          with_items:
+            - sed -i 's:/opt/cni/bin:/usr/libexec/cni:g' /etc/containerd/config.toml
```
kad, line 306: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html might be more efficient (faster than shell + sed)

kad (approved):

pfl (approved):

test/e2e/playbook/provision.yaml
klihub: Updated to use `ansible.builtin.lineinfile`. Tested with `cni_plugin=bridge` and it still worked.

askervin (approved):
LGTM

PR #362, opened 2024-09-19
e2e: use bridge CNI plugin by default.

klihub:
*Note: This PR is stacked and depends on #361.*

The latest cilium release is [broken](https://github.com/cilium/cilium-cli/issues/2795). Let's switch to use the bridge plugin by default.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -19,7 +19,7 @@ export COMMAND_OUTPUT_DIR="$TEST_OUTPUT_DIR"/commands
 
 distro=${distro:-$DEFAULT_DISTRO}
 export k8scri=${k8scri:-"containerd"}
-export cni_plugin=${cni_plugin:-cilium}
+export cni_plugin=${cni_plugin:-bridge}
 export cni_release=${cni_release:-latest}
 TOPOLOGY_DIR=${TOPOLOGY_DIR:=e2e}
 


test/e2e/playbook/provision.yaml
```diff
@@ -290,12 +290,37 @@
...
+            - mkdir -p /etc/containerd
+            - containerd config default > /etc/containerd/config.toml
+            - sed -i 's/^.*disabled_plugins *= *.*$/disabled_plugins = []/' /etc/containerd/config.toml
+            - sed -i 's/SystemdCgroup = false/SystemdCgroup = true/g' /etc/containerd/config.toml
+
+        - name: Update CNI plugin directory on Fedora
+          when: ansible_facts['distribution'] == "Fedora"
+          ansible.builtin.shell: "{{ item }}"
+          with_items:
+            - sed -i 's:/opt/cni/bin:/usr/libexec/cni:g' /etc/containerd/config.toml
```
kad, line 306: I'm not sure that sed is the best way in ansible to replace one string to another (it will be executed every time, instead of detecting that file change is already done and no more action needed). but ok for now.

kad (approved):

test/e2e/playbook/provision.yaml
klihub: Updated to use `ansible.builtin.lineinfile`.

askervin (approved):
LGTM

askervin (approved):
LGTM

PR #356, opened 2024-08-27
.codespell*,.github,*: add codespell configuration, workflow, fix codespell errors.

klihub:
This PR 
- adds codespell configuration
- adds a codespell workflow
- fixes existing spelling errors detected by codespell

This PR will conflict with #332 if/when that gets merged. I'd prefer us to focus our limited review capacity on #332, Also, I expect reviewing this to mostly be a one-time skim-through. Therefore, I'll keep this a draft until #332 gets merged.

PR #368, opened 2024-09-30
balloons: add "preserve" option to match containers whose pinning must not be modified

askervin:
Some containers must be able to run with their resources unmanaged by a resource policy. Currently this can be achieved only by annotating pods. However, in certain environments it is more convenient to change a node-specific policy configuration rather than change annotations of workloads.

This PR adds such an option to the balloons policy. Once the option takes its final shape, we can add similar option to the topology-aware policy, too.


klihub (commented):
LGTM, but I wouldn't call this ignored but preserved, since it basically has the same effect as annotating a container preserved.

config/crd/bases/config.nri_balloonspolicies.yaml
```diff
@@ -295,6 +295,54 @@ spec:
...
+                description: |-
+                  Ignore specifies containers that should be completely
+                  ignored by the policy.
+                properties:
+                  matchExpressions:
+                    description: MatchExpressions specifies one or more expressions.
+                    items:
+                      description: |-
+                        Expression describes some runtime-evaluated condition. An expression
+                        consist of a key, an operator and a set of values. An expressions is
```
fmuyassarov, line 308:
```suggestion
                        consist of a key, an operator and a set of values. An expression is
```

deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
```diff
@@ -295,6 +295,54 @@ spec:
...
+                description: |-
+                  Ignore specifies containers that should be completely
+                  ignored by the policy.
+                properties:
+                  matchExpressions:
+                    description: MatchExpressions specifies one or more expressions.
+                    items:
+                      description: |-
+                        Expression describes some runtime-evaluated condition. An expression
+                        consist of a key, an operator and a set of values. An expressions is
```
fmuyassarov, line 308:
```suggestion
                        consist of a key, an operator and a set of values. An expression is
```

deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
```diff
@@ -295,6 +295,54 @@ spec:
...
+                        evaluated against an object which implements the Evaluable interface.
+                        Evaluating an expression consists of looking up the value for the key
+                        in the object, then using the operator to check it agains the values
```
fmuyassarov, line 311:
```suggestion
                        in the object, then using the operator to check it against the values
```

fmuyassarov: Another nit, I suspect this description was added manually here since it doesn't exist in the field description in pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go file? Should we perhaps add this description there and let the code generator to take care of adding the same description to the CRD here? Feel free to not do it since this is not a showstopper.

klihub: Definitely. The resource management plugin CRDs should never be manually edited. Instead the corresponding golang types should be updated (and annotated with the necessary comments) then CRD (among the whole shebang) regenerated by `make gneerate`).

klihub: No, I think @askervin did it correctly here, in this PR...

That description properly propagates from `pkg/apis/resmgr/v1alpha1/types.go` where `Expression`, which is only referenced here, is originally defined.

klihub: Also, if I run `make generate && git status --short` everything looks clean.

askervin:
Renamed to `preserve`. Now it's aligned with the preserve annotation.

I'd prefer fixing typos in resmgr.Expression documentation (in pkg/apis/resmgr/v1alpha1/types.go) in a separate PR. That's because including those fixes into this patch series would only make it more difficult to understand, just by checking modified files in the PR, that which exact source files needed a change in order to add a new option+implementation+test+documentation. Sort of minimizing the blast radius. What do you think, @fmuyassarov?

klihub (approved):

klihub:
> Renamed to `preserve`. Now it's aligned with the preserve annotation.
> 
> Thanks for spotting the typos! I'd prefer fixing typos in resmgr.Expression documentation (in pkg/apis/resmgr/v1alpha1/types.go) in a separate PR. That's because including those fixes into this patch series would only make it more difficult to understand, just by checking modified files in the PR, that which exact source files needed a change in order to add a new option+implementation+test+documentation. Sort of minimizing the blast radius. What do you think, @fmuyassarov?
> 
> Because the fix modifies generated code files in this and another option, filing that PR would cause conflicts before this PR is merged. So I could file it after merging.

And I hope we could get #356 merged soon (once libmem is in) and that will bring in a bunch or spelling fixes and make it easier to avoid them in the future.

deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
fmuyassarov: Ah true, [Expression](https://github.com/containers/nri-plugins/blob/5d4ed7c4defcbc30a1c10eb0d91887ca055a1419/pkg/apis/resmgr/v1alpha1/types.go#L17) definition is already there on the main. I thought it is being added here instead of referencing to the something that was already there. Then all good here.

fmuyassarov (approved):
LGTM

PR #372, opened 2024-10-02
codespell: ignore more files.

klihub:
Ignore the build directory, test directories (at the top level), PNG files and tarballs.

codespell just keeps on giving...


askervin (approved):
LGTM

PR #373, opened 2024-10-03
e2e: skip long running tests by default.

klihub:
Skip tests matching \*long\*, \*fuzz\*, or \*stress\* by default. Run them when `skip_long_tests=no` is set.

### DIFF
--- a/test/e2e/run_tests.sh
+++ b/test/e2e/run_tests.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 TESTS_DIR="$1"
+SKIP_LONG_TESTS="${skip_long_tests:-yes}"
 RUN_SH="${0%/*}/run.sh"
 
 DEFAULT_DISTRO=${DEFAULT_DISTRO:-"generic/fedora37"}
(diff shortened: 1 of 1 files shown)


klihub:
@askervin @fmuyassarov @kad @pfl PTAL.

test/e2e/run_tests.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 TESTS_DIR="$1"
+TESTS_RUN_LONG="${long_tests:-no}"
```
pfl, line 4: Just checking: the intention of long_tests is for it to be an environment variable or specified on the command line before the command, right?

pfl (approved):

askervin (approved):
LGTM.

The only improvement that I could come up with is printing something like "SKIP $TEST_DIR (long_tests=no)", as I can imagine myself wondering after few months why ./run_tests.sh <path-to-a-fuzz-test> doesn't do anything.

But this is a comment rather than a request because "Running tests matching:..." already should reveal what is going on. And I can imagine also finding the explanation soon enough from there. :D

test/e2e/run_tests.sh
askervin: In the perspective of the script, there is no difference between an exported environment variable and a variable that is specified on the command line before the command. Both will work. (In case of name collision, the definition on the command line takes precedence.)

klihub:
> LGTM.
> 
> The only improvement that I could come up with is printing something like "SKIP $TEST_DIR (long_tests=no)", as I can imagine myself wondering after few months why ./run_tests.sh doesn't do anything.
>

Good point. I'm think I myself would not realize after a week why long tests are not run at all. Updated, adding printouts when we skip tests. I also inverted the control variable name and semantics to `skip_long_tests`, which feels more correct.

askervin (approved):
LGTM

pfl (approved):

PR #377, opened 2024-10-04
e2e: fix provisioning for Ubuntu cloud image.

klihub:
Ubuntu cloud image does not need a DNS fixup.


kad (approved):
I'm ok with it, but probably some day we would need to fix those sed with internal ansible line replacement, it would be faster in overall script runtime.

fmuyassarov:
> I'm ok with it, but probably some day we would need to fix those sed with internal ansible line replacement, it would be faster in overall script runtime.

+1

fmuyassarov (approved):

PR #378, opened 2024-10-04
e2e: adjust $VM_HOSTNAME for policy node config usage.

klihub:
Take into account that $VM_HOSTNAME's like xyzzy-foobar-22.04 will propagate to the VM distro hostname as xyzzy-foobar-22.

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -481,26 +481,31 @@ get-config-generation() {
     vm-command-q "kubectl get -n kube-system $resource -ojsonpath={.metadata.generation}"
 }
 
+get-hostname-for-vm() {
+    local node="${node:-$VM_HOSTNAME}"
+    echo ${node%.*}
+}
+
 get-config-node-status-generation() {
-    local resource="$1" node="${node:-$VM_HOSTNAME}"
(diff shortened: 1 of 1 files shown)


kad (approved):
LGTM

fmuyassarov (approved):

test/e2e/run.sh
```diff
@@ -481,26 +481,32 @@ get-config-generation() {
...
 
+get-hostname-for-vm() {
+    local node="${node:-$VM_HOSTNAME}"
+    echo ${node%.*}
+}
+
 get-config-node-status-generation() {
-    local resource="$1" node="${node:-$VM_HOSTNAME}"
+    local resource="$1" node="$(get-hostname-for-vm)"
+    node="${node%.*}"
```
askervin, line 491: 491 looks suspicious, as we're cutting again the suffix after the last dot, but we don't do that after line 497 yet both $node act as keys to the same map/dict/object.

askervin (changes requested):
Marking to "Request changes" to prevent merging by accident.

askervin (approved):
LGTM

PR #379, opened 2024-10-10
balloons: set frequency scaling governor only when requested

fmuyassarov:
Avoid enforcing the frequency scaling governor if the user hasn't explicitly requested it. Previously, we attempted to set it regardless, leading to unnecessary error logs. Furthermore, fix formatting issue when logging error case.

### DIFF
--- a/pkg/resmgr/control/cpu/cpu.go
+++ b/pkg/resmgr/control/cpu/cpu.go
@@ -16,7 +16,6 @@ package cpu
 
 import (
 	"fmt"
-	"strconv"
 
 	"github.com/containers/nri-plugins/pkg/utils/cpuset"
 
@@ -123,30 +122,15 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
 	return nil
 }
 
-// enforceCpufreqGovernor enforces a class-specific cpufreq governor to a cpuset.
-func (ctl *cpuctl) enforceCpufreqGovernor(class string, cpusIds ...int) error {
-	if _, ok := ctl.classes[class]; !ok {
-		return fmt.Errorf("non-existent cpu class %q", class)
-	}
-	governor := ctl.classes[class].FreqGovernor
-	for cpu := range cpusIds {
-		log.Info(strconv.Itoa(cpu), governor)
-	}
-	log.Debug("enforcing cpu frequency governor %q on %v", governor, cpusIds)
-	if err := utils.SetScalingGovernorForCPUs(cpusIds, governor); err != nil {
-		return fmt.Errorf("Cannot set cpufreq governor %d: %w", governor, err)
-	}
-	return nil
-}
-
 // enforceCpufreq enforces a class-specific cpufreq configuration to a cpuset
 func (ctl *cpuctl) enforceCpufreq(class string, cpus ...int) error {
-	if _, ok := ctl.classes[class]; !ok {
+	c, ok := ctl.classes[class]
+	if !ok {
 		return fmt.Errorf("non-existent cpu class %q", class)
 	}
 
-	min := int(ctl.classes[class].MinFreq)
-	max := int(ctl.classes[class].MaxFreq)
+	min := int(c.MinFreq)
+	max := int(c.MaxFreq)
 	log.Debug("enforcing cpu frequency limits {%d, %d} from class %q on %v", min, max, class, cpus)
 
 	if err := utils.SetCPUsScalingMinFreq(cpus, min); err != nil {
@@ -155,6 +139,14 @@ func (ctl *cpuctl) enforceCpufreq(class string, cpus ...int) error {
 
 	if err := utils.SetCPUsScalingMaxFreq(cpus, max); err != nil {
 		return fmt.Errorf("Cannot set max freq %d: %w", max, err)
+	}
+
+	if governor := c.FreqGovernor; governor != "" {
+		log.Debug("enforcing cpu frequency governor %q from class %q on %v", governor, class, cpus)
+
+		if err := utils.SetScalingGovernorForCPUs(cpus, governor); err != nil {
+			return fmt.Errorf("cannot set cpufreq governor %q: %w", governor, err)
+		}
 	}
 
 	return nil
@@ -274,9 +266,6 @@ func (ctl *cpuctl) configure(cfg *cfgapi.Config) error {
 		if _, ok := ctl.classes[class]; ok {
 			// Re-configure cpus (sysfs) according to new class parameters
 			if err := ctl.enforceCpufreq(class, cpus.SortedMembers()...); err != nil {
-				log.Error("cpufreq enforcement on re-configure failed: %v", err)
-			}
-			if err := ctl.enforceCpufreqGovernor(class, cpus.SortedMembers()...); err != nil {
 				log.Error("cpufreq enforcement on re-configure failed: %v", err)
 			}
 		} else {
--- a/pkg/resmgr/control/cpu/api.go
+++ b/pkg/resmgr/control/cpu/api.go
@@ -61,9 +61,6 @@ func Assign(c cache.Cache, class string, cpus ...int) error {
 		if err := ctl.enforceCpufreq(class, cpus...); err != nil {
 			log.Error("cpufreq enforcement failed: %v", err)
 		}
-		if err := ctl.enforceCpufreqGovernor(class, cpus...); err != nil {
-			log.Error("cpufreq governor enforcement failed: %v", err)
-		}
 		if err := ctl.enforceUncore(assignments, cpus...); err != nil {
 			log.Error("uncore frequency enforcement failed: %v", err)
 		}


fmuyassarov:
Thanks @askervin for spotting the bug. PTAL
cc @klihub

klihub (dismissed):

fmuyassarov:
@klihub I have now removed the leftover loop.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -128,13 +127,12 @@ func (ctl *cpuctl) enforceCpufreqGovernor(class string, cpusIds ...int) error {
...
 		return fmt.Errorf("non-existent cpu class %q", class)
 	}
-	governor := ctl.classes[class].FreqGovernor
-	for cpu := range cpusIds {
-		log.Info(strconv.Itoa(cpu), governor)
-	}
-	log.Debug("enforcing cpu frequency governor %q on %v", governor, cpusIds)
-	if err := utils.SetScalingGovernorForCPUs(cpusIds, governor); err != nil {
-		return fmt.Errorf("Cannot set cpufreq governor %d: %w", governor, err)
+	if ctl.classes[class].FreqGovernor != "" {
```
askervin, line 130: I think this would be a tiny bit prettier if there would be only one look-up to the class map:
```
if governor := ctl.classes[class].FreqGovernor; governor != "" {
    ....
```

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -128,13 +127,12 @@ func (ctl *cpuctl) enforceCpufreqGovernor(class string, cpusIds ...int) error {
...
+		governor := ctl.classes[class].FreqGovernor
+		log.Debug("enforcing cpu frequency governor %q on %v", governor, cpusIds)
```
askervin, line 132: It would be helpful for debugging to print out the class name, too, like in `enforceCpufreq`.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -128,13 +127,12 @@ func (ctl *cpuctl) enforceCpufreqGovernor(class string, cpusIds ...int) error {
 	if _, ok := ctl.classes[class]; !ok {
```
klihub, line 127: nit: `ctl.classes[class]` is now evalutated 3 times. Once here, to bail out if class is unknown, once to check if there is a class-specific governor set (IOW it is != "") , and finally a 3rd time to get the actual value it is set to. It would be better to reorganize this function to evaluate both `ctl.classes[class]` and `ctl.classes[class].FreqGovernor` at most once.

pkg/resmgr/control/cpu/cpu.go
fmuyassarov: definitely!

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -271,28 +265,26 @@ func (ctl *cpuctl) configure(cfg *cfgapi.Config) error {
...
-				log.Error("cpufreq enforcement on re-configure failed: %v", err)
-			}
-		} else {
+		if _, ok := ctl.classes[class]; !ok {
 			// TODO: what should we really do with classes that do not exist in
 			// the configuration anymore? Now we remember the CPUs assigned to
 			// them. A further config update might re-introduce the class in
 			// which case the CPUs will be reconfigured.
-			log.Warn("class %q with cpus %v missing from the configuration", class, cpus)
+			return fmt.Errorf("non-existent cpu class %q", class)
```
fmuyassarov, line 273: The reason for changing this from a warning to `ErrorF` is that, while the original code doesn't raise an error immediately, it would eventually fail when calling `enforceCpufreq` or `enforceCpufreqGovernor` if the class isn't valid. By restructuring the code this way, we fail early.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -125,16 +124,11 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
...
-	}
-	governor := ctl.classes[class].FreqGovernor
-	for cpu := range cpusIds {
-		log.Info(strconv.Itoa(cpu), governor)
-	}
-	log.Debug("enforcing cpu frequency governor %q on %v", governor, cpusIds)
-	if err := utils.SetScalingGovernorForCPUs(cpusIds, governor); err != nil {
-		return fmt.Errorf("Cannot set cpufreq governor %d: %w", governor, err)
+	if governor := ctl.classes[class].FreqGovernor; governor != "" {
+		log.Debug("enforcing cpu frequency governor %q from class %q on %v", governor, cpusIds)
```
klihub, line 128: The log message uses a format string with 3 %verbs but only 2 arguments.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -271,28 +265,26 @@ func (ctl *cpuctl) configure(cfg *cfgapi.Config) error {
...
+		if err := ctl.enforceCpufreqGovernor(class, cpus.SortedMembers()...); err != nil {
+			log.Error("cpufreq governor enforcement on re-configure failed: %v", err)
 		}
 	}
 	if err := ctl.enforceUncore(assignments); err != nil {
 		log.Error("uncore frequency enforcement on re-configure failed: %v", err)
 	}
 
 	log.Debug("cpu controller configured")
-
```
klihub, line 295: nit: unrelated white space change in a code block not touched by the PR.

fmuyassarov: fixed

pkg/resmgr/control/cpu/cpu.go
fmuyassarov: good catch. Fixed.

klihub (approved):

askervin:
Looking at the governor solution in high level (first implementation and this patch), I'm starting to think that we are adding unnecessary complexity. And I'm slightly worried about it, as we should be adding here other controls as well. If we'd follow the same pattern then, there would be possibly third enforeCpufreqXXX (C-state controls and/or something else) that would needed to be called in the sequence of these two. Currently the two already have different assumption on who is checking if the class exists.

Instead of this complexity and introducing new function --- please tell me if I'm mistaken --- we could add only a single if to the original enforeCpufreq() after checking that the class exists. If governor is defined, then set it to the CPUs, and that's it. We would not need to introduce extra function, or have different error paths.

What do you think, @fmuyassarov and @klihub , could this be stripped down to single "if", which would make much easier to add new controls in the future in the same two-line manner?

fmuyassarov:
> Looking at the governor solution in high level (first implementation and this patch), I'm starting to think that we are adding unnecessary complexity. And I'm slightly worried about it, as we should be adding here other controls as well. If we'd follow the same pattern then, there would be possibly third enforeCpufreqXXX (C-state controls and/or something else) that would needed to be called in the sequence of these two. Currently the two already have different assumption on who is checking if the class exists.
> 
> Instead of this complexity and introducing new function --- please tell me if I'm mistaken --- we could add only a single if to the original enforeCpufreq() after checking that the class exists. If governor is defined, then set it to the CPUs, and that's it. We would not need to introduce extra function, or have different error paths.
> 
> What do you think, @fmuyassarov and @klihub , could this be stripped down to single "if", which would make much easier to add new controls in the future in the same two-line manner?

I think it makes sense.

klihub:
> Looking at the governor solution in high level (first implementation and this patch), I'm starting to think that we are adding unnecessary complexity. And I'm slightly worried about it, as we should be adding here other controls as well. If we'd follow the same pattern then, there would be possibly third enforeCpufreqXXX (C-state controls and/or something else) that would needed to be called in the sequence of these two. Currently the two already have different assumption on who is checking if the class exists.
> 
> Instead of this complexity and introducing new function --- please tell me if I'm mistaken --- we could add only a single if to the original enforeCpufreq() after checking that the class exists. If governor is defined, then set it to the CPUs, and that's it. We would not need to introduce extra function, or have different error paths.
> 
> What do you think, @fmuyassarov and @klihub , could this be stripped down to single "if", which would make much easier to add new controls in the future in the same two-line manner?

Sounds like a really good idea to me. We should consider the governor part of the freq. setting and things get much cleaner.

fmuyassarov:
@askervin @klihub I have refactored the code now. PTAL.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -123,22 +122,6 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
...
-	log.Debug("enforcing cpu frequency governor %q on %v", governor, cpusIds)
-	if err := utils.SetScalingGovernorForCPUs(cpusIds, governor); err != nil {
-		return fmt.Errorf("Cannot set cpufreq governor %d: %w", governor, err)
-	}
-	return nil
-}
-
 // enforceCpufreq enforces a class-specific cpufreq configuration to a cpuset
 func (ctl *cpuctl) enforceCpufreq(class string, cpus ...int) error {
 	if _, ok := ctl.classes[class]; !ok {
```
klihub, line 127: Since we're now using `ctl.classes[class]` quite a bit in this function, cannot we rewrite the sentinel-test to take, keep and use the looked up class if it exists ? So something like

```golang
...
    c, ok := ctl.classes[class]
    if !ok {
        return fmt.Errorf("non-existent cpu class %q", class)
    }

    min, max := int(c.MinFreq), int(c.MaxFreq)
    ...

    if governor = c.FreqGovernor; governor != "" {
        ...
    }

    ...
```

pkg/resmgr/control/cpu/cpu.go
askervin: aThe logic was originally:
```
for each class and associated cpus:
    if there is cpu configuration for this class, apply it (and if applying fails, log an error but **don't tell it to the caller**),
    otherwise log a warning that cpu configuration for a class is missing.
    Then continue with the next class in the next round of the loop.
```

This change proposes logic:
```
for each class and associated cpus:
    if there is no cpu configuration for this class, bail out and **return an error to the caller**, leave remaining classes and cpus unhandled (yet they would have configurations), and forget about possible uncore configurations after the for loop.
    apply the configuration (and if it fails, log an error but don't tell it to the caller)
```

And if you look at the caller of configure() (see Start()), you can see that what used to be a warning will now prevent starting the controller. And if you go up in the call stack (StartStopControllers(), startControllers()....) that now missing CPU properties for single class prevents starting the whole resource manager (policy).

So I think this change would fail early, and do it in a very fatal way.

I'd prefer the original, quite forgiving handling of missing cpu class information. Maybe we can make these issues more visible to the user through policy status at some point. However, those changes would be topics for other PRs.

askervin (changes requested):
In my opinion, we should not introduce a fatal failure due to missing parameters for a CPU class.

klihub:
> In my opinion, we should not introduce a fatal failure due to missing parameters for a CPU class.

+1. Definitely not.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -271,21 +263,17 @@ func (ctl *cpuctl) configure(cfg *cfgapi.Config) error {
...
-				log.Error("cpufreq enforcement on re-configure failed: %v", err)
-			}
-		} else {
+		if _, ok := ctl.classes[class]; !ok {
 			// TODO: what should we really do with classes that do not exist in
 			// the configuration anymore? Now we remember the CPUs assigned to
 			// them. A further config update might re-introduce the class in
 			// which case the CPUs will be reconfigured.
 			log.Warn("class %q with cpus %v missing from the configuration", class, cpus)
 		}
```
klihub, line 272: If you flip the if-ok-else branch to if-!ok, you either need to use a if-!ok-else construct (IOW, move enforceCpufreq into an else block) or add a continue here in the if block. Otherwise we'll log a warning for a non-existent class but still try to enforce CPU frequencies and the governor for it...

klihub (approved):

fmuyassarov:
ping @askervin

askervin (approved):
LGTM, Thanks @fmuyassarov!

PR #381, opened 2024-10-14
libmem: typo and thinko fixes.

klihub:
This patch set fixes
- a docstring/comment typo
- node mask parsing of empty masks
- node mask parsing ID range upper boundary (should be inclusive)
- node mask parsing incorrect slice index for single node IDs

It also adds an obviously necessary minimal test for `ParseNodeMask()`.


fmuyassarov (approved):
LGTM

askervin (approved):
LGTM.

PR #382, opened 2024-10-16
dev: add Tiltfile for local development

fmuyassarov:
This commit adds a Tiltfile to simplify local development with Tilt.

How to use it?
To run balloons policy: `tilt up balloons`
To run topology-aware policy: `tilt up topology-aware`

All it does is:
1. build the binary
2. create a container image and push to a temporary registry
3. deploy the policy Helm chart

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,0 +1,63 @@
+# -*- mode: Python -*-
+
+# Extensions 
+load('ext://helm_resource', 'helm_resource', 'helm_repo')
+
+# Specifies that Tilt is allowed to run against the specified k8s context name. 
+allow_k8s_contexts('kubernetes-admin@kubernetes')
+# Anonymous, ephemeral image registry.
+default_registry('ttl.sh')
+
+# Run a defined set of services via config.set_enabled_resources.
+config.define_string_list("to-run", args=True)
+cfg = config.parse()
+groups = {
+    'balloons': ['binary-build-logs', 'controller-logs'],
+    'topology-aware': ['binary-build-logs', 'controller-logs'],
+}
+
+resources = []
+for resource_name in cfg.get('to-run', []):
+    if resource_name in groups:
+        resources += groups[resource_name]
+config.set_enabled_resources(resources)
+
+# Fail if the policy name is not provided.
+if len(resources) == 0:
+    fail("🚨 ERROR: No policy passed! Please run: tilt up <POLICY_NAME>")
+
+POLICY = resource_name
+IMAGE_BASE = "ttl.sh/ghcr.io/containers/nri-plugins/nri-resource-policy-"
+IMAGE = IMAGE_BASE + POLICY
+COMPILE_CMD = ('make BINARIES= OTHER_IMAGE_TARGETS= PLUGINS=nri-resource-policy-' + POLICY + ' build-plugins')
+DEPS = ['./pkg', './cmd', './deployment']
+
+###################################
+# The main tasks are executed here.
+###################################
+
+# Builds a binary.
+local_resource(
+    'binary-build-logs',
+    COMPILE_CMD,
+    deps=DEPS
+)
+
+# Builds a container image.
+docker_build(
+    IMAGE,
+    '.',
+    dockerfile='./cmd/plugins/' + POLICY + '/Dockerfile'
+)
+
+# Deploy the policy Helm chart.
+helm_resource(
+    'controller-logs',
+    './deployment/helm/' + POLICY,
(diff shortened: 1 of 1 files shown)


Tiltfile
```diff
@@ -0,0 +1,65 @@
...
+# Fail if the policy name is not provided.
+if len(resources) == 0:
+    fail("🚨 ERROR: No policy passed! Please run: tilt up <POLICY_NAME>")
+
+POLICY = resource_name
+IMAGE_BASE = "ttl.sh/ghcr.io/containers/nri-plugins/nri-resource-policy-"
+IMAGE = IMAGE_BASE + POLICY
+COMPILE_CMD = (
+    'cd cmd/plugins/' + POLICY +
+    '; CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -x -o ../../../build/bin/nri-resource-policy-' + POLICY + ' ./'
```
klihub, line 34: I'm fine with an in-repo Tiltfile, if you want to use tilt and this helps you. But if we do so, I very much want the Tiltfile to build binaries and images the 'official way', using our provided build infra instead of trying to repeat the necessary steps here manually... which looks like just asking for unnecessary maintenance trouble to me.

klihub: So for instance, use something like `make BINARIES= OTHER_IMAGE_TARGETS= PLUGINS=nri-resource-policy-+POLICY  build-plugins` here, if you only want to build a single policy binary.

Tiltfile
```diff
@@ -0,0 +1,65 @@
...
+    'image-build-logs',
+    COMPILE_CMD,
+    deps=['./pkg', './cmd', './deployment']
+)
+
+# Builds a container image.
+docker_build(
+    IMAGE,
+    '.',
+    dockerfile='./cmd/plugins/' + POLICY + '/Dockerfile'
```
klihub, line 52: Ditto here. Try something like `'make BINARIES= OTHER_IMAGE_TARGETS= PLUGINS=nri-resource-policy'+POLICY build-images` or something along those lines here...

klihub (changes requested):
Should use the 'stock build infra' already offered by the repo instead of hand-rolling copies of the commands necessary to build plugin binaries or images.

askervin:
I suppose this is ok, but the maintenance burden should be minimized. That is, go or docker build logic and parameterization should not be duplicated but used through `make` instead.

fmuyassarov:
I have updated the binary build process to use the existing Makefile target. However, we can't replace the container image build process (`docker_build`) with a Makefile target. The `docker_build` is needed for instructing Tilt on how to build the image. Without it, Tilt won't be able to trigger dependent tasks, such as deploying Kubernetes manifests into a cluster.

Tiltfile
fmuyassarov: docker_build is needed for instructing the Tilt how to build the container image. Even though it could be achieved via something like this


```
BUILD_CMD = ('make BINARIES= OTHER_IMAGE_TARGETS= PLUGINS=nri-resource-policy-' + POLICY + ' build-images')
local_resource(
    'image-build-logs',
    BUILD_CMD,
    resource_deps=['binary-build-logs'],
    deps=DEPS
)
```
it is still important to have a docker_build step to indicate how the image needs to be build when required.

klihub:
> However, we can't replace the container image build process (`docker_build`) with a Makefile target. The `docker_build` is needed for instructing Tilt on how to build the image. Without it, Tilt won't be able to trigger dependent tasks, such as deploying Kubernetes manifests into a cluster.

That should be fine, the docker builds/Dockerfiles use the build system.

fmuyassarov:
> > However, we can't replace the container image build process (`docker_build`) with a Makefile target. The `docker_build` is needed for instructing Tilt on how to build the image. Without it, Tilt won't be able to trigger dependent tasks, such as deploying Kubernetes manifests into a cluster.
> 
> That should be fine, the docker builds/Dockerfiles use the build system.

Do you think current state is okay to merge?

klihub:
> > > However, we can't replace the container image build process (`docker_build`) with a Makefile target. The `docker_build` is needed for instructing Tilt on how to build the image. Without it, Tilt won't be able to trigger dependent tasks, such as deploying Kubernetes manifests into a cluster.
> > 
> > 
> > That should be fine, the docker builds/Dockerfiles use the build system.
> 
> Do you think current state is okay to merge?

Yes, I think it should be fine.

klihub (approved):

askervin (approved):
LGTM

PR #383, opened 2024-10-16
sysfs: enable faking CPU cache configurations using OVERRIDE_SYS_CACHES

askervin:
This PR is an enabler for testing CPU allocations when there are several CPUs that share the same level 2 cache.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -15,6 +15,7 @@
 package sysfs
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -277,6 +278,8 @@ var (
 		PerformanceCore: "OVERRIDE_SYS_CORE_CPUS",
 		EfficientCore:   "OVERRIDE_SYS_ATOM_CPUS",
 	}
+	cacheEnvOverridesVar  = "OVERRIDE_SYS_CACHES"
+	cacheEnvOverridesJson = os.Getenv(cacheEnvOverridesVar)
 )
 
 // MemInfo contains data read from a NUMA node meminfo file.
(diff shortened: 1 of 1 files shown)


pkg/sysfs/system.go
```diff
@@ -306,6 +309,21 @@ type Cache struct {
...
 }
 
+// cacheOverrides is a list of cache overrides for specific CPU sets.
+// Every CPU set gets cache id that starts from 0 on each override object.
+// Example: two 128 MB L3 Caches shared by CPUs 0-15,32-47 (socket 0)
+// and 16-31,48-63 (socket 1). The cache id is 0 for the first, 1 for the second.
+// [{"cpusets": ["0-15,32-47", "16-31,48-63"], "level": 3, "size": "128M"}]
+type cacheOverrides []cacheOverride
+type cacheOverride struct {
+	Cpusets []string // cpusets to which this override applies
```
klihub, line 319: nit: what is the reason for allowing multiple cpusets strings here ? Isn't it always possible to present them as a single one ?

klihub: Ah, ok. It is for being able to use different cache IDs for different CPU sets, which I guess is necessary to be able to emulate some hybrid core platforms...

askervin: Yes, this gives a possibility to have quite realistic id numbering without forcing user to write a lot (by specifying ids), or a developer to write a lot (come up with a magic that combines CPU topology with caches). Something like:

```
[
  {"cpusets": ["0-7", "8-15"], "level": 3, "size": "128M"},
  {"cpusets": ["0-3", "4-7"], "level": 2, "size": "16M"},
  {"cpusets": ["8-11", "12-15"], "level": 2, "size": "16M"},
  {"cpusets": ["0-1", "2-3", "4-5", "6-7"], "level": 1, "size": "256k", "kind": "d"},
  {"cpusets": ["0-1", "2-3", "4-5", "6-7"], "level": 1, "size": "128k", "kind": "i"},
  {"cpusets": ["8-9", "10-11", "12-13", "14-15"], "level": 1, "size": "256k", "kind": "d"},
  {"cpusets": ["8-9", "10-11", "12-13", "14-15"], "level": 1, "size": "128k", "kind": "i"}
  ]
```

pkg/sysfs/system.go
```diff
@@ -1526,6 +1554,88 @@ func (sys *system) discoverSst() error {
...
+		size *= 1 << 40
+	case "":
+	default:
+		return 0, fmt.Errorf("unknown unit %q", unit)
+	}
+	return size, nil
+}
+
+// Parse cache overrides from a JSON string.
+func parseCpuCacheOverrides(overrideJson string) map[int][]*Cache {
```
klihub, line 1582: nit: consider changing this to return signature of `(map[int][]*Cache, error)` and return a `nil, fmt.Errorf("failed to parse CPU cache overrides: %w", err)` in case of a parsing or some other error, to be consistent with how core kind overrides work (IIRC).

klihub (commented):
LGTM, although I have a small nitgestion about consistency with core kind overrrides, IOW to fail on parse errors instead of logging the errors but then starting up with a wrong emulated cache layout... considering that this is intended for test use.

pkg/sysfs/system.go
askervin: Return value changed.

Moved the parse function to be called only on demand as part of discoverCPU() and next to discoverCache(). This enabled making it sys.parseCpuCacheOverrides(), so I can use the same Debug() as everywhere else.

I think it's much better now. Thank you for these comments, once again!

klihub (approved):

Issue #385, opened 2024-10-18
The default balloons configuration should run bestefforts on idle CPUs


askervin:
### Description

Improve the default balloons configuration so that it will have its own balloon type for besteffort containers, and that balloons of this type will be able to use idle CPUs in a convenient zone level.

TBD: should we allow creating multiple balloons from this type? Like one per CPU socket?

Note that in case of multiple besteffort balloon instances, the policy needs to be able to spread besteffort containers to these instances. At minimum by striping, or perhaps based on headroom on allowed (= allocated + borrowed) CPUs.

### Rationale

Currently besteffort containers do not get properly spread to balloon instances (due to their zero CPU consumption), and they can therefore interfere with running containers in the same balloons.

askervin:
Fixed in PR #391.

PR #386, opened 2024-10-22
balloons: improve handling of containers with no CPU requests

askervin:
Containers without CPU requests always pile up to the first balloon where they fit.

This patch enables having multiple balloons in the system so that this kind of containers will be spread equally into them.

This patch is an enabler to fix the issue #385.


fmuyassarov (approved):
LGTM

klihub (approved):

PR #387, opened 2024-10-23
resmgr: fix "qosclass" in policy expressions

askervin:
Resolving "qosclass" failed silently and returned an empty string due to error when "casting" object from interface{} to string. This fails because the resolved value behind the interface{} is of type k8s.io/api/core/v1.PodQOSClass and not (directly) a string. Making sure that the interface{} is directly a string solves the problem.


klihub (approved):
LGTM, although I wonder if we could have other silent failures or if this is the only such error path...

PR #391, opened 2024-11-01
helm: new balloons default configuration

askervin:
Introduce a default configuration where Guaranteed containers are all executed on dedicated CPUs while BestEffort and Burstable share CPUs in package (socket) level.


klihub (approved):
LGTM.

PR #392, opened 2024-11-01
topology-aware: implement metrics collection.

klihub:
Implement metrics collection for the topology-aware policy. Collect per zone (IOW per pool) prometheus
  - name, cpuset and memset
  - shared pool capacity, allocation, available amount
  - memory capacity, allocation, available amount
  - number of containers
  - number of containers in the shared pool

### DIFF
--- a/cmd/plugins/topology-aware/policy/metrics.go
+++ b/cmd/plugins/topology-aware/policy/metrics.go
@@ -1,0 +1,299 @@
+// Copyright The NRI Plugins Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package topologyaware
+
+import (
+	"slices"
+	"strings"
+
+	libmem "github.com/containers/nri-plugins/pkg/resmgr/lib/memory"
+	policyapi "github.com/containers/nri-plugins/pkg/resmgr/policy"
+	"github.com/containers/nri-plugins/pkg/utils/cpuset"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	descZone = iota
+	descZoneCpuSharedCapacity
+	descZoneCpuSharedAssigned
+	descZoneCpuSharedAvailable
+	descZoneMemCapacity
+	descZoneMemAssigned
+	descZoneMemAvailable
+	descZoneContainerCount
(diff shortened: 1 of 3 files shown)


cmd/plugins/topology-aware/policy/metrics.go
```diff
@@ -0,0 +1,299 @@
...
+
+func (zm *ZoneMetrics) Collect() []prometheus.Metric {
+	if zm == nil {
+		return nil
+	}
+
+	var metrics []prometheus.Metric
+
+	metrics = append(metrics,
+		prometheus.MustNewConstMetric(
```
marquiz, line 151: Why the use of "throwaway" const metric? I'd think about initializing a list of persistent gauges and using the Set method to update the value.

pkg/resmgr/lib/memory/zones.go
```diff
@@ -154,6 +159,27 @@ func (a *Allocator) zoneFree(zone NodeMask) int64 {
...
+	}
+
+	// Cap available amount to the smallest available free in any of our ancestors.
+	for _, z := range a.zones {
+		if z.nodes != zone && (z.nodes&zone) == zone {
+			if free := a.zoneFree(z.nodes); free < available {
+				available = free
+			}
+		}
+		if available == 0 {
```
askervin, line 175: Should we worry about available < 0 case here, too, like in line 164?

pkg/resmgr/lib/memory/zones.go
```diff
@@ -45,6 +45,11 @@ func (a *Allocator) ZoneFree(zone NodeMask) int64 {
 	return a.zoneFree(zone & a.masks.nodes.hasMemory)
 }
 
+// ZoneAvailable returns the amount of available memory in the zone.
```
askervin, line 48: The meaning of "available" is not very clear to me. Maybe this public method comment could help understanding why ZoneAvailable != ZoneFree (where ZoneFree = ZoneCapacity - ZoneUsage). Can it be that ZoneAvailable at t=0 is smaller than ZoneUsage at t=1?

askervin:
Beside these comments, I think this already looks very good.

cmd/plugins/topology-aware/policy/metrics.go
klihub: Because I mindlessly copied the pattern of implementation from balloons without understanding what I am actually doing... Let me try to read up on this a bit more and fix it.

pkg/resmgr/lib/memory/zones.go
klihub: ZoneFree returns the amount of free capacity in this zone, taking into account allocations in this zone and any only this zone and its subzones. ZoneAvailable returns ZoneFree capped to the minimum free capacity of any of the (larger super-)zones this zone is a genuine subzone of. IOW, it returns the amount of memory allocatable from this zone.

pkg/resmgr/lib/memory/zones.go
klihub: Yes, does not make sense to return anything < 0 here...

klihub:
@marquiz @askervin I close this and reopened a reworked version stacked on top of the rest of my metrics series as #406.

PR #393, opened 2024-11-02
helm: enable prometheus autodiscovery

klihub:
helm: allow autodiscovery by prometheus.

Allow autodiscovery of prometheus metrics by annotation. IOW, annotate our plugins' pods according to the Helm `config.instrumentation.{prometheusExport,httpEndpoint}` values. Since this PR adds annotations to our plugins' pods, add support for additional injection of other custom annotations using the Helm `plugin.annotations` value.
 
Also, since it's now uneccessary to expose our custom metrics port on the host side, remove the required and dedicated `hostPort` Helm value. Add instead an optional generic `ports` array to allow exposing arbitrary ports to the host. Remove `hostPort` from the `nriplugindeployments` CRD as well.


fmuyassarov:
I'm not sure if I'm missing some context, but could you clarify why we need to introduce these annotations when we already have Helm flags that control the instrumentation of the balloons policy CR? The only reason I can think of for having two methods to enable scraping on the pod would be to allow overriding the default port and scraping settings. Is that the case, or is there something else at play here?

Also, the `nri.plugin.annotations` setting initially gave an impression that users could pass any annotation they wanted, suggesting it wasn't exclusively linked to Prometheus. However,  in the daemonSet spec, it seems that whatever is passed through 1nri.plugin.annotations1 ultimately maps to `prometheus.io/port`, which is a bit odd.

klihub:
> I'm not sure if I'm missing some context, but could you clarify why we need to introduce these annotations when we already have Helm flags that control the instrumentation of the balloons policy CR? The only reason I can think of for having two methods to enable scraping on the pod would be to allow overriding the default port and scraping settings. Is that the case, or is there something else at play here?

That will cause prometheus to automatically discover and start collecting metrics from the pod.

klihub:
> Also, the nri.plugin.annotations setting initially gave an impression that users could pass any annotation they wanted, suggesting it wasn't exclusively linked to Prometheus. However, in the daemonSet spec, it seems that whatever is passed through 1nri.plugin.annotations1 ultimately maps to prometheus.io/port, which is a bit odd.

If that is the case then I screwed up the latest rebase. The ide is exactly to provide a way to inject arbitrary annotations. But in addition to any custom annotations by Helm config we always annotate for prometheus auto-discovery.

fmuyassarov:
> prometheus auto-discovery

btw, what is prometheus auto-discovery, do you have some reference documentation ?

klihub:
> > prometheus auto-discovery
> 
> btw, what is prometheus auto-discovery, do you have some reference documentation ?

I just googled for autodiscovery then found a few articles describing it. For instance, https://acagroup.be/en/blog/auto-discovery-of-kubernetes-endpoint-services-prometheus/, https://learn.microsoft.com/en-us/azure/azure-monitor/containers/prometheus-metrics-scrape-configuration?tabs=CRDConfig%2CCRDScrapeConfig%2CConfigFileScrapeConfigBasicAuth%2CConfigFileScrapeConfigTLSAuth#enable-pod-annotation-based-scraping, or https://developer.ibm.com/tutorials/monitoring-kubernetes-prometheus/).

deployment/helm/balloons/templates/daemonset.yaml
```diff
@@ -67,11 +73,14 @@ spec:
...
             {{- end }}
+        {{- if (not (eq .Values.ports nil)) }}
           ports:
-            - containerPort: 8891
-              protocol: TCP
-              hostPort: {{ .Values.hostPort }}
-              name: metrics
+          {{- range $port := .Values.ports }}
+            - name: {{ $port.name }}
+              container: {{ $port.container }}
```
fmuyassarov, line 80:
```suggestion
              containerPort: {{ $port.containerPort }}
```

deployment/helm/balloons/values.schema.json
```diff
@@ -114,11 +113,34 @@
...
         "podPriorityClassNodeCritical": {
             "type": "boolean"
+        },
+        "ports": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "required": [
+                    "name",
+                    "container",
```
fmuyassarov, line 125:
```suggestion
                    "containerPort",
```

deployment/helm/balloons/values.schema.json
```diff
@@ -114,11 +113,34 @@
...
+                    "host"
+                ],
+                "properties": {
+                    "name": {
+                        "type": "string"
+                    },
+                    "container": {
```
fmuyassarov, line 132:
```suggestion
                    "containerPort": {
```

deployment/helm/balloons/values.yaml
```diff
@@ -99,3 +102,13 @@ nodeSelector: []
...
 podPriorityClassNodeCritical: true
+
+# extra ports to expose to the host
+#
+# Example
+#
+# # Note: This is not necessary for collecting metrics.
+# ports:
+#   - name: metrics
+#     container: 8891
```
fmuyassarov, line 113:
```suggestion
#     containerPort: 8891
```

fmuyassarov: and the same issue in other instances

deployment/helm/balloons/values.yaml
```diff
@@ -28,6 +28,8 @@ config:
     klog:
       skip_headers: true
   instrumentation:
+    httpEndpoint: ":8891"
+    prometheusExport: true
```
fmuyassarov, line 32: Adding this setting will turn on metrics exposing by default. Is that your intention?

deployment/helm/template/values.yaml
```diff
@@ -66,3 +69,13 @@ tolerations: []
...
+
+# extra ports to expose to the host
+#
+# Example
+#
+# # Note: This is not necessary for collecting metrics.
+# ports:
+#   - name: metrics
+#     container: 8891
+#     host: 8891
```
fmuyassarov, line 81: ` helm install ....--set ports[0].name=metrics   --set ports[0].container=8891   --set ports[0].host=8891`

can/should we add another example of how to pass those fields via cmd when doing helm install?

fmuyassarov:
Don't we need `prometheus.io/path: '/metrics'` annotation alongside other too?

fmuyassarov:
Also, I would merge https://github.com/containers/nri-plugins/pull/393/commits/b437b94d90bdd57c610a3a6ed2df067d57349d51 and  https://github.com/containers/nri-plugins/pull/393/commits/cdce5b6959c5d380a8f3b086594ab0b07616093f into a single commit.

deployment/helm/balloons/templates/daemonset.yaml
klihub: Ah, thanks for spotting. Fixed (but I still chose to not tautologically repeat 'Port' in the Helm value schema port property names).

deployment/helm/balloons/values.schema.json
klihub: That's not necessary. I chose not to tautologically repeat 'Port' in the port object property names.

deployment/helm/template/values.yaml
klihub: Well we could, but I think that's the only port we can have open ATM in any of the policies.

deployment/helm/balloons/values.yaml
klihub: Thanks for spotting. Probably no reason to change the current defaults in that regard. Changed to `false`.

Another thing I realized is that maybe we always want to unconditionally annotate the pod for autodiscovery, even if we don't start with prometheus metrics exporting enabled. Since we can
can dynamically change the configuration to enable it any time.

klihub:
> > There is no existing helm flag to control instrumentation, only a CR configuration setting.
> 
> There is:
> 
> https://github.com/containers/nri-plugins/blob/a0d1476769b0fa4d7dae5f6871f46cf4ef26f3bb/deployment/helm/balloons/values.yaml#L43

That is not a genuine Helm-level setting. It is part of the Helm-level opaque config YAML blob we inject into default policy CRD. If you look at `values.schema.json`, it (intentionally) says nothing about `config.instrumentation.prometheusExport` or even `config.instrumentation`.

deployment/helm/balloons/values.yaml
```diff
@@ -54,15 +56,16 @@ config:
 plugin-test:
   enableAPIs: false
 
-hostPort: 8891
-
 resources:
```
fmuyassarov, line 59: This is going to break the metric expose. Because, you don't have a default value for `port.host` field. Meaning, if you set `prometheusExport: true`, you don't get any metric because the port is closed.

fmuyassarov (changes requested):

deployment/helm/balloons/values.yaml
klihub: > This is going to break the metric exposure. Because, you don't have a default value for `port.host` field. Meaning, if you set `prometheusExport: true`, you don't get any metric because the port is closed.

Why would it break it ? Metric exposure does not need any port mapping as this PR switches to prometheus autodiscovery by annotation. Then this PR also removes that metrics-dedicated `hostPort` Helm value altogether. It is replaced with an optional `ports` array. That is a generic port mapping setting and defaults to `nil`. In the daemonset template we evaluate `ports` with a `range`, emitting the necessary corresponding `name`, `hostPort` and `containerPort` triplet `port` entry.

So if metric exposure is now broken I don't see how, neither could I reproduce it in my local tests. Can you elaborate more on that ?

askervin (commented):
I've got only one question on this.

Could we enable installation so that only the port `name` is defined but the `host` is not? That is, we would not ever specify `hostPort` unless user very very much wants it there.

klihub:
> I've got only one question on this.
> 
> Could we enable installation so that only the port `name` is defined but the `host` is not? That is, we would not ever specify `hostPort` unless user very very much wants it there.

Updated, making `ports.host/hostPort` optional.

askervin (approved):
LGTM

PR #394, opened 2024-11-03
topology-aware: correctly reconfigure implicit affinities for configuration changes.

klihub:
Fix the handling of implicit affinities corresponding to the `colocate{Pods,Namespaces}` configuration settings. Unregister and reregister disabled and enabled implicit affinities as necessary when we receive a new configuration.

### DIFF
--- a/cmd/plugins/topology-aware/policy/topology-aware-policy.go
+++ b/cmd/plugins/topology-aware/policy/topology-aware-policy.go
@@ -111,7 +111,9 @@ func (p *policy) Setup(opts *policyapi.BackendOptions) error {
 		return policyError("failed to initialize %s policy: %w", PolicyName, err)
 	}
 
-	p.registerImplicitAffinities()
+	if err := p.registerImplicitAffinities(); err != nil {
+		return policyError("failed to initialize %s policy: %w", PolicyName, err)
+	}
 
 	log.Info("***** default CPU priority is %s", defaultPrio)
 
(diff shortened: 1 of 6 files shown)


pkg/resmgr/cache/cache.go
```diff
@@ -390,6 +390,8 @@ type Cache interface {
 	EvaluateAffinity(*Affinity) map[string]int32
 	// AddImplicitAffinities adds a set of implicit affinities (added to all containers).
 	AddImplicitAffinities(map[string]ImplicitAffinity) error
+	// AddImplicitAffinities deletes a set of implicit affinities by name.
+	DeleteImplicitAffinities(names []string)
```
fmuyassarov, line 394:
```suggestion
	// 	DeleteImplicitAffinities(names []string) deletes a set of implicit affinities by name.
	DeleteImplicitAffinities(names []string)
```

klihub: Ouch, thanks for spotting that ! Fixed.

cmd/plugins/topology-aware/policy/topology-aware-policy.go
```diff
@@ -484,6 +484,8 @@ func (p *policy) Reconfigure(newCfg interface{}) error {
 		return policyError("failed to reconfigure: %v", err)
 	}
 
+	p.registerImplicitAffinities()
```
fmuyassarov, line 487: since we are here, can we perhaps handle the error instead of ignoring here?

klihub: Okay. I added an error check to both call sites, `Setup` and `Reconfigure`.

askervin (approved):
LGTM

fmuyassarov (approved):

PR #395, opened 2024-11-04
ballons: add memoryTypes to balloon types

askervin:


pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -149,6 +149,10 @@ type BalloonDef struct {
 	// this will be the number of CPUs reserved for it even if a container
 	// would request less.
 	MinCpus int `json:"minCPUs,omitempty"`
+	// MemoryTypes lists memory types allowed to containers in a
+	// balloon. Supported types are: DRAM, HBM, PMEM. By default
+	// all memory types in the system are allowed.
+	MemoryTypes []string `json:"memoryTypes,omitempty"`
```
klihub, line 155: I wonder if something like this would provide working extended validation for this:

```golang
type BalloonDef struct {
...
        // MemoryTypes lists memory types allowed to containers in a
        // balloon. Supported types are: DRAM, HBM, PMEM. By default
        // all memory types in the system are allowed.
        // +kubebuilder:validation:XValidation:rule="self.values.all(value == 'DRAM' || value == 'HBM' || value == 'PMEM')", message="invalid memoryTypes value, should be one of DRAM, HBM, or PMEM"
	MemoryTypes []string `json:"memoryTypes,omitempty"`
...
}
```

klihub: Okay, so we'll need this after all (since CEL-expressions are just unusable for this as we found out due to very limited expression evaluation runtime budget which is getting exceeded here due to double nested slices):
```golang
...
    // +kubebuilder:validation:items:Pattern=^(DRAM|HBM|PMEM)$
    MemoryTypes []string `json:"memoryTypes,omitempty"`
...
```

klihub (approved):
LGTM.

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
askervin: Thank you, @klihub! This has now a nice solution and we did not consume CEL budget for it.

Makefile
```diff
@@ -475,7 +475,7 @@ CLIENT_GEN ?= $(LOCALBIN)/client-gen
 GIT_CLONE    := git clone
 
 # External generation tool versions
-CONTROLLER_TOOLS_VERSION ?= v0.14.0
+CONTROLLER_TOOLS_VERSION ?= v0.16.5
```
fmuyassarov, line 478: you don't need this, since it was already bumped here: https://github.com/containers/nri-plugins/commit/b51c35bb921d83986d3daf0b3e9891d8ccb5304b

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -149,6 +149,12 @@ type BalloonDef struct {
 	// this will be the number of CPUs reserved for it even if a container
 	// would request less.
 	MinCpus int `json:"minCPUs,omitempty"`
+	// MemoryTypes lists memory types allowed to containers in a
+	// balloon. Supported types are: DRAM, HBM, PMEM. By default
+	// all memory types in the system are allowed.
+	// +kubebuilder:validation:MaxItems=3
+	// +kubebuilder:validation:items:Pattern=`^(DRAM|HBM|PMEM)$`
+	MemoryTypes []string `json:"memoryTypes,omitempty"`
```
fmuyassarov, line 157: @klihub have you verified this?

because I just deployed it a cluster and tried to add non DRAM|HBM|PMEM values and validation didn't complain
```YAML
apiVersion: config.nri/v1alpha1
kind: BalloonsPolicy
metadata:
  annotations:
    meta.helm.sh/release-name: controller-logs
    meta.helm.sh/release-namespace: kube-system
  creationTimestamp: "2024-11-07T12:46:22Z"
  generation: 2
  labels:
    app.kubernetes.io/instance: controller-logs
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: nri-resource-policy-balloons
    helm.sh/chart: nri-resource-policy-balloons-v0.0.0
  name: default
  namespace: kube-system
  resourceVersion: "5660"
  uid: 2f144427-c400-469e-999c-f8a0bf476101
spec:
  allocatorTopologyBalancing: true
  balloonTypes:
  - allocatorPriority: normal
    memoryTypes:
    - foo
    - bar
    minBalloons: 0
    minCPUs: 1
    name: default
    namespaces:
    - default
    preferIsolCpus: false
    shareIdleCPUsInSame: system

```

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -149,6 +149,12 @@ type BalloonDef struct {
 	// this will be the number of CPUs reserved for it even if a container
 	// would request less.
 	MinCpus int `json:"minCPUs,omitempty"`
+	// MemoryTypes lists memory types allowed to containers in a
+	// balloon. Supported types are: DRAM, HBM, PMEM. By default
+	// all memory types in the system are allowed.
+	// +kubebuilder:validation:MaxItems=3
+	// +kubebuilder:validation:items:Pattern=`^(DRAM|HBM|PMEM)$`
```
fmuyassarov, line 156: I would also add `// +listType=set` too to avoid having duplicated values.

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
klihub: @fmuyassarov Are you sure you really have that CRD in place. Just asking to make sure you didn't try to helm install and helm silently refused to replace an existing version without that validation pattern in place, and that's why it does not work...

fmuyassarov: I can try to delete and install crds too to double check. I must say I am not 100% sure. Will update soon.

fmuyassarov: You were right, I was still using the old CRD. 
But at least we should consider the next comment below to avoid this

```
    memoryTypes:
    - HBM
    - HBM

```

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
askervin: Added +listType, thanks @fmuyassarov!

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
askervin: Now handled with listType=set

askervin:
Switched from Pattern to XValidation CEL expression for a nice error message.

fmuyassarov (approved):
Thanks @askervin 
LGTM

PR #401, opened 2024-11-08
Tighten crd validation to have only one occurrence of a value in a list

fmuyassarov:
Tighten CRD validation to have only one occurrence of a value in a list.

### DIFF
--- a/pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
+++ b/pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
@@ -129,6 +129,7 @@ type BalloonDef struct {
 	// Namespaces control which namespaces are assigned into
 	// balloon instances from this definition. This is used by
 	// namespace assign methods.
+	// +listType=set
 	Namespaces []string `json:"namespaces,omitempty"`
 	// GroupBy groups containers into same balloon instances if
 	// their GroupBy expressions evaluate to the same group.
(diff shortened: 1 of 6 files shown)


pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -220,11 +221,13 @@ type BalloonDef struct {
 	ShareIdleCpusInSame CPUTopologyLevel `json:"shareIdleCPUsInSame,omitempty"`
 	// PreferCloseToDevices: prefer creating new balloons of this
 	// type close to listed devices.
+	// +listType=set
```
askervin, line 224: The documentation says:

"Equality on arrays with x-kubernetes-list-type of set or map ignores element order."

Would this mean that if a user changes the order of items in the PreferCloseToDevices list, then some validation logic could say that a CR has not been changed? Or that it does not matter in which order items in the list are displayed or stored? Or that CEL expressions that would compare these lists of two balloon types would consider preferences similar only if they contain the same devices?

Any of these cases would be problematic, because we document that the order of items in this list is important. A device listed before another device has higher preference than the other.

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -220,11 +221,13 @@ type BalloonDef struct {
...
 	PreferCloseToDevices []string `json:"preferCloseToDevices,omitempty"`
 	// PreferFarFromDevices: prefer creating new balloons of this
 	// type far from listed devices.
 	// TODO: PreferFarFromDevices is considered too untested for usage. Hence,
 	// for the time being we prevent its usage through CRDs.
+	// +listType=set
```
askervin, line 230: I'm not sure if we should add validation to a field that is hidden. And here order matters, too.

Furthermore, as we fill PreferFarFromDevices automatically, I wonder what happens if the automation has added a device twice in the list, and the result would be stored into a yaml, then restored from the yaml, it would be quite surpising if validation would reject a yaml that seemed to be a working configuration.

I'd propose leaving this out.

fmuyassarov:
Thanks for the comments @askervin 
I think i will try to look into this again once we have migrated to v2 generators. Closing for now.

PR #403, opened 2024-11-09
[stacked 1/5] metrics: cleanup metrics registration, collection and gathering.

klihub:
This patch series refactors our metrics collection code, moving most of the code to handle metrics registration, collection, and gathering into a single package in `pkg/metrics`. In particular, this series

- consolidates all common metrics-handling code in `pkg/metrics`
- removes metrics polling scattered around in `resmgr`, `resmgr/{metrics,policycollector}`
- moves policy-specific metrics collection/polling to the `policy` package itself
- switches from `opencensus` to `promhttp` for metrics exporting in `pkg/instrumentation`


pkg/apis/config/v1alpha1/metrics/config.go
```diff
@@ -14,21 +14,15 @@
...
-
-	if pend == nil {
-		log.Debug("no data to proxy to prometheus...")
-	} else {
-		log.Debug("proxying data to prometheus...")
-	}
-
-	return pend, nil
+// Config provides runtime configuration for metrics collection.
+// +k8s:deepcopy-gen=true
```
fmuyassarov, line 18: please drop this line, it is not used anymore. See the migration guide for the reference: https://github.com/kubernetes-sigs/kubebuilder/blob/782a4da58456003c7da3c3ccdac6558328218de8/docs/book/src/migration/legacy/migration_guide_v1tov2.md

pkg/cgroupstats/register/collector.go
```diff
@@ -0,0 +1,39 @@
...
+	err := metrics.Register(
+		"stats",
+		cgroupstats.NewCollector(),
+		metrics.WithGroup("cgroup"),
+		metrics.WithCollectorOptions(
+			metrics.WithoutNamespace(),
+		),
+	)
+	if err != nil {
+		log.Error("failed register cgroup/stats collector: %v", err)
```
fmuyassarov, line 37:
```suggestion
		log.Error("failed to register cgroup/stats collector: %v", err)
```

pkg/apis/config/v1alpha1/metrics/config.go
klihub: Okay, so that talks about kubebuilder v2. What is the relation of that to `controller-tools v0.16.5` and `code-generator v0.28.1`. Are those v2 or v1 ? We are non-urgently a bit stuck with those versions ATM, until we figure out what is the equivalent of this
```shell
code-generator/generate-groups.sh client \
            github.com/containers/nri-plugins/pkg/generated \
            github.com/containers/nri-plugins/pkg/apis \
            "config:v1alpha1" --output-base=. \
            --go-header-file=./docs/license-header.go.txt && \
        rm -fr pkg/generated && \
        mv github.com/containers/nri-plugins/pkg/generated pkg && \
        rm -fr github.com
```
in the new version, which is neither a drop-in replacement nor offers a backward-compatibility wrapper script.

That is why we have our versions locked to the above mentioned ones, because nothing forces us to bump those versions but the lack of backward-compatibility prevents us until we figure out how to generate all we need (types, manifests/CRDs, clientsets) with the new toolset. The last thing forcing a version update was the lack of CEL-expression based validation, but bumping to a more but not too recent version solved that.

fmuyassarov: Ok, I will take care of that then. You can ignore my comment here.

askervin (approved):
Looks great!

@klihub, would you prefer to address my nits & supernits (to the extent you find reasonable) now or later? I'm fine merging this as is.

klihub:
> Looks great!
> 
> @klihub, would you prefer to address my nits & supernits (to the extent you find reasonable) now or later? I'm fine merging this as is.

Let me take a look at them and see how many I could address right now.

klihub:
> > Looks great!
> > @klihub, would you prefer to address my nits & supernits (to the extent you find reasonable) now or later? I'm fine merging this as is.
> 
> Let me take a look at them and see how many I could address right now.

@askervin I tried to address all your comments.

pkg/metrics/metrics.go
```diff
@@ -1,57 +1,705 @@
...
+		str = ""
+		sep = ""
+	)
+
+	if !(s.IsEnabled() || s.IsPolled()) {
+		str = "disabled"
+		sep = ","
+	} else {
+		if s.IsEnabled() {
+			str += "enabled"
```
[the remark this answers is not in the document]
klihub: Updated, and also changed the logic a bit to always dump all the state flags separately.

PR #404, opened 2024-11-11
[stacked 2/5] metrics: add de-facto standard collectors.

klihub:
Notes: This PR is *stacked on top* of #403.

Add a `metrics/collectors` subpackage to register the fairly standard `buildinfo`, `process` and `golang` runtime collectors. Turn on the `buildinfo` collector by default.


pfl (approved):
Apparently only the last commit was different from 403.

klihub:
> Apparently only the last commit was different from 403.

Yes, this is effectively a single-commit PR, it's just stacked on top of #403.

PR #409, opened 2024-11-12
build: enable race-detector in DEBUG=1 builds.

klihub:
When building with DEBUG=1, also use the '-race' golang build flag unless NORACE=1 is also set.


pfl (approved):
LGTM

askervin (approved):
Nice idea! LGTM.

PR #410, opened 2024-11-12
build: enable race-detector in image build, too.

klihub:


klihub:
My head's in Mississippi... apparently. I wonder how many force pushes it may take to get a 2-liner in 3 files right. I might have broken some kind of world record here...

fmuyassarov (approved):

PR #412, opened 2024-11-14
Migrate code-gen to kube_codegen.sh 

fmuyassarov:
Migrate to [kube_codegen.sh](https://github.com/kubernetes/code-generator/blob/master/kube_codegen.sh)  and bump k8s libraries along the way to the same version as `k8s.io/code-generator`.

I decided to wait for https://github.com/containers/nri-plugins/pull/406 to land but it should not block you from giving review on this.


klihub (changes requested):
Could you alter the split into two commits like this:

1. import new deps + update Makefile+any other tooling necessary
2. regenerate everything (`make generate`) and commit the resulting changes

It is much easier for us to review (review first commit, skim through resulting changes in the second), much easier for you to rebase (just drop the tip-most commit, rerun `make generate` the commit the changes) and much easier for me to test it on top of my tree (just cherry-pick for first commit with the tooling changes and run `make generate`). With the current version, the last two of these will cause unavoidable conflicts because the relevant changes are not split from the side-effects.

fmuyassarov:
> Could you alter the split into two commits like this:
> 
>     1. import new deps + update Makefile+any other tooling necessary
> 
>     2. regenerate everything (`make generate`) and commit the resulting changes
> 
> 
> It is much easier for us to review (review first commit, skim through resulting changes in the second), much easier for you to rebase (just drop the tip-most commit, rerun `make generate` the commit the changes) and much easier for me to test it on top of my tree (just cherry-pick for first commit with the tooling changes and run `make generate`). With the current version, the last two of these will cause unavoidable conflicts because the relevant changes are not split from the side-effects.

Sure, I will do as you suggested here because I currently it is messed up :)

fmuyassarov:
@klihub PTAL. Updated the commits.

klihub:
@fmuyassarov When I try a `make generate` with this, I get this failure:

```
klitkey1-mobl1 generators-v2 $ make generate
/home/kli/src/work/devel/nri-plugins/metrics/generators-v2/build-aux/bin/controller-gen \
        crd:generateEmbeddedObjectMeta=true output:crd:artifacts:config=config/crd/bases\
        paths="./pkg/apis/..."
/home/kli/src/work/devel/nri-plugins/metrics/generators-v2/build-aux/bin/controller-gen \
        rbac:roleName=manager-role crd paths="./pkg/apis/..." output:crd:artifacts:config=config/crd/bases
/home/kli/src/work/devel/nri-plugins/metrics/generators-v2/build-aux/bin/controller-gen object:headerFile="scripts/hack/boilerplate.go.txt" paths="./pkg/apis/..."
./scripts/hack/update_codegen.sh go
Generating client code for 1 targets
F1115 11:08:28.607834 2185078 main.go:65] Error: failed making a parser: error(s) in "/u/src/kli/work/devel/nri-plugins/metrics/generators-v2/pkg/apis/config/v1alpha1":
-: directory /u/src/kli/work/devel/nri-plugins/metrics/generators-v2/pkg/apis/config/v1alpha1 outside main module or its selected dependencies
Cleaning up temporary vendor directory...
make: *** [Makefile:505: generate-clients] Error 255
```

klihub:
@fmuyassarov I still get the same failure with this latest version.

scripts/hack/update_codegen.sh
```diff
@@ -0,0 +1,33 @@
...
+"${GO_CMD}" mod vendor
+mv vendor "${TEMP_VENDOR_DIR}"
+
+CODEGEN_PKG=$(cd "${TEMP_VENDOR_DIR}"; "${GO_CMD}" list -m -mod=readonly -f "{{.Dir}}" k8s.io/code-generator)
+popd > /dev/null
+
+# shellcheck source=/dev/null
+source "${CODEGEN_PKG}/kube_codegen.sh"
+
+kube::codegen::gen_client \
```
klihub, line 29:
```suggestion
source "${CODEGEN_PKG}/kube_codegen.sh"

cd ${REPO_ROOT}/pkg/apis

kube::codegen::gen_client \
```

klihub (changes requested):
Can you test on your side with my proposed one-liner change ?

With that in place now `make generate` works for me. I wonder if this then breaks for you, since the original version worked with $REPO_ROOT as the current working directory...

scripts/hack/update_codegen.sh
fmuyassarov: no it doesn't break.

klihub: Really weird voodoo, but let's go with that then.

Makefile
```diff
@@ -474,62 +474,35 @@ LOCALBIN ?= $(BUILD_AUX)/bin
...
+CONTROLLER_GEN_VERSION := v0.16.5
+
 ## External generation tool binaries
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 CLIENT_GEN ?= $(LOCALBIN)/client-gen
 
-GIT_CLONE    := git clone
+# client generation rules
+.PHONY: controller-gen
+controller-gen: ## install controller-gen
```
klihub, line 487: Nit: the ##-suffix does not make sense for us, because we lack the corresponding `make help` rule which would grep those out from the Makefile and produce a nice help page. Let's take that out, let's add the help-rule and then annotate the top-level targets with the help messages, all in part of a single PR.

Makefile
```diff
@@ -474,62 +474,35 @@ LOCALBIN ?= $(BUILD_AUX)/bin
...
+	@GOBIN=$(LOCALBIN) $(GO_CMD) install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION)
 
-# External generation tool versions
-CONTROLLER_TOOLS_VERSION ?= v0.16.5
-CLIENT_GEN_VERSION ?= v0.28.1
-CODE_GENERATOR_REPO ?= https://github.com/kubernetes/code-generator
-CODE_GENERATOR_VERSION ?= v0.28.1
-GENERATE_GROUPS ?= $(BUILD_AUX)/code-generator/generate-groups.sh
+.PHONY: code-generator
+code-generator: ## install client-gen
```
klihub, line 491: nit: ditto for this...

Makefile
```diff
@@ -474,62 +474,35 @@ LOCALBIN ?= $(BUILD_AUX)/bin
...
-.PHONY: generate-manifests
-generate-manifests: controller-gen
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd paths="./pkg/apis/..." output:crd:artifacts:config=$(CRD_BASE_DIR)
-
-# code generation rules for CRD types (DeepCopy, DeepCopyInto, etc.)
-.PHONY: generate-types
-generate-types: controller-gen
-	$(CONTROLLER_GEN) object:headerFile="./docs/license-header.go.txt" paths="./pkg/apis/..."
+.PHONY: manifests
+manifests: controller-gen ## generate RBAC and CRD objects.
```
klihub, line 495: nit: ditto, let's ditch the '##'-help message for the time being.

scripts/hack/update_codegen.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -e
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+trap cleanup EXIT
+GO_CMD=${1:-go}
```
klihub, line 8: nit: shouldn't this ideally be "${1:-go}", which would also keep this more consistent.

scripts/hack/update_codegen.sh
```diff
@@ -0,0 +1,35 @@
...
+BUILD_AUX=${2:-"${REPO_ROOT}/build-aux"}
```
klihub, line 9: nit: shouldn't this be ideally `"${2:-${REPO_ROOT}/build-aux}" ?

scripts/hack/update_codegen.sh
```diff
@@ -0,0 +1,35 @@
...
+"${GO_CMD}" mod vendor
+mv vendor "${TEMP_VENDOR_DIR}"
+
+CODEGEN_PKG=$(cd "${TEMP_VENDOR_DIR}"; "${GO_CMD}" list -m -mod=readonly -f "{{.Dir}}" k8s.io/code-generator)
+popd > /dev/null
+
+# shellcheck source=/dev/null
+source "${CODEGEN_PKG}/kube_codegen.sh"
+
+cd ${REPO_ROOT}/pkg/apis
```
klihub, line 29: nit: ideally cd "${REPO_ROOT}/pkg/apis"

klihub (commented):
This really LGTM now. I have a few small nits. Some of the shell-quoting ones should be taken with a grain of salt though, especially if shellcheck disagrees with me.

marquiz:
I think this could be merged. Preferably with klihubs comments addressed.

Now that I looked at this: just a general note that my personal feeling is that the Makefile is getting somewhat hairy'ish with all the phony rules and army of variables (like `CLIENT_GEN_VERSION` has no effect if `CLIENT_GEN` is modified and such). I'd find it easier to follow a script (á la `scripts/hack/update_codegen.sh`) that just does everything (installs tools, runs them, cleanup if needed) in sequential order. But not much to do with this PR...

fmuyassarov:
> I think this could be merged. Preferably with klihubs comments addressed.
> 
> Now that I looked at this: just a general note that my personal feeling is that the Makefile is getting somewhat hairy'ish with all the phony rules and army of variables (like `CLIENT_GEN_VERSION` has no effect if `CLIENT_GEN` is modified and such). I'd find it easier to follow a script (á la `scripts/hack/update_codegen.sh`) that just does everything (installs tools, runs them, cleanup if needed) in sequential order. But not much to do with this PR...

This a valid point but I would prefer to do that clean up as a follow up.

marquiz:
> This a valid point but I would prefer to do that clean up as a follow up.

Yeah, was just a note, food for thought... And I'm not sure that everybody agrees...

marquiz (approved):

fmuyassarov:
rebased....

fmuyassarov:
@klihub This is now rebased on top of your merged patches.

klihub:
> > This a valid point but I would prefer to do that clean up as a follow up.
> 
> Yeah, was just a note, food for thought... And I'm not sure that everybody agrees...

Yes, as long as we think it through beforehand and make such changes so that 1) we can still pass the tooling version(s) to be used from the Makefile rule to the script, and 2) the script is smart enough to not re-clone and re-compile the tooling binaries if I run twice `make generate` without overriding the tooling versions.

But maybe I'd prefer to let things settle now for a while and go with these changes before we start again touching this stuff...

klihub (approved):

PR #413, opened 2024-11-17
fixes: copy assigned memory zone in grant clone.

klihub:
When cloning a grant, copy also the assigned memory zone. Failing to do so can result in errors like this during reconfiguration:

```
E: failed to reinstate grants verbatim: topology-aware: failed to get libmem offer for pool "NUMA node #1", grant of default/test/c2: libmem: invalid NodeMask: request without affinity
```

Additionally, also include the container's pretty name in saved grants, just to help post-mortem interpretation.


cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -993,6 +993,8 @@ func (cg *grant) Clone() Grant {
 		cpuType:    cg.CPUType(),
```
askervin, line 993: Would it be good to copy PrettyName, too?

PR #416, opened 2024-11-22
resmgr,agent: propagate startup config error back to CR.

klihub:
Update the agent's config notification callback signature and the resmgr notification callaback to propagate errors for the initial configuration back to the source CR before exiting.


pfl (approved):

fmuyassarov (approved):
lgtm

PR #417, opened 2024-11-30
agent, resmgr: add support for topology hints based on Pod Resource API

klihub:
Notes: This PR is currently marked a draft, as it is still missing updates to the documentation...

This patch series add support for generating topology hints based on pod- and container-specific resource data acquired from the kubelet's Pod Resource API. In particular this PR

- adds a low-level Pod Resource API client to the agent
- adds agent configuration snippets to config CRs to allow enabling/disabling Pod Resource API support
- exposes synchronous and asynchronous Pod Resource queries via the agent to the resource manager
- if enabled, fetches pod resource data during initial sync and pod creation
- if enabled, generates extra topology hints based on pod resource container data


klihub:
Closing in favor of equivalent split/stacked PR #418, #419, #420.

PR #418, opened 2024-12-01
[podresapi 1/3] agent,config,helm: make agent runtime configurable.

klihub:
This PR implements runtime configuration for the agent itself, in preparation for allowing configurable access to the kubelet's Pod Resource API. Currently the only existing control enables access to Node Resource Topology Custom Resources and it is on by default. Update the agent to dig out and honor its own configuration when configuration is updated.


pkg/agent/agent.go
```diff
@@ -262,42 +262,87 @@ func (a *Agent) Stop() {
...
+		if err != nil {
+			return err
+		}
+		cli, err := nrtapi.NewForConfigAndClient(cfg, a.httpCli)
+		if err != nil {
+			return fmt.Errorf("failed to setup NRT client: %w", err)
+		}
+		a.nrtCli = cli
+
+	case !cfg.NodeResourceTopology && a.nrtCli != nil:
```
kad, line 327: if / else ?

kad (approved):
lgtm

PR #419, opened 2024-12-01
[podresapi 2/3] resmgr,agent: generate topology hints from Pod Resource API.

klihub:
This patch series
- adds support to query the kubelet's Pod Resource API via the agent
- generates extra topology hints using the Pod Resource API, if it is enabled
- updates Helm charts to provide access to the kubelet's Pod Resource API socket

Pod Resource API queries are triggered asynchronously during pod creation/handling `RunPodSandbox` NRI events. Container creation waits for any pending queries (with a short timeout) then generates extra topology hints if the results contain devices with declared NUMA-locality. The existing device allow-/deny-annotations can be used to explicitly allow or deny topology hints corresponding to Pod Resource API hints.


pkg/agent/podresapi/client.go
```diff
@@ -0,0 +1,165 @@
...
+
+		c.noGet = true
+	}
+
+	l, err := c.List(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get pod resources: %w", err)
+	}
+
+	return l.GetPodResources(namespace, pod), nil
```
kad, line 164: should we cache whole reply? not only for single pod if we got the whole list....

pkg/agent/agent.go
```diff
@@ -262,42 +264,101 @@ func (a *Agent) Stop() {
...
+		if err != nil {
+			return fmt.Errorf("failed to setup kubernetes config resource client: %w", err)
+		}
 	}
 
-	a.httpCli, err = rest.HTTPClientFor(cfg)
-	if err != nil {
-		return fmt.Errorf("failed to setup kubernetes HTTP client: %w", err)
+	switch {
+	case cfg.PodResourceAPI && a.podResCli == nil:
```
kad, line 317: if / else ?

pkg/agent/podresapi/client.go
klihub: Yes, I was considering that, but decided to save that for another PR.

klihub: Also I'm not sure if the current podresapi handling during CreateContainer is guaranteed to be always correct... or if is only guaranteed to be correct for the first creation of a named container within the pod. IOW, I'm not sure if we'd need to refetch the pod resources when an already formerly existing container exits/crashes in a pod and gets recreated...

kad: To best of my memory, this data is populated from devices/resources allocated at pod admission time. those are not re-allocated on container restarts. So, in theory, if we got info about the pod, it should be complete. I don't know if kubelet handles properly race conditions if pod is in the middle of admission at the time of List() call. (As admission is pod level thingy, I suspect it shouldn't be an issue).
But I agree, we can leave it as separate PR.

pkg/agent/agent.go
kad: well, maybe some comments might be added then, because for me this switch was less readable, until I figured out that it is also handling re-configure scenario where option can be enabled/disabled dynamically.

pkg/agent/podresapi/client.go
```diff
@@ -0,0 +1,167 @@
...
+	conn       *grpc.ClientConn
+	cli        api.PodResourcesListerClient
+	noGet      bool
+}
+
+const (
+	// these constants were obtained from NFD sources, cross-checked against
+	//   https://github.com/kubernetes/kubernetes/blob/release-1.31/test/e2e_node/util.go#L83
+	defaultSocketPath = "/var/lib/kubelet/pod-resources/kubelet.sock"
+	timeout           = 10 * time.Second
```
kad, line 46: shall we reduce timeout in future? it might be too long for our operations.

kad (approved):
lgtm

pkg/agent/podresapi/client.go
klihub: This timeout is now only used for establishing a client connection. For actually listing all pod resources or querying a single one the caller provides `time.Duration` timeout (in the agent interface) or a `context.Context` which can then be a `context.WithTimeout` to use a timeout for the call.

Maybe that const should be/have been called a `connectionTimeout` ?

klihub: Asynchronous List()s/Get()s we fire off with a 2 and 1 second timeouts respectively. However (these could also much longer), since the async. functions return a channel that can be used to wait for and eventually obtain the result, it is possible to programmatically override with a simple select with a channel receive and a time.After receive cases.

PR #420, opened 2024-12-01
[podresapi 3/3] topology-aware: use Pod Resource API hints if present.

klihub:
This patch series adds handling for extra topology hints acquired from the kubelet's Pod Resource API. If any such hints are found, they are fully resolved (from NUMA-only hints to cpuset-specific ones), before taken into account.


cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -831,6 +831,7 @@ func (cs *supply) GetScore(req Request) Score {
 
 	// calculate real hint scores
 	hints := cr.container.GetTopologyHints()
+	hints.ResolvePartialHints(cs.GetNode().System().NodeHintToCPUs)
```
kad, line 834: I wonder, how the hints were working for nvme storage? As I recall our GetTopologyHint function was able to return for those only node number, without cpulist. but I might recall inctorrectly.

klihub: Because when the package constructs the hints itself using data from sysfs, then whenever it encounters such a partial hint in the sysfs tree, it tries to resolve it by walking up the parent directories looking for a `local_cpulist` and that usually works. Here a podresapi based hint is  externally constructed using the only topology information podresapi provides: one or more NUMA node ID, which we then need to resolve to CPUs.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -486,6 +486,16 @@ metadata:
         - /xy-zy/another-blacklisted-path5
 ```
 
+## Using Pod Resource API for Extra Topology Hints
+
+If access to the `kubelet`'s `Pod Resource API` is enabled in the
+[Node Agent's](../developers-guide/architecture.md#node-agent) configuration,
+it is automatically used to generate per-container topology hints when a
+device with locality to a NUMA node is advertised by the API. Annotated allow
+and deny lists can be used to selective disable or enable per-resource hints,
```
fmuyassarov, line 495:
```suggestion
and deny lists can be used to selectively disable or enable per-resource hints,
```

pkg/agent/agent.go
```diff
@@ -556,6 +625,10 @@ func (a *Agent) updateConfig(cfg metav1.Object) {
 	}
 
 	a.currentCfg = cfg
+
+	if err := a.configure(cfg); err != nil {
+		log.Warn("failed to reconfigure agent: %v", err)
+	}
```
fmuyassarov, line 631: Isn't this unreachable code?
[setupClinets()](https://github.com/containers/nri-plugins/blob/72bfdfcf4cf0998449e327892075a2d570ae82c5/pkg/agent/agent.go#L183C11-L183C23) returns an error within the `Start()` before we reach `updateNodeConfig()`.

pkg/agent/pod-resource-api.go
```diff
@@ -0,0 +1,80 @@
...
+
+// GetPodResources queries the given pod's resources.
+func (a *Agent) GetPodResources(ns, pod string, timeout time.Duration) (*podresapi.PodResources, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	return a.podResCli.Get(ctx, ns, pod)
+}
+
+// GetPodResources queries the given pod's resources asynchronously.
```
fmuyassarov, line 32:
```suggestion
// GoGetPodResources queries the given pod's resources asynchronously.
```

pkg/agent/pod-resource-api.go
```diff
@@ -0,0 +1,80 @@
...
+
+// ListPodResources lists all pods' resources.
+func (a *Agent) ListPodResources(timeout time.Duration) (podresapi.PodResourcesList, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	return a.podResCli.List(ctx)
+}
+
+// ListPodResources lists all pods' resources asynchronously.
```
fmuyassarov, line 61:
```suggestion
// GoListPodResources lists all pods' resources asynchronously.
```

fmuyassarov (commented):
Since NodeResourceTopology can be disabled as well, should we extend this [line](https://github.com/containers/nri-plugins/blob/3b69120482eb8a24604d7f7198677e5e4cab79ef/pkg/resmgr/resource-manager.go#L276C23-L276C39) to check if the `NodeResourceTopology=true`?

docs/resource-policy/policy/topology-aware.md
klihub: Fixed.

pkg/agent/pod-resource-api.go
klihub: Fixed.

pkg/agent/pod-resource-api.go
klihub: Fixed.

klihub:
> Since NodeResourceTopology can be disabled as well, should we extend this [line](https://github.com/containers/nri-plugins/blob/3b69120482eb8a24604d7f7198677e5e4cab79ef/pkg/resmgr/resource-manager.go#L276C23-L276C39) to check if the `NodeResourceTopology=true`?

Let's leave that for another PR. There are a few other similar things we could then take care of in the same PR, for instance not updating policy metrics if their collection is not enabled.

pkg/apis/config/v1alpha1/types.go
```diff
@@ -78,6 +81,9 @@ type BalloonsPolicySpec struct {
 	Log log.Config `json:"log,omitempty"`
 	// +optional
 	Instrumentation instrumentation.Config `json:"instrumentation,omitempty"`
+	// +optional
+	// +kubebuilder:default={"nodeResourceTopology": true }
+	Agent AgentConfig `json:"agent,omitempty"`
```
fmuyassarov, line 86: Wouldn't adding the policy-agnostic agent configuration under each policy CR lead to conflicts when multiple policies attempt to set different values for the same fields, such as `nodeResourceTopology`?

kad (approved):
lgtm

pkg/apis/config/v1alpha1/types.go
klihub: I'm not sure what you mean. This newly added agent-specific configuration value configures the behavior of the agent (wrt. whether the the Node Resource Topology and Pod Resource API functionality is made available) for each policy. It is similar to the other policy-agnostic configuration bits, for instance instrumentation. Everything in there is purely an input there for the resource manager and other infra bits, policies never set there anything. Multiple policies cannot conflict, because they have the same piece of data in one of their own policy-specific configuration CR, in some instance of `ballonspolicies.config.nri`, `topologyawarepolicies.config.nri` or `templatepolicies.config.nri`.

fmuyassarov: What i meant is that you are allowing the agent configuration to be modified by all all the policies (well not all but more than one). Meaning, when more than one policy is installed a user can disable NRT from one policy and enable from the other. How do you handle that?

klihub: Sure, but that is deliberately done so by design. I mean... how is this different from, for instance, instrumentation being enabled for one policy but not for another ?

klihub: > Meaning, when more than one policy is installed, a user can disable NRT from one policy and enable from the other. How do you handle that?

Well, those policy instances for which NRT is disabled in the agent will not be exporting NRT CRs, and those instance for which NRT is enabled in the agent will be exporting NRT CRs.

fmuyassarov: Sorry, I had misunderstanding of the agent instances. We can close this chat.

PR #421, opened 2024-12-02
docs: fix build error, avoid testdata scan infinite loop.

klihub:
This PR
- bumps the golang toolchain used when building docs
  - fixes toolchain complaining about patch-level `go` declaration in `go.mod`
- adds an ignore rule to avoid for extracted sample sysfs test data
  - fixes `sphinx-build` falling into an infinite loop while scanning the source tree for docs
- bumps the sphinx toolchain while we're at it


docs/resource-policy/policy/topology-aware.md
```diff
@@ -443,7 +443,7 @@ metadata:
     prefer-reserved-cpus.resource-policy.nri.io/container.special: "false"
 ```
 
-## Allowing or denying mount/device paths via annotations
+## Controlling Topology Hints Via Annotations
```
kad, line 447: this probably should not be part of this PR

klihub: Argh, true.

kad (approved):
LGTM

fmuyassarov (approved):

PR #423, opened 2024-12-05
agent,resmgr: merge PodResources{List,Map}, cache last List() result.

klihub:
Merge `PodResourceMap` into `PodResourcesList`. Avoid unnecessary re-listing of pod resources by caching the result of the last `List()` and trying to reuse it to look up resources.

### DIFF
--- a/pkg/agent/pod-resource-api.go
+++ b/pkg/agent/pod-resource-api.go
@@ -51,7 +51,7 @@ func (a *Agent) GoGetPodResources(ns, pod string, timeout time.Duration) <-chan 
 }
 
 // ListPodResources lists all pods' resources.
-func (a *Agent) ListPodResources(timeout time.Duration) (podresapi.PodResourcesList, error) {
+func (a *Agent) ListPodResources(timeout time.Duration) (*podresapi.PodResourcesList, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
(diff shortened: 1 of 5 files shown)


askervin (commented):
Do you see need and natural point for cache invalidation? Now we are caching pod resources based on namespace + pod name, if I'm reading this correctly. Should we clear the cache on removing pod sandbox, or perhaps in some other events that could make us suspicious that pod has changed....

klihub:
> Do you see need and natural point for cache invalidation? Now we are caching pod resources based on namespace + pod name, if I'm reading this correctly. Should we clear the cache on removing pod sandbox, or perhaps in some other events that could make us suspicious that pod has changed....

In the current version we tried to implement (and use) this so that both caching and invalidation would be transparent/implicit, with the trigger for invalidation being if we fail to find a pod in the cached last list result. To be honest, I am not sure if this is enough, we'll need to dig in more to understand better.

If we use the pod resource API at all, we use it to generate extra topology hints for device/memory+CPU alignment. My bigger fear than inaccuracy was the extra cost, especially in increased latency for decision making. That is why if enabled, we now we trigger an asynchronous pod resource API lookup during pod creation and use the result during container creation, if it is available by that time.

It might be that we'd need also explicit invalidation, which we could then trigger, for instance, whenever a container is stopped. But we'll need to dig in more to understand it better.

askervin:
> It might be that we'd need also explicit invalidation, which we could then trigger, for instance, whenever a container is stopped. But we'll need to dig in more to understand it better.

As this feature is behind an off-by-default option, I suppose we could start with it, experiment, and consider explicit invalidation in a separate PR.

askervin (approved):
LGTM

PR #425, opened 2024-12-10
agent: golangci-lint fixes.

klihub:
Fix golangci-lint issues.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -212,10 +212,10 @@ func (a *Agent) Start(notifyFn NotifyFn) error {
 				break
 			}
 			if e.Type == watch.Added || e.Type == watch.Modified {
-				group, _ := e.Object.(*corev1.Node).Labels[a.groupLabel]
+				group := e.Object.(*corev1.Node).Labels[a.groupLabel]
 				if group == "" {
 					for _, l := range deprecatedGroupLabels {
-						group, _ = e.Object.(*corev1.Node).Labels[l]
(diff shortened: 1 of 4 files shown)


askervin (approved):
LGTM.

My only concern here is dropping connection timeout. It would definitely be a corner case that the socket exists but kubelet is not responding. I'm not too familiar with grpc, so I don't know if this could lead into a situation where our plugins that use podresapi would stop responding, too, and get kicked out from container runtime.

@klihub, do you think this can cause issues, or do you think plugins will recover just fine when kubelet is alive again?

klihub:
> LGTM.
> 
> My only concern here is dropping connection timeout. It would definitely be a corner case that the socket exists but kubelet is not responding. I'm not too familiar with grpc, so I don't know if this could lead into a situation where our plugins that use podresapi would stop responding, too, and get kicked out from container runtime.
> 
> @klihub, do you think this can cause issues, or do you think plugins will recover just fine when kubelet is alive again?

Dropping (connection establishment) timeout and the other changes were made based on the gRPC documentation and the API functions we stopped using being marked deprecated in gRPC. That's why for instance `WithBlock()` had to be dropped.

I haven't checked the code but I think what happens is that an asynchronous connection establishment in the background is triggered by `grpc.NewClient()` and it returns immediately, without blocking. This was already the default behavior without the extra `WithBlock()` option. Then if a request is coming while connection establishment is still going on, the timeout for that request will kick in if connection establishment is too slow.

PR #437, opened 2024-12-10
log: golangci-lint fixes.

klihub:
Fix golangci-lint issues.


pkg/log/signal.go
```diff
@@ -35,12 +35,7 @@ func SetupDebugToggleSignal(sig os.Signal) {
 	go func(sig <-chan os.Signal) {
 		state := map[bool]string{false: "off", true: "on"}
 		for {
-			select {
-			case _, ok := <-sig:
-				if !ok {
-					return
-				}
-			}
+			<-sig
```
askervin, line 38: Ok... it seems to be really weird if the channel would be closed, so we don't need to react on `!ok`

askervin (approved):
LGTM

PR #446, opened 2024-12-10
scripts: clean up all tests VMs after test.

klihub:
Update test VM directory glob pattern to `test/e2e/n[0-9]*-c*`. This should also catch and clean up the latest VM topology addition (n6-hbm-cxl-fedora-40-cloud-base-containerd) which is now left lingering after nightly test runs.


askervin (approved):
LGTM

PR #449, opened 2024-12-12
pkg/udev: implement udev event reading/monitoring.

klihub:
Implement low-level udev raw reader, event reader and a monitor, with support for property-based event filtering. This should provide enough plumbing to implement memory hotplug/unplug detection.

### DIFF
--- a/pkg/udev/monitor.go
+++ b/pkg/udev/monitor.go
@@ -1,0 +1,164 @@
+// Copyright The NRI Plugins Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package udev
+
+import (
+	"fmt"
+	"path"
+)
+
+// MonitorOption is an opaque option which can be applied to a Monitor.
+type MonitorOption func(*Monitor)
+
+// WithFilters returns a MonitorOption for filtering events by properties.
+// Properties within a map have AND semantics: the map matches an event if
+// all key-value pairs match the event. Multiple maps have OR semantics:
+// they match an event if at least one map matches the event. Events which
+// are matched are passed through. Others are filtered out.
+func WithFilters(filters ...map[string]string) MonitorOption {
+	return func(m *Monitor) {
+		m.filters = append(m.filters, filters...)
+	}
+}
+
+// WithGlobFilters returns a MonitorOption for filtering events by properties.
+// Semantics are similar to WithFilters, but properties are matched using glob
+// patterns instead of verbatim comparison.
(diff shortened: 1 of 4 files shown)


pkg/udev/reader.go
```diff
@@ -0,0 +1,183 @@
...
+	e := &Event{
+		Properties: map[string]string{},
+	}
+
+	hdr, err := r.b.ReadString(0)
+	if err != nil {
+		return nil, err
+	}
+
+	e.Header = hdr
```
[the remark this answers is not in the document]
klihub: Yes, and maybe we should not blindly trust `len(kv[1]) > 0` after `SplitN()` below either...

pkg/udev/reader.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil, fmt.Errorf("failed to bind udev reader: %w", err)
+	}
+
+	return &Reader{sock: fd}, nil
+}
+
+// Read implements the io.Reader interface.
+func (r *Reader) Read(p []byte) (int, error) {
+	n, err := syscall.Read(r.sock, p)
+	if r.closed {
```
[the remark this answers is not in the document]
klihub: No, there is no reason. Fixed.

pkg/udev/monitor.go
```diff
@@ -0,0 +1,161 @@
...
+		m.globbers = append(m.globbers, globbers...)
+	}
+}
+
+// Monitor monitors udev events.
+type Monitor struct {
+	r        *EventReader
+	filters  []map[string]string
+	globbers []map[string]string
+	stopCh   chan struct{}
```
marquiz, line 50: What is this used for?

pkg/udev/monitor.go
```diff
@@ -0,0 +1,161 @@
...
+	for evt := range events {
+		if !m.filter(evt) {
+			continue
+		}
+
+		select {
+		case filtered <- evt:
+			stuck = false
+		default:
+			if !stuck {
```
marquiz, line 114: So we don't want to flood the log with all dropped events? Just wondered would it be good in this case to get a log message when we get back to the non-stuck state?

pkg/udev/monitor.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return m, nil
+}
+
+// Start starts event monitoring and delivery.
+func (m *Monitor) Start(events chan *Event) {
+	if len(m.filters) == 0 && len(m.globbers) == 0 {
+		go m.reader(events)
+	} else {
+		filter := make(chan *Event, 64)
```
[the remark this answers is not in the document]
klihub: True. Updated to `unfiltered` here and in `Monitor.filterer()` as well.

pkg/udev/monitor.go
klihub: Unused now. Thanks for spotting it!

pkg/udev/monitor.go
klihub: Yes, the idea is to do edge-triggered logging of transitions, not level-triggered flooding of logs. Added logging to the other transition as well.

marquiz (approved):
Haven't tested, but looks good to me :)

scripts/udev-monitor/udev-monitor.go
```diff
@@ -0,0 +1,67 @@
+package main
```
marquiz, line 1: License header?

marquiz (approved):
LGTM, if you ask me

askervin (approved):
LGTM!

PR #450, opened 2024-12-13
.github: add golangci-lint, split PR verification into multiple jobs.

klihub:
This patch set adds a golangci-lint worklflow. Additionally, it splits PR verification into multiple parallel jobs.


fmuyassarov (approved):
LGTM.
Is there any reason to split out the golangci-lint into its own workflow file?

.github/workflows/verify.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - main
+      - "release-*"
+      - "test/**"
+  pull_request:
+    branches:
+      - main
+      - "release-*"
+
+env:
+  GO_VERSION: "1.22.1"
```
kad, line 15: should it be 1.23.x? or latest 1.22.10?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - main
+      - "release-*"
+      - "test/**"
+  pull_request:
+    branches:
+      - main
+      - "release-*"
+
+env:
+  GO_VERSION: "1.22.1"
```
kad, line 15: same, should it be 1.23.x or latest 1.22.10?

kad (commented):
generally ok, but minor question about golang versions.

klihub:
> LGTM. Is there any reason to split out the golangci-lint into its own workflow file?

Well, nothing too specific, really. I wanted for a while now to get rid of that 'project checks' workflow, and split it up to smaller, more granular jobs. And, well... there is never a better time than right now. About golangci-lint, since @kad mentioned the other day that he might want to enable other linters too, I decided to add the new golangci-lint job readily to a new lint workflow.

.github/workflows/verify.yaml
klihub: I can file another PR for that.

.github/workflows/verify.yaml
```diff
@@ -0,0 +1,94 @@
...
+          make verify-godeps
+
+  verify-formatting:
+    name: Formatting
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
klihub: We really do need it in two places: git-diff --check white-space error checking (technically wouldn't need to fetch all, just enough to get base_ref), and for verifying/generating docs. For binary building it could be dropped, it would just result in a different built version than what you'd get locally (since git describe would produce a different result).

klihub:
> > > LGTM. Is there any reason to split out the golangci-lint into its own workflow file?
> > 
> > 
> > Well, nothing too specific, really. I wanted for a while now to get rid of that 'project checks' workflow, and split it up to smaller, more granular jobs. And, well... there is never a better time than right now. About golangci-lint, since @kad mentioned the other day that he might want to enable other linters too, I decided to add the new golangci-lint job readily to a new lint workflow.
> 
> maybe we can then add it to the same workflow since there is no separation in terms of when the job should be executed. (Not a blocker though, feel free to ignore it).

I prefer to keep it separate. I want to see it as `Lint / golangci-lint` and not as `Verify / golangci-lint` in the check summary...

.github/workflows/verify.yaml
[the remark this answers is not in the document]
askervin: Thanks for the clarification, @klihub!

askervin (approved):

PR #451, opened 2024-12-16
Add balloon type specific pinMemory option

askervin:
This option enables overriding policy level `pinMemory` option in balloon types.

Use cases:
1. Enable pinning memory of latency critical container to the lowest latency nodes while leaving all other containers completely unpinned.
2. Allow certain containers to access all memory nodes while memory of all other containers would be pinned.

### DIFF
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -1411,7 +1411,7 @@ func (p *balloons) updatePinning(blns ...*Balloon) {
 				} else {
 					allowedCpus = pinnableCpus
 				}
-				p.pinCpuMem(c, allowedCpus, bln.Mems, bln.memTypeMask)
+				p.pinCpuMem(c, allowedCpus, bln.Mems, bln.memTypeMask, bln.Def.PinMemory)
 			}
 		}
 	}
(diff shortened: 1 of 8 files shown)


cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1526,7 +1526,13 @@ func (p *balloons) pinCpuMem(c cache.Container, cpus cpuset.CPUSet, mems idset.I
 			c.SetCPUShares(int64(cache.MilliCPUToShares(int64(mCpu))))
 		}
 	}
-	if p.bpoptions.PinMemory == nil || *p.bpoptions.PinMemory {
+	// Start from policy-level PinMemory...
+	pinMemory := p.bpoptions.PinMemory == nil || *p.bpoptions.PinMemory
+	// ...and allow override in balloon-type-level PinMemory
+	if (blnDefPinMemory != nil) {
```
klihub, line 1532: nit: have been writing C code lately, now haven't we... or then just good ol' habits die really hard...

klihub (commented):
Looks fine, apart from a formatting error which needs fixing. Also, I think we should try to keep libmem in the loop about total memory usage. The container either might have some pinning already calculated by some other component or it will be pinned to all memory nodes. In either case, its max. memory consumption would need to be accounted where it is going to be pinned, even if that pinning has not been selected by us. But it is perfectly fine to address that in another PR.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1526,7 +1526,13 @@ func (p *balloons) pinCpuMem(c cache.Container, cpus cpuset.CPUSet, mems idset.I
...
+		pinMemory = *blnDefPinMemory
+	}
+	if pinMemory {
```
klihub, line 1535: I think we should let also libmem know about the new container consuming memory, even if we don't ask libmem to determine the pinning according to its own taste. Otherwise it cannot even in theory determine correctly when a node is in danger of running out of memory.

klihub (approved):
LGTM.

kad (approved):
lgtm

PR #452, opened 2024-12-17
topology-aware: fix mockCache.RefresPods.

klihub:
This PR fixes a problem that crept in via #423, which was branched off before #450 and was not rebased once that got merged.


askervin (approved):
LGTM

PR #459, opened 2024-12-20
golang-fixup: bump go version for memory-qos, too.

klihub:
Fixup commit with unsaved editor buffer snafu. Missed in the review of #458.


askervin (approved):
LGTM

PR #460, opened 2024-12-23
[1/2] OLM workflow: add automatic OLM bundle submission.

fmuyassarov:
GitHub bot account: https://github.com/nri-plugins-bot


.github/workflows/publish-olm-bundle.yaml
```diff
@@ -0,0 +1,58 @@
...
+          gpg_private_key: ${{ secrets.MACHINE_GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.MACHINE_PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+          workdir: community-operators
+
+      - name: Copy the bundle to the community-operators repo
+        run: |
+          mkdir -p operators/nri-plugins-operator/${{ github.ref_name }}
+          cp -r ../deployment/operator/bundle/* operators/nri-plugins-operator/${{ github.ref_name }}
```
klihub, line 42: Can it happen, that for some reason the operator bundle will have an entirely new file which is not part of the repo yet and we'd need to add it to the repo ?

fmuyassarov: No, if we don't bump the operator-sdk in the same PR, without testing it.

fmuyassarov: even if it adds, I think it will always follow the same path, meaning dump everything under `bundle/ `directory.

fmuyassarov: I believe the action will add that according to the statement on the action page

[`Commit all changes to a new branch, or update an existing pull request branch.`](https://github.com/peter-evans/create-pull-request).

klihub:
The other remaining thing is what we discussed offline today: adding a secondary trigger (maybe a push to a branch or a new tag matching a pattern) which would trigger the workload but with an eventual PR filed against a test repostory (which probably should the real community operators repo cloned under the bot account).

fmuyassarov:
> The other remaining thing is what we discussed offline today: adding a secondary trigger (maybe a push to a branch or a new tag matching a pattern) which would trigger the workload but with an eventual PR filed against a test repository (which probably should be the clone of the real community operators repo under the bot's account).

It won't be that easy to test even with extra conditional trigger. Because, I don't have permissions to push anything that could result in the workflow to start. However, I have tested the same workflow file with only one difference being the source repo set as fmuyassarov/nri-plugins (my fork) instead of containers/nri-plugins. And the result you can see here: https://github.com/k8s-operatorhub/community-operators/pull/5518.

.github/workflows/publish-olm-bundle.yaml
```diff
@@ -0,0 +1,79 @@
...
+          fetch-depth: 0
+
+      - name: Determine Repository Variable
+      # Remove this conditional logic once the workflow behavior
+      # is validated
+          run: |
+            if [[ "${{ github.event_name }}" == "issues" && "${{ github.event.action }}" == "opened" ]]; then
+              echo "repository=k8s-operatorhub/community-operators" >> $GITHUB_ENV
+            else
+              echo "repository=fmuyassarov/community-operators" >> $GITHUB_ENV
```
klihub, line 29: Isn't this condition the wrong way around ? If I read this correctly, we use the real repo if this workflow is triggered by a 'issue opened' event, otherwise we use your personal fork. Shouldn't that be the other way around ?

.github/workflows/publish-olm-bundle.yaml
```diff
@@ -0,0 +1,79 @@
...
+
```
klihub, line 30: Also, I think there is a missing terminating `fi` for the shell `if` here...

fmuyassarov: oh my bad

.github/workflows/publish-olm-bundle.yaml
fmuyassarov: thanks for catch.

kad:
Generally looks ok, just need a cleanup of all debugging bits, and hardcoded versions to make it mergable.
Second thing to invistigate: don't use primary gpg private key, it should be enough to store in secrets just a signing sub-key.

fmuyassarov:
> > Generally looks ok, just need a cleanup of all debugging bits, and hardcoded versions to make it mergable.
> 
> Those are added intentionally for the testing purpose. The idea is that we merge it as it is now and create an issue to test the workflow and if it behaves as it should there will be a follow up PR to clean up testing bits.

@kad  that's what we discussed with @klihub offline.

.github/workflows/publish-olm-bundle.yaml
```diff
@@ -0,0 +1,80 @@
...
+          mkdir -p community-operators/operators/nri-plugins-operator/v0.8.1
+          cp -r deployment/operator/bundle/ community-operators/operators/nri-plugins-operator/v0.8.1
+# mkdir -p community-operators/operators/nri-plugins-operator/${{ github.ref_name }}
+# cp -r deployment/operator/bundle/ community-operators/operators/nri-plugins-operator/${{ github.ref_name }}
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6
+        with:
+          signoff: true
+          author: nri-plugins-bot
```
klihub, line 64: I think this will fail with `The format of 'nri-plugins-bot' is not a valid email address with display name.`

klihub: I have a fix for it here: https://github.com/klihub/nri-plugins/commit/090e942eabdde4e5de67121f8122d0d421bc2795.

@fmuyassarov I also played around (maybe more like wasted a few hours) with some ideas of how we could keep the test-trigger around. It is here if you're interested: https://github.com/klihub/nri-plugins/commit/cff767c9e1c27439c4aed374d52e8b2acfaa3999

.github/workflows/publish-olm-bundle.yaml
```diff
@@ -0,0 +1,80 @@
...
+      # is validated
+          run: |
+            if [[ "${{ github.event_name }}" == "issues" && "${{ github.event.action }}" == "opened" ]]; then
+              echo "repository=fmuyassarov/community-operators" >> $GITHUB_ENV
+            else
+              echo "repository=k8s-operatorhub/community-operators" >> $GITHUB_ENV
+            fi
+
+      - name: Build the bundle
+        run: pushd deployment/operator && VERSION=0.8.1 make bundle && popd
```
askervin, line 33: Two questions from a complete github workflow n00b:

Would it be good to dig out VERSION from checked out repo? Like "echo nri_plugins_ver=$(git describe --tags --abbrev=0) >> $GITHUB_ENV" (similarly to conditional repository above)? Now we have hardcoded 0.8.1 in many locations in this yaml. Has there been some issue with `${{ github.ref_name }}`, or are they commented out just for testing?

What happens here if `make bundle` fails? Is it possible that the workflow continues to the next steps and creates a PR with invalid bundle content?

fmuyassarov: hardcoded versioning is only for testing purpose which will  be removed in a follow up PR after this gets merged and tested.
Eventually, it will look like this

```YAML
name: Submit release OLM bundle

on:
  release:
    types:
      - published

jobs:
  createPullRequest:
    name: Publish new OperatorHub release
    runs-on: ubuntu-22.04
    steps:
      - name: Checkout
        uses: actions/checkout@v4
        with:
          fetch-depth: 0

      - name: Build the bundle
        run: pushd deployment/operator && make bundle && popd

      - name: Checkout upstream community-operators repo
        uses: actions/checkout@v4
        with:
          repository: k8s-operatorhub/community-operators
          path: community-operators
          ref: main
          token: ${{ secrets.BOT_PAT }}

      - name: Import GPG key
        uses: crazy-max/ghaction-import-gpg@v6
        id: import-gpg
        with:
          gpg_private_key: ${{ secrets.BOT_GPG_PRIVATE_KEY }}
          passphrase: ${{ secrets.BOT_GPG_PASSPHRASE }}
          git_user_signingkey: true
          git_commit_gpgsign: true
          workdir: community-operators

      - name: Copy the bundle to the community-operators repo
        run: |
          mkdir -p community-operators/operators/nri-plugins-operator/${{ github.ref_name }}
          cp -r deployment/operator/bundle/ community-operators/operators/nri-plugins-operator/${{ github.ref_name }}

      - name: Create Pull Request
        uses: peter-evans/create-pull-request@v6
        with:
          signoff: true
          author: "NRI Plugins Bot <${{ steps.import-gpg.outputs.email }}>"
          committer: "Github Actions <${{ steps.import-gpg.outputs.email }}>"
          add-paths: |
            operators/nri-plugins-operator/**
          path: community-operators
          push-to-fork: nri-plugins-bot/community-operators
          branch: olm-${{ github.ref_name }}
          token: ${{ secrets.BOT_PAT }}
          delete-branch: false
          title: 'nri-plugins-operator ${{ github.ref_name }}'
          commit-message: 'Submit operator nri-plugins-operator ${{ github.ref_name }}'
          body: |
            Added OLM bundle for [nri-plugins operator ${{ github.ref_name }}](https://github.com/containers/nri-plugins/releases/tag/${{ github.ref_name }})
            > Auto-generated by `Github Actions Bot`

```

.github/workflows/publish-olm-bundle.yaml
fmuyassarov: Good catch about the author. Apparently I forgot to add an author in a format that git understans even though I was using it in my tests https://github.com/fmuyassarov/nri-plugins/actions/runs/12763538771/workflow#L52.

askervin:
I think this is great work. Thank you @fmuyassarov for automating these pull requests!

I don't want to request big changes before merging the first version. Having it tested without hardcoded versions would be enough for me.

Two questions, mainly on future steps:

Testability: Would it be convenient to enable testing this workflow so that it would recognize when it is running on a nri-plugins repo fork (instead of owned by containers), and in that case it would create the pull request to the community-operators repo fork of the same owner instead of upstream? And perhaps this workflow could be triggered by a pull_request_comment like "/test-operator-release" or something (that would obviously never create pull request to the upstream repo).

Re-releasing: do we ever want to create a pull request that changes an existing release? If not, maybe we should then explicitly fail on the "Copy the bundle..." step if the version directory already exists.

klihub (approved):
@fmuyassarov @askervin My suggestion would be to merge this and test-drive it, so Feruz can get rid of it.

Then we can beat it into final shape with one or more new PRs as necessary. WDYT ?

askervin:
Tried to play with this a bit, and it's quite broken.

@klihub, this looks necessary: https://github.com/containers/nri-plugins/commit/afd0c374256e14a1f120be754c06bce91f5e7308

Maybe we should coordinate merging so that it'll get in very soon after this one.

klihub:
> Tried to play with this a bit, and it's quite broken.
> 
> @klihub, this looks necessary: [afd0c37](https://github.com/containers/nri-plugins/commit/afd0c374256e14a1f120be754c06bce91f5e7308)
> 
> Maybe we should coordinate merging so that it'll get in very soon after this one.

@askervin I have updated that slightly, https://github.com/klihub/nri-plugins/commit/5b80b676f92c8220306d864c1c7f7fc107ea5dcd, and have it on this [devel/olm-bundle-submission branch](https://github.com/klihub/nri-plugins/tree/devel/olm-bundle-submission). If a test trigger is used, it now files the PR against the fork of the user who opened (or reopened) the test trigger issue. Also, if filing the PR succeeded, it closes the test trigger issue in the end.

If we decide to merge this, I can open immediately a new PR with those additional commits. An alternative would be to replace this with a new PR with the additional commits included and merge that one in a single go.

.github/workflows/publish-olm-bundle.yaml
```diff
@@ -0,0 +1,80 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Determine Repository Variable
+      # Remove this conditional logic once the workflow behavior
+      # is validated
+          run: |
```
klihub, line 25: This looks to  me like an incorrect indentation of `run`.

askervin (approved):
This can be merged when we are ready to merge #464 right after it.

.github/workflows/publish-olm-bundle.yaml
fmuyassarov: fixed

PR #464, opened 2025-01-21
.github: update OLM submission to allow both release and test triggers.

klihub:
This PR is stacked on top of #460.

This PR updates OLM submission to allow both real PR submissions, triggered by publishing a release, and test submissions. In particular trigger a test run if an issue is opened or reopened with a title matching an expected special pattern. For test runs, take the pattern from the title and use the issue filer's fork as the target repo for the PR. If filing a PR against that repo succeeds, close the triggering issue at the end.


askervin (approved):
Very nice! Thanks @klihub!

askervin (approved):
LGTM

kad (approved):
lgtm

pfl (approved):
LGTM

.github/workflows/publish-olm-bundle.yaml
```diff
@@ -0,0 +1,162 @@
...
+          path: community-operators
+          ref: main
+          token: ${{ secrets.BOT_PAT }}
+
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        id: import-gpg
+        with:
+          gpg_private_key: ${{ secrets.BOT_GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.BOT_GPG_PASSPHRASE }}
```
fmuyassarov, line 128:
```suggestion
          passphrase: ${{ secrets.BOT_GPG_PASSPHRASE }}
          fingerprint: "15CF4622AD8FE3794F15EED3E3063FF042266F81"
```

fmuyassarov, line 128:
```suggestion
          passphrase: ${{ secrets.BOT_GPG_PASSPHRASE }}
          fingerprint: "15CF4622AD8FE3794F15EED3E3063FF042266F81"
```
@kad has asked if it would be possible to use the subKey. It is possible via the fingerprint field and this an actual fingerprint for the same secret stored on GitHub.

kad: hardcoding that value here doesn't make sense. better to have it as a separate secret configured in repo (if really needed)

fmuyassarov: Well I am just telling that it could be done this way since you asked.

kad: I think @klihub  already did that in the repo settings, and gpg_private_key is pointing to signing subkey.

fmuyassarov: Have you tested that you can use the subkey via `gpg_private_key` instead of using `fingerprint` ?

fmuyassarov: You could have asked me for the subkey. Seems I just wasted some time. But as long as things work as you expected i'm fine with that.

fmuyassarov (approved):

PR #468, opened 2025-01-22
workflow: sign Helm packages and upload provenance files

fmuyassarov:
This commit ensures Helm packages are signed and their provenance files are uploaded for verification.

### DIFF
--- a/.github/workflows/package-helm.yaml
+++ b/.github/workflows/package-helm.yaml
@@ -25,6 +25,29 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+     -  name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        id: import-gpg
+        with:
+          gpg_private_key: ${{ secrets.BOT_GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.BOT_PASSPHRASE }}
+
+      - name: Export GPG private key
+        run: |
+          gpg --batch \
+              --yes \
+              --pinentry-mode loopback \
+              --passphrase ${{ secrets.BOT_PASSPHRASE }} \
+              --export-secret-keys ${{ secrets.BOT_GPG_ID }} \
+              > ~/.gnupg/secring.gpg
+
+      - name: Verify GPG secret key file
+        run: |
+          if [ ! -f ~/.gnupg/secring.gpg ]; then
+            echo "Error: GPG secret key file '~/.gnupg/secring.gpg' not found!" >&2
+            exit 1
+          fi
+
       - name: Install Helm
         uses: azure/setup-helm@v4.0.0
 
@@ -32,7 +55,14 @@ jobs:
         run: |
           find "$CHARTS_DIR" -name values.yaml | xargs -I '{}' \
               sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" -i '{}'
-          helm package --version "$GITHUB_REF_NAME" --app-version "$GITHUB_REF_NAME" "$CHARTS_DIR"/*
+          echo ${{ secrets.BOT_PASSPHRASE }} | helm package \
+            --sign \
+            --key ${{ steps.import-gpg.outputs.email }} \
+            --keyring ~/.gnupg/secring.gpg \
+            --version "$GITHUB_REF_NAME" \
+            --app-version "$GITHUB_REF_NAME" \
+            "$CHARTS_DIR"/* \
+            --passphrase-file "-"
           find . -name '*.tgz' -print | while read SRC_FILE; do
             DEST_FILE=$(echo $SRC_FILE | sed 's/v/helm-chart-v/g')
             mv $SRC_FILE $DEST_FILE
@@ -44,7 +74,9 @@ jobs:
           name: ${{ github.ref_name }}
           draft: true
           append_body: true
-          files: nri-*helm-chart*.tgz
+          files: |
+            nri-*helm-chart*.tgz
+            nri-*helm-chart*.tgz.prov
 
   unstable:
     if: ${{ !startsWith(github.ref, 'refs/tags/v') }}


fmuyassarov:
@klihub Would it be okay for you add the secret key ID as `BOT_GPG_ID` secret?

klihub:
> @klihub Would it be okay for you add the secret key ID as `BOT_GPG_ID` secret?

Sure, I'll add it.

klihub:
> > @klihub Would it be okay for you add the secret key ID as `BOT_GPG_ID` secret?
> 
> Sure, I'll add it.

@fmuyassarov I added `21866B474ADD98EC737D40A27202502C40EF3A9F` as the repository environment variable `BOT_GPG_ID`, since AFAICT it does not need to be a secret.

fmuyassarov:
> 21866B474ADD98EC737D40A27202502C40EF3A9F

Thank you. That's the correct one.

.github/workflows/package-helm.yaml
```diff
@@ -25,14 +25,43 @@ jobs:
...
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.BOT_GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.BOT_PASSPHRASE }}
+
+      - name: Export GPG private key
+        run: |
+          gpg --batch \
+              --yes \
+              --pinentry-mode loopback \
```
fmuyassarov, line 38: this is added so that passphrase can be passed non-interactively

.github/workflows/package-helm.yaml
```diff
@@ -25,14 +25,43 @@ jobs:
...
               sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" -i '{}'
-          helm package --version "$GITHUB_REF_NAME" --app-version "$GITHUB_REF_NAME" "$CHARTS_DIR"/*
+          echo ${{ secrets.BOT_PASSPHRASE }} | helm package \
+            --sign \
+            --key ${{ secrets.BOT_GPG_ID }} \
+            --keyring ~/.gnupg/secring.gpg \
+            --version "$GITHUB_REF_NAME" \
+            --app-version "$GITHUB_REF_NAME" \
+            "$CHARTS_DIR"/* \
+            --passphrase-file "-"
```
fmuyassarov, line 64: same intention here

.github/workflows/package-helm.yaml
```diff
@@ -25,14 +25,43 @@ jobs:
...
+              --passphrase ${{ secrets.BOT_PASSPHRASE }} \
+              --export-secret-keys ${{ secrets.BOT_GPG_ID }} \
+              > ~/.gnupg/secring.gpg
+
+      - name: Verify GPG secret key file
+        run: |
+          if [ ! -f ~/.gnupg/secring.gpg ]; then
+            echo "Error: GPG secret key file '~/.gnupg/secring.gpg' not found!" >&2
+            exit 1
+          fi
```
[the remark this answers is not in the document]
klihub: Ah, I see... it's mentioned in the docs that Helm needs the keyring in legacy format.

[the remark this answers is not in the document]
fmuyassarov: yes, it is Helm who requires it unfortunately

klihub (commented):
@fmuyassarov Have you had time to try this somewhere for real, for instance in your own forked repo ?

.github/workflows/package-helm.yaml
```diff
@@ -25,14 +25,43 @@ jobs:
...
         uses: azure/setup-helm@v4.0.0
 
       - name: Package Stable Helm Charts
         run: |
           find "$CHARTS_DIR" -name values.yaml | xargs -I '{}' \
               sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" -i '{}'
-          helm package --version "$GITHUB_REF_NAME" --app-version "$GITHUB_REF_NAME" "$CHARTS_DIR"/*
+          echo ${{ secrets.BOT_PASSPHRASE }} | helm package \
+            --sign \
+            --key ${{ secrets.BOT_GPG_ID }} \
```
klihub, line 59: In my local test, this does not work. Looks like helm requires the uid of the key, so AFAICT in this case that would be `nripluginsbot@yahoo.com`.

klihub: So locally, with my normal keyring in use which has both the bot's main key and the subkeys present, the only way how I got helm to successfully sing the chart package with the signing subkey was to:
- editing the main key, removing the signing capability (leaving only key certification)
- then using `nripluginsbot@yahoo.com` uid for `--key`

Otherwise if there are multiple keys, since there is no way to force stupid Helm to use the correct one, it always tried to sign using the master key.

fmuyassarov:
> @fmuyassarov Have you had time to try this somewhere for real, for instance in your own forked repo ? 

Yes, on my fork.

.github/workflows/package-helm.yaml
fmuyassarov: > So locally, with my normal keyring in use which has both the bot's main key and the subkeys present, the only way how I got helm to successfully sing the chart package with the signing subkey was to:
> 
> - editing the main key, removing the signing capability (leaving only key certification)
> 
> - then using `nripluginsbot@yahoo.com` uid for `--key`
> 
> 
> 
> Otherwise if there are multiple keys, since there is no way to force stupid Helm to use the correct one, it always tried to sign using the master key.

We can export a specific secret key and use that particular keyring with Helm?

fmuyassarov: Hey @klihub .
I've tested using the subkey with Helm, and it worked successfully. Regarding the value for the` --key `flag, you’re correct - I tested it with the email address, and it passed.

You can see the example run [here](https://github.com/fmuyassarov/nri-plugins/actions/runs/12928042336)  and the workflow I used [here](https://github.com/fmuyassarov/nri-plugins/actions/runs/12928042336/workflow). 

Based on the result above, I have now updated the PR. 
When dealing with multiple keys, it can be challenging indeed if several GPG keys share the same email address. However, for the purpose workflow, which is based on a single GPG key (subkey), this won't be an issue (I hope).

klihub: > Based on the result above, I have now updated the PR. When dealing with multiple keys, it can be challenging indeed if several GPG keys share the same email address. However, for the purpose of the workflow, which is based on a single GPG key (subkey), this won't be an issue (I hope).

It's a bit worth than that because if you have a signing-only subkey but you haven't removed the signing capability from your main key, there is no way to select which one Helm will use (picked by uid vs. instead of fingerprint), and AFAICT it then will then try to use the main one. But this should not be a problem for us, because I removed signing from the bots main key.

klihub (approved):
I'd be ready to merge this. @askervin @kad WDYT ?

klihub:
@fmuyassarov Do you know if it is possible to push the resulting provenance file also to OCI registry next to the chart itself ?

It does not matter for this PR, but I was just wondering if we could sign our development charts, too, which are hosted in the image registry...

fmuyassarov:
> @fmuyassarov Do you know if it is possible to push the resulting provenance file also to OCI registry next to the chart itself ?
> 
> It does not matter for this PR, but I was just wondering if we could sign our development charts, too, which are hosted in the image registry...

let me check....

fmuyassarov:
> Do you know if it is possible to push the resulting provenance file also to OCI registry next to the chart itself ?
@klihub 
According to https://helm.sh/docs/topics/registries/#the-push-subcommand, the provenance file gets uploaded automatically via `helm push`.

PR #469, opened 2025-01-22
balloons: expose balloons and optionally containers with affinity in NRT

askervin:


fmuyassarov:
Would you mind to share why did you mean (optionally) on the PR title?

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -326,8 +329,186 @@ func (p *balloons) ExportResourceData(c cache.Container) map[string]string {
...
-	return nil
+func (p *balloons) GetTopologyZones() []*policy.TopologyZone {
+	exposeAllocations := true // make this a balloons policy configuration/debug option
+	zones := []*policyapi.TopologyZone{}
+	sysmCpu := 1000 * p.cpuTree.cpus.Size()
+	for _, bln := range p.balloons {
+		// Expose every balloon as a separate zone.
+		zone := &policyapi.TopologyZone{
+			Name: bln.PrettyName(),
+			Type: "balloon",
```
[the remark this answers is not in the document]
askervin: I'd rather keep the zone name "balloon", as it is a real zone whose resources can be allocated.

On the other hand, "allocation for container" is more of an allocation in its parent's zone than a true zone itself. In typical scenarios it has access to all resources in the zone, just like any other allocation in the same zone, too.

askervin:
> Would you mind to share why did you mean (optionally) on the PR title?

`showContainersInNrt` policy and balloon option are now added. They didn't exist in the draft yet. They default to false, that is, only balloon instances are shown by default.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -326,8 +329,186 @@ func (p *balloons) ExportResourceData(c cache.Container) map[string]string {
...
+		// "allocation for container" subzone whose parent is
+		// the balloon zone. The subzone has following
+		// resources:
+		//
+		// "Capacity": container's resource usage
+		// limit. Regarding CPU capacity specifically, CPU
+		// usage of the container is limited by
+		// resources.limits.cpu and the number of allowed CPUs
+		// (balloon + shared). "Capacity" reflects the
+		// effective (the tighter) limit.
```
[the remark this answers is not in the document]
fmuyassarov: Thanks for the detailed explanation.

cmd/plugins/balloons/policy/balloons-policy.go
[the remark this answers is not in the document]
fmuyassarov: fair enough

config/crd/bases/config.nri_balloonspolicies.yaml
```diff
@@ -513,6 +520,18 @@ spec:
                   type: string
                 description: Reserved (CPU) resources for kube-system namespace.
                 type: object
+              showContainersInNrt:
+                description: |-
+                  ShowContainersInNrt controls whether containers in balloons
```
klihub, line 525: nit: To address the concerns raised by @fmuyassarov, shouldn't we add here a word of warning about how enabling this will affect the amount of data pushed to the CR / API server from nodes with a large amount of containers ?

deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
```diff
@@ -513,6 +520,18 @@ spec:
                   type: string
                 description: Reserved (CPU) resources for kube-system namespace.
                 type: object
+              showContainersInNrt:
+                description: |-
+                  ShowContainersInNrt controls whether containers in balloons
```
klihub, line 525: nit: Same here, especially since this is the global default... So to address the concerns raised by @fmuyassarov, shouldn't we add here a word of warning about how enabling this will affect the amount of data pushed to the CR / API server from nodes with a large amount of containers ?

config/crd/bases/config.nri_balloonspolicies.yaml
klihub: Then again, I guess considering how mild the constraints are when running with the default etcd configuration, maybe there is no strict need for such a warning...

askervin: I think this is a valid concern anyway. At least better to give a hint that there could be issues, this does not come for free. If it's not the size of the message (due to hundreds of containers running on the node), then at least about the extra api-server + etcd traffic (a lively 1000+ node cluster)...

Thanks @klihub & @fmuyassarov on these notes!

askervin: A sentence about potential costs added.

klihub (approved):

fmuyassarov (commented):
Thanks Antti. Looks better I think. 
IINW Changing the `showContainersInNrt` field has no effect until the next container event. If true, do you think it is worth of mentioning it in the document? Otherwise, LGTM.

fmuyassarov (approved):

kad:
LGTM. It might be good to have `NRT` capitalisation instead of `Nrt`, for readability, but that's not critical.

pfl (approved):

PR #473, opened 2025-02-21
WIP: balloons: add RDT support

askervin:
This PR adds rudimentary RDT support to balloons.

Assuming that resctrl file system is mounted and a Class of Service (COS) defined, processes in containers whose balloon has matching `rdtClass`, will be placed into that RDT class.

### DIFF
--- a/config/crd/bases/config.nri_balloonspolicies.yaml
+++ b/config/crd/bases/config.nri_balloonspolicies.yaml
@@ -268,6 +268,13 @@ spec:
                         placed on separate balloons. The default is false: prefer
                         placing containers of a pod to the same balloon(s).
                       type: boolean
+                    rdtClass:
+                      description: |-
+                        RdtClass controls RDT class of processes in the containers
+                        of a balloon. The default is empty: no RDT class is set. If
+                        defined, processes will be added to the directory
+                        corresponding the class in the resctrl file system.
+                      type: string
                     shareIdleCPUsInSame:
                       description: |-
                         ShareIdleCpusInSame <topology-level>: if there are idle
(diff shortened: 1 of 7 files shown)


config/crd/bases/config.nri_balloonspolicies.yaml
```diff
@@ -268,6 +268,13 @@ spec:
                         placed on separate balloons. The default is false: prefer
                         placing containers of a pod to the same balloon(s).
                       type: boolean
+                    rdtClass:
+                      description: |-
+                        RdtClass controls RDT class of processes in the containers
+                        of a balloon. The default is empty: no RDT class is set. If
+                        defined, processes will be added to the directory
+                        corresponding the class in the resctrl file system.
+                      type: string
```
klihub, line 277: nit: I would update the wording, since (with #483 and the other already merged related PRs) we still might end up setting some RDT class for the container (for instance based on its QoS class).

So maybe rephrase to something like this ?

```
RdtClass controls RDT class of processes in the containers
of a balloon. The default is empty: no balloon specific RDT
class is set. If defined, processes will be added to the directory
corresponding the class in the resctrl file system. Note that if
this class is omitted, the container might still get assigned to
an RDT class based on its Pod QoS class.
```

docs/resource-policy/policy/balloons.md
```diff
@@ -188,6 +188,10 @@ Balloons policy parameters:
   - `cpuClass` specifies the name of the CPU class according to which
     CPUs of balloons are configured. Class properties are defined in
     separate `cpu.classes` objects, see below.
+  - `rdtClass` specifies the name of the RDT class. Corresponding
```
klihub, line 191: nit: whether a non-existent class will result in failure to create the container depends (unfortunately) on the runtime in use and its configuration. Notably in containerd the default configuration implies failure, but this can be changed by flipping a configuration variable.

askervin:
@klihub, IIRC, after your proper work in Balloons / RDT this draft PR is of no use. I'm closing it now, and let's re-open if needed.

PR #474, opened 2025-03-05
topology-aware: exclude isolated CPUs from policy-picked reserved cpusets.

klihub:
This PR updates the reserved cpuset setup code to
- accept a reserved cpuset if it consists of only a single hand-picked isolated CPU
- exclude isolated CPUs from policy-picked reserved cpusets

### DIFF
--- a/cmd/plugins/topology-aware/policy/topology-aware-policy.go
+++ b/cmd/plugins/topology-aware/policy/topology-aware-policy.go
@@ -577,11 +577,11 @@ func (p *policy) checkConstraints() error {
 		}
 
 		p.reserveCnt = (int(qty.MilliValue()) + 999) / 1000
-		// Use CpuAllocator to pick reserved CPUs among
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
Looks good. 
One small nit: would it make sense to include this logging commit https://github.com/containers/nri-plugins/pull/474/commits/caea3ee72fd72a436ed489d29b88b443721bd405 in a related commit (https://github.com/containers/nri-plugins/pull/474/commits/598ec9c4bddc2718c73bccceed48b7a831bac740) instead of a separate one? Just thinking about keeping the commit history clean, but I'm okay to keep as it is as well if there is a reason behind.

Just to clarify for myself, you are not allowing to have more than one isolated CPU for the reserved cpuset because isolated CPUs don't participate in load balancing right?

I wonder if we should do the similar check in Balloons as well...

fmuyassarov (approved):

askervin (approved):
LGTM

PR #477, opened 2025-03-10
topology-aware: rework building the topology pool tree.

klihub:
This PR reworks and simplifies how the tree of pools is built in the topology-aware policy, using the detected hardware topology.

We now build the tree by walking the topology from sockets to dies to NUMA nodes closest to dies. We assign local CPU and memory close these CPUs in tandem to each pool as we build the tree. Once the tree is built, we go through each pool node again and assign CPUless NUMA nodes to the same nodes where their closest NUMA nodes with CPUs are assigned to.

Like earlier, the tree is constructed to be minimal. If the system has a single socket, we omit the virtual root and use the sole socket as the root pool to represent the whole system (since it would have the same resources as its parent virtual root node). If a socket has a single die, we omit that die from the subtree under the socket (since it would have the same resources as its parent socket node). If a die has a single close NUMA node, we omit the tree node that would represent CPUs close to a die-local NUMA node (since it would have the same resources as its parent die or socket node).

### DIFF
--- a/cmd/plugins/topology-aware/policy/pools.go
+++ b/cmd/plugins/topology-aware/policy/pools.go
@@ -29,308 +29,300 @@ import (
 
 // buildPoolsByTopology builds a hierarchical tree of pools based on HW topology.
 func (p *policy) buildPoolsByTopology() error {
-	omitDies, err := p.checkHWTopology()
-	if err != nil {
+	if err := p.checkHWTopology(); err != nil {
 		return err
 	}
 
-	// Notes:
-	//   we never create pool nodes for PMEM-only NUMA nodes (as these
-	//   are always without any close/local set of CPUs). We instead
-	//   assign the PMEM memory of such a node to one of the closest
-	//   normal (DRAM) pool NUMA nodes.
+	// We build a tree of pools by examining sockets, dies, and NUMA nodes.
+	// If we have multiple sockets, we create an extra virtual root for the
+	// whole system, otherwise we use the socket as the root of the tree.
+	// Under sockets, we create die nodes, and under dies, we create NUMA
+	// nodes. We omit a die pool if is is the only die in a socket. We omit
+	// a NUMA node if it is the only child pool under its parent.
 	//
-	//   Akin to omitting lone dies from their parent, we omit from the
-	//   pool tree each NUMA node that would end up being the only child
-	//   of its parent (a die or a socket pool node). Resources for each
-	//   such node will get discovered by and assigned to the would be
-	//   parent which is now a leaf (die or socket) node in the tree.
-	//
-	//   The PMEM memory of (omitted) PMEM-only nodes is assigned
-	//   to one of the closest normal (DRAM) NUMA nodes. This right
-	//   assignment has already been calculated by assignPMEMNodes().
-	//   However, making the corresponding assignment in the pool
-	//   tree is a bit more involved as the DRAM node where a PMEM
-	//   node has been assigned to might have gotten omitted from the
-	//   tree if it ended up being a lone child. We use the recorded
-	//   per NUMA node surrogates to find both if and where resources
-	//   of omitted DRAM NUMA nodes need to get assigned to, and also
-	//   where PMEM NUMA node resources need to get assigned to.
+	// We discover pool resources by first assigning the closest CPUs to
+	// the pool, then the set of NUMA nodes closest to any of the CPUs in
+	// the pool. Then those NUMA nodes without CPU locality get assigned
+	// to the pool for which any of the closest neighbor with CPU locality
+	// is already assigned to the pool. For the root node, we also assign
+	// the system-wide memory resources without CPU locality. Finally, any
+	// NUMA node otherwise unassigned is assigned to the root pool (which
+	// is done by simply assigning all NUMA nodes without CPU locality to
+	// the root pool).
 
-	log.Debug("building topology pool tree...")
+	log.Info("building pools by HW topology...")
 
 	p.nodes = make(map[string]Node)
-
-	// create a virtual root node, if we have a multi-socket system
-	if p.sys.SocketCount() > 1 {
-		p.root = p.NewVirtualNode("root", nilnode)
-		p.nodes[p.root.Name()] = p.root
-		log.Debug("  + created pool (root) %q", p.root.Name())
-	} else {
-		log.Debug("  - omitted pool virtual root (single-socket system)")
-	}
-
-	// create socket nodes, for a single-socket system set the only socket as the root
-	sockets := map[idset.ID]Node{}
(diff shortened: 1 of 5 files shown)


fmuyassarov:
sorry for radio silence, I will go through it tomorrow morning ...

fmuyassarov:
meanwhile, perhaps it would be good to rebase

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -29,308 +29,249 @@ import (
...
+			if reverse[closeID] == nil {
+				reverse[closeID] = idset.NewIDSet(id)
+			} else {
+				reverse[closeID].Add(id)
 			}
 		}
 	}
 
+	log.Info("closest memory nodes with CPUs: %v", close)
+	log.Info("reverse closest memory nodes with CPUs: %v", reverse)
```
fmuyassarov, line 268: I'm not sure about significant benefits of these logs, but I don't see any reason to block the PR due to this.

fmuyassarov (approved):

cmd/plugins/topology-aware/policy/pools.go
klihub: The returned information directly controls how nodes without any CPU locality will get assigned to pools. These nodes are currently often PMEM or HBM, and I think in the future they could be CXL-attached memory as well. That's the reason why we log it, to provide some info related to their eventual assignment.

But I'll check and see if I can either make the logging more reasonable, or get rid of at least one of those messages.

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -29,308 +29,249 @@ import (
...
-	socketNodes := map[idset.ID]cpuset.CPUSet{}
-	for _, socketID := range p.sys.PackageIDs() {
-		pkg := p.sys.Package(socketID)
-		socketNodes[socketID] = system.CPUSetFromIDSet(idset.NewIDSet(pkg.NodeIDs()...))
-	}
-	for id1, nodes1 := range socketNodes {
-		for id2, nodes2 := range socketNodes {
-			if id1 == id2 {
+func (p *policy) getCloseMemsWithCpu() map[idset.ID]idset.IDSet {
+	close := map[idset.ID]idset.IDSet{}
```
askervin, line 234: I find this implementation a bit complicated, as it constructs first "close" map from a cpuful to any (cpuless or cpuful) node, and returns that reversed. As far as I can see, the reverse map will not necessarily include closest cpuful node for every cpuless node, yet I suppose that might have been wanted. (If I misunderstood the purpose, perhaps a comment would be good addition.)

Could this be simplified into single loop that constructs only the map that is returned, with descriptive variable names? Pseudolike:

```golang
for _, nodeWithoutCpus := range p.sys.NodeIDs() {
    if nodeWithoutCpus has CPUs: continue
    for nodeWithCpus, dist := range nodeWithoutCpus.Distance() {
        if nodeWithCpus has no CPUs: continue
        switch {
        ....
        case when appropriate: closestCpuNodes[nodeWithoutCpus].Add(nodeWithCpus)
        ....
        }
    }
    return closestCpuNodes
}
```

askervin (commented):
I have a feeling that I might have misunderstood something, so maybe it's best to get clarification....

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -29,308 +29,249 @@ import (
...
+	// omit a NUMA node if it would be the only child pool in its parent.
 	//
-	//   Akin to omitting lone dies from their parent, we omit from the
-	//   pool tree each NUMA node that would end up being the only child
-	//   of its parent (a die or a socket pool node). Resources for each
-	//   such node will get discovered by and assigned to the would be
-	//   parent which is now a leaf (die or socket) node in the tree.
+	// We assign the closest CPUs to each pool, then the set of NUMA nodes
+	// closest to any of the CPU in the pool. NUMA nodes without close CPUs
+	// get assigned to alls pools where the closest NUMA node(s) with CPUs
```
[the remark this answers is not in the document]
klihub: I deliberately phrased it as 'close CPUs'. It feels wrong to me to say something like 'NUMA node without CPUs', as it has traditionally been an abstraction for related to memory controllers/channels and attached memory, with additional information about which CPUs are closest from a latency point of view to that NUMA node.

cmd/plugins/topology-aware/policy/pools.go
klihub: Hmm, true. If the closest nodes are without CPUs then we fail to assign memory only NUMA nodes to ones with close CPUs. Let me try to rework and fix that.

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -29,308 +29,249 @@ import (
...
-		}
-	}
-	return count
+func (p *policy) buildNumaNodePool(socketID, nodeID idset.ID, parent Node) {
+	node := p.NewNumaNode(nodeID, parent)
+	p.nodes[node.Name()] = node
+
+	cpus := p.sys.Node(nodeID).CPUSet()
+	node.node.noderes = p.getCpuSupply(node, cpus)
+	node.node.freeres = node.node.noderes.Clone()
```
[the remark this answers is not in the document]
klihub: I chose the current discovery-during-construction implementation over a discovery-with-a-depth-first-traversal, because this way we always have a concrete node type at hand instead of a Node interface. When/if we need to do some node type specific quirks or exceptions, it's easier/more straighforward this way.

cmd/plugins/topology-aware/policy/pools.go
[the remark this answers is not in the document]
klihub: Okay, I tried to now consistently use the terms 'NUMA node with CPU locality' and 'NUMA node without CPU locality' in this patch series.

cmd/plugins/topology-aware/policy/pools.go
klihub: Okay, I reworked both the CPU and memory discovery code. Both of them are now consolidated into a single functions, the corresponding `getCpuSupply()` and `getMemSupply()`, to reduce repetitive code. Memory discovery now largely relies on the newly introduced sysfs functions for getting NUMA node neighbor memsets by distance, and node filtering predicates. NUMA nodes without CPU locality are deliberately now assigned to any pool where any of their closest neighbor with CPU locality is assigned.

cmd/plugins/topology-aware/policy/pools.go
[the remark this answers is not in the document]
klihub: I consolidated CPU and memory resource discovery to a single function each to reduce the amount of repetitive code.

askervin (approved):
LGTM

marquiz (approved):
LGTM

PR #478, opened 2025-03-11
[resmgr RDT+blkio 1/2]: resolve QoS-to-RDT class markers in updates.

klihub:
When generating container updates, the current code leaves a QoS-to-RDT-Class marker unresolved, which is a leftover from CRI Resource Manager where the RDT controller did the final resolution. We have no such logic in NRI enabled runtimes, so resolve QoS class markers before generating an container update.


askervin (approved):
LGTM

klihub:
Closing this.

I will roll another PR which will
- make RDT and Block I/O class support globally configurable (and disabled by default), and
- take care of RDT PodQoS-to-class resolution

PR #480, opened 2025-03-12
resmgr: eliminate extra cache-level container state 'overlay'.

klihub:
Eliminate the cache-level extra container state 'overlay' and start tracking container state using the NRI provided container member variable alone. For containers being created, make the temporary creating state transition (almost) atomic by using a new insertion option to set the desired state.


kad (approved):
LGTM

PR #481, opened 2025-03-12
resmgr: eliminate extra RDT class 'overlay'.

klihub:
Eliminate cache level RDT class 'overlay' and start tracking container RDT class using the NRI provided container member field alone. Update RDT class setter to generate a container adjustment or update.


fmuyassarov (approved):

kad (approved):
LGTM

PR #482, opened 2025-03-12
resmgr: eliminate extra block I/O class 'overlay'.

klihub:
Eliminate cache level block I/O class 'overlay' and start tracking container block I/O class using the NRI provided container member field alone. Update I/O class setter to generate a container adjustment or update.


fmuyassarov (approved):

kad (approved):
LGTM

PR #483, opened 2025-03-12
resmgr: configurable RDT and block I/O class control.

klihub:
Note: this PR is stacked on #481 and #482.

This PR implements configurable RDT and block I/O class control. In particular, it
- adds RDT class configuration to the ballons, topology-aware and common resmgr configuration
- adds configuration for enabling/disabling RDT class and block I/O adjustment
- adds configuration for RDT and block I/O class fallback to Pod QoS class
- updates the resource manager and the cache to obey the new configuration settings
- disabled both RDT and block I/O class control by default (which avoids failure by default with stock runtime configurations)


pkg/apis/config/v1alpha1/resmgr/blockio/config.go
```diff
@@ -0,0 +1,29 @@
...
+// Config provides runtime configuration for class-based block I/O prioritization.
+// +k8s:deepcopy-gen=true
+type Config struct {
+	// Enable class based block I/O prioritization and throttling. When
+	// enabled, policy implementations can adjust block I/O priority by
+	// by assigning containers to block I/O priority classes.
+	// +optional
+	Enable bool `json:"enable,omitempty"`
+	// usePodQoSAsDefault controls whether a container's Pod QoS class
+	// is be used as its block I/O class, if this is otherwise unset.
```
askervin, line 26: s/is be/is to be/

askervin (commented):
Suggesting only a tiny typo fix.

I'm trying to imagine that in the future we could enable blockio class definitions in the policy-agnostic "control" configuration section. I mean, we could allow user to define "control.blockio.classes" similarly to "control.cpu.classes". Orthogonally to that step, at some point balloon types could have their blockIOClass and rdtClass attributes, too.

I do not foresee any problems that this implementation as a starting point would introduce on those paths. I mean, as far as I can see, we don't need to cause "API break" in user perspective. Would you see any? Or would you have other ideas as next steps on these fronts?

So this looks very good in my point of view. Only the supernit.

pkg/apis/config/v1alpha1/resmgr/blockio/config.go
klihub: Thanks. Fixed this and the same typo in the corresponding RDT config.

klihub:
> Suggesting only a tiny typo fix.
> 
> I'm trying to imagine that in the future we could enable blockio class definitions in the policy-agnostic "control" configuration section. I mean, we could allow user to define "control.blockio.classes" similarly to "control.cpu.classes". Orthogonally to that step, at some point balloon types could have their blockIOClass and rdtClass attributes, too.
>
> I do not foresee any problems that this implementation as a starting point would introduce on those paths. I mean, as far as I can see, we don't need to cause "API break" in user perspective. Would you see any? Or would you have other ideas as next steps on these fronts?

This is similar to my pondering/dilemma: where to place these config bits, primarily in the custom resource definitions, secondarily in the package tree. I was considering `control`, too.  The reasons why I did not put them there are more based on current implementation details and code layout than end user UX. Currently everything under `control` (basically CPU frequency) has a dedicated controller and is outside of/extra to the realm of what NRI controls. The only UX-apect-ish consideration was that we are not truly in control of RDT or block I/O prioritization here. The control is in the CRI runtime/low level runtime, we merely try to initiate those controls over the NRI protocol.

But I think these considerations related to implementation details should not be the only or even the primary decisive factor. Especially not so, if we consider updating the model for RDT and block I/O in the future, for instance by allowing the user to define classes in our configuration. Then it would probably make more sense to put the configuration already now under `control`.

So maybe then it could be something like this:

```
control:
  rdt:
    enable: true
    usePodQoSAsDefaultClass: true
# And then in the future we could stitch in here the actual class definitions.
# Bare with me with the incorrect details here, it's just an illustration...
    classes:
      Gold:
        "100%"
      Silver:
        "50%"
      Orange:
        "1%"
```

klihub:
Orthogonally to that step, at some point balloon types could have their blockIOClass and rdtClass attributes, too.

Isn't that what we plan/try to do in the currently draft/WiP #473 ? I was especially rolling this series of RDT and block I/O class control PR series to pave the way for #473 at this point.

pkg/apis/config/v1alpha1/resmgr/rdt/config.go
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rdt
+
+// Config provides runtime configuration for class-based LLC cache control.
+// +k8s:deepcopy-gen=true
+type Config struct {
+	// Enable class based last level cache control. When enabled, policy
```
marquiz, line 20: Maybe _cache allocation and memory bandwidth control_ or so. RDT can also do L2 (on some platforms) and MBM

pkg/apis/config/v1alpha1/resmgr.go
```diff
@@ -41,4 +43,8 @@ type CommonConfig struct {
 	Log log.Config `json:"log,omitempty"`
 	// +optional
 	Instrumentation instrumentation.Config `json:"instrumentation,omitempty"`
+	// +optional
+	RdtClasses rdt.Config `json:"rdt,omitempty"`
+	// +optional
+	BlockIOClasses blockio.Config `json:"blockIO,omitempty"`
```
marquiz, line 49: nit/question: just wondering why the field names are *classes as the json name don't have that

```suggestion
	Rdt rdt.Config `json:"rdt,omitempty"`
	// +optional
	BlockIO blockio.Config `json:"blockIO,omitempty"`
```

pkg/resmgr/cache/cache.go
```diff
@@ -367,6 +367,11 @@ type InsertContainerOption func(c *container)
 // Cache also saves its state upon changes to secondary storage and restores
 // itself upon startup.
 type Cache interface {
+	// EnableRDTControl is used to enable RDT class control.
+	EnableRDTControl(bool)
```
marquiz, line 371: super nit :roll_eyes: just caught my eye that it looks somewhat odd that Enable() func can be used to disable it. Not sure if `ConfigureRDTControl()`, `SetRDTControl()`, `RDTControlOnOff()` or smth would be any better though...

pkg/apis/config/v1alpha1/resmgr.go
```diff
@@ -41,4 +43,8 @@ type CommonConfig struct {
 	Log log.Config `json:"log,omitempty"`
 	// +optional
 	Instrumentation instrumentation.Config `json:"instrumentation,omitempty"`
+	// +optional
+	RdtClasses rdt.Config `json:"rdt,omitempty"`
```
marquiz, line 47: We now have both mixed-case `Rdt` and all caps ` RDT` 

```suggestion
	RDTClasses rdt.Config `json:"rdt,omitempty"`
```

marquiz (commented):
Thanks @klihub, looks good to me. I just found some minor nits which can also be ignored

pkg/apis/config/v1alpha1/resmgr/rdt/config.go
klihub: Updated.

pkg/apis/config/v1alpha1/resmgr/blockio/config.go
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blockio
+
+// Config provides runtime configuration for class-based block I/O prioritization.
+// +k8s:deepcopy-gen=true
```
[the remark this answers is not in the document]
klihub: Okay, fair enough. But I'd prefer to handle that with a single PR, which converts all deprecated markers to the new one. And since that will be intrusive in `pkg/apis/config` and cause conflicts for some of the pending PRs which touch configuration, I'd propose to wait with that PR until the current pending ones are merged.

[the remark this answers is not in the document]
klihub: Updated.

pkg/apis/config/v1alpha1/resmgr/rdt/config.go
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rdt
+
+// Config provides runtime configuration for class-based LLC cache control.
+// +k8s:deepcopy-gen=true
```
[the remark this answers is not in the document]
klihub: Updated.

pkg/apis/config/v1alpha1/resmgr.go
klihub: Okay, now that you asked, I just realized that all of these JSON tags are useless and unnecessary. `CommonConfig` is not something we'd ever pass as a type over the wire. It is only used and constructed locally...

pkg/resmgr/cache/cache.go
klihub: Updated to use `ConfigureRDTControl()` (and `ConfigureBlockIOControl()`).

pkg/apis/config/v1alpha1/resmgr.go
klihub: Fixed to use everywhere RDT, where we can.

pkg/apis/config/v1alpha1/resmgr/blockio/config.go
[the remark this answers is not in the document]
fmuyassarov: @klihub I've created this [PR](https://github.com/containers/nri-plugins/pull/490) to replace all the existing markers.

klihub:
@askervin @fmuyassarov @marquiz I reworked the configuration bits according to [this earlier comment]( https://github.com/containers/nri-plugins/pull/483#issuecomment-2726765807). PTAL, if it looks more reasonable this way.

askervin (approved):
LGTM

klihub:
> super nit 🙄 just caught my eye that it looks somewhat odd that Enable() func can be used to disable it. Not sure if ConfigureRDTControl(), SetRDTControl(), RDTControlOnOff() or smth would be any better though...

@marquiz Do you have any other suggestions beside these ? There was a question in today's review if anyone has a better suggestion than `Configure{RDT,BlockIO}Control()` and I don't. Maybe you do.

marquiz (approved):
> There was a question in today's review if anyone has a better suggestion than `Configure{RDT,BlockIO}Control()` and I don't. Maybe you do.

I don't have a better name in mind :sweat_smile: I think we can get this merged.

LGTM

fmuyassarov (commented):
LGTM

PR #488, opened 2025-03-14
e2e-runner: collect all e2e test result data from the e2e test workdir.

klihub:
Now that we use a dedicated e2e test workdir, collect unconditionally all test result data from the workdir without using any selective globbing.


askervin (approved):
LGTM. Thanks for fixing our CI!

PR #491, opened 2025-03-18
topology-aware: allocate burstable container memory by requests.

klihub:
This patch series updates resource management common infra and the topology aware policy to

- pre-calculate a table of OOM Score adjustment to minimum memory request esimates
- use this table to estimate memory request for burstable containers based on their OOM score adjustment
- allocate memory for containers using memory requests (instead of limits)
- prefer pools with enough attached memory to fulfill both a containers memory request and limit

With these patches in place, the K8s scheduler and the NRI reference policies should be better aligned about the amount of available memory on the node.

Notes:
The estimate table is calculated assuming the simpler/simplest request-to-oom-adj formula, IOW assuming 0 pod level extra resource requests.

### DIFF
--- a/pkg/kubernetes/resources.go
+++ b/pkg/kubernetes/resources.go
@@ -14,6 +14,12 @@
 
 package kubernetes
 
+import (
+	"fmt"
+
+	"github.com/containers/nri-plugins/pkg/sysfs"
+)
+
 const (
 	// Constants for converting back and forth between CPU requirements in
 	// terms of milli-CPUs and kernel cgroup/scheduling parameters.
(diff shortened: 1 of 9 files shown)


pkg/kubernetes/resources.go
```diff
@@ -82,3 +103,94 @@ func QuotaToMilliCPU(quota, period int64) int64 {
...
+	for i, req := 0, int64(0); req < memCapacity; i, req = i+1, req+win {
+		adj := MemReqToOomAdj(req)
+		if req == 0 {
+			adjToReq[adj] = req
+			continue
+		}
+
+		lim := req - 10
+		for {
+			if next := MemReqToOomAdj(lim); next != adj {
```
askervin, line 140: Is there a risk that the inner loop starts to last pretty long when system memory capacity is high? Like terabytes?

klihub: Okay, I updated/fixed the estimate generation. Still a bit ugly, but the unit test, which does a fairly exhaustive checking of calculation in both directions, now passes.

askervin (approved):
LGTM

PR #492, opened 2025-03-24
system: add a helper for finding CPUs sharing caches

askervin:
This is a prerequisite for selecting CPUs so that sharing the same cache between many memory-bandwidth hungry workloads could be avoided.


kad (approved):
generally ok, but maybe unit test to add?

klihub (approved):

PR #493, opened 2025-03-25
balloons: introduce loadClasses for avoiding unwanted overloading in critical locations

askervin:
This patch series, stacked on top of PR #492, introduces user-defined `loadClasses` and associated balloonType option `loads`. The latter lists loads (names of loadClasses) that each balloon instance of the type is expected to generate.

The purpose is to start with quite minimal implementation that, however, has been designed to be extendable without configuration API breaks, to most obvious directions at least.

In the first version of user-defined loadClasses, one can choose a CPU topology `level` where load is generated. Supported levels are physical CPU `core` and `l2cache`.

Orthogonally to the level, user can specify if a balloon should be aware of the load generated by itself, or only loads generated by other balloons. Examples follow. If it avoids the load generated by itself, the balloon can get one hyperthread per physical CPU `core`, and leave the other threads to balloons that do not load the core. On the other hand, if it does not avoid its own load, then CPU allocation enables sharing the same L2 cache within the balloon while avoiding sharing the same L2 cache with other balloons generate load on the cache.

Regarding extensions. First, this is very straightforward to extended to other levels, too. For instance, if one wants to create a single balloon instance that contains one CPU from package0 and the other from package1, it would become possible by adding `package` level support to loadClasses. This can be useful for deploying monitors or daemons that need to run on every package for accessing or best locality to external devices.

Second, currently we create virtual devices only based on loaded level name. We can introduce options to include also loadClass name to the virtual device. This enables, for instance, specifying separately that loadClass "x" loads unit "x" from a physical core, while loadClass "y" loads unit "y" from a physical core. Then we can allocate thread0 to a balloon with load x and thread1 to a balloon with load y, but never both threads to balloons that both load x or y.

If more granularity is wanted, we can add "devices" or "accelerators" to load classes and use them in the virtual device name. And we can add some sort of capacity to a loadClass. For instance, if user would know that it takes 48 CPUs to saturate memory channels of single memory zone, then we could allow 2 x 24 CPU membw-balloons on one zone and then start filling other zones.

Making room for these extensions is the reason why I chose to repeat `Level` in option `overloadsLevelInBalloon`, because there may be other options, too (Level/Class/Device).


kad (approved):
Looks good. Maybe one more example for `l2cache` might be good to add, but in overall good already now.

klihub (approved):

PR #494, opened 2025-03-27
sysfs: only discover topology of online cpus

marquiz:
Fix a bug where all offline cpus were listed under package/die/cluster 0. The Linux kernel hides the topology information of offline cpus so we do not know where they belong.


klihub (approved):
LGTM.

klihub (changes requested):
onlineCPUs returns a cpuset.CPUSet, so we can use it for filtering, but as a replacement in the for range-loop.

klihub:
> LGTM.

But I was hasty and wrong.

askervin (approved):
LGTM

PR #496, opened 2025-03-28
e2e: fix isolcpu-resetting snippet in test.

klihub:
This patch set
- adds a `vm-set-kernel-cmdline` frontend to the already existing `{fedora,ubuntu}-set-kernel-cmdline`
- moves all of these from `run.sh` to `lib/vm.bash`
- fixes isolcpus-resetting snippet in topology-aware test


askervin (approved):
LGTM

kad (approved):
LGTM.

As side note, it might be not the best idea to use cpu0 as isolcpu.

fmuyassarov (approved):
LGTM

PR #498, opened 2025-03-31
topology-aware: better semantics for globally configured shared CPU preference.

klihub:
This patch series updates improves the semantics of a globally configured shared CPU allocation preference to behave more intuitively. When interpreting shared CPU allocation preferences, a globally configured positive preference now applies to all containers. For a mixed allocation of more than 1 full core, we continue requiring an explicitly annotated shared negative preference (IOW a globally configured negative shared preference does not imply mixed allocation with more than 2 full cores). The patch series also adds basic end-to-end tests for globally configured shared CPU allocation preference.


askervin (approved):
LGTM

PR #502, opened 2025-04-01
topology-aware: more consistent setup error handling.

klihub:
Don't log as fatal (and abort) when we fail to allocate reserved cpuset. Try to propagate the error back to the source config CR as with the rest of the errors.


askervin (approved):
LGTM

PR #503, opened 2025-04-01
e2e: fail reserved pool setup for the right reason.

klihub:
When we try to allocate too much reserved CPUs, don't fail startup due to passing an expected string value as an int. Alos, fetch node CR status error after expected failure.


askervin (approved):
LGTM

PR #505, opened 2025-04-01
.github: sign one Helm chart at a time.

klihub:
Work around Helm flakiness when packaging/signing multiple charts with a passphrase provided on stdin/pipe by signing one chart at a time.

### DIFF
--- a/.github/workflows/package-helm.yaml
+++ b/.github/workflows/package-helm.yaml
@@ -53,16 +53,19 @@ jobs:
 
       - name: Package Stable Helm Charts
         run: |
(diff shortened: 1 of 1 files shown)


fmuyassarov (approved):
This is a good finding and totally unexpected from Helm :( 
LGTM

klihub:
> This is a good finding and totally unexpected from Helm :( LGTM

I am guessing (for the time being without reading the code) that it tries to read the passphrase separately for each packaged chart. Will check it out later, once we get the release out of the way.

PR #507, opened 2025-04-02
resmgr: purge cached pod resource list upon pod stop/removal.

klihub:
This patch set fixes a problem which occurs if the Pod Resource API Get() is feature-gated off and one creates subsequently two identical pods with identical containers but with slightly different resources which then have different (NUMA node) locality. Currently, when podresapi.Get() is unavailable we emulate it with a List, followed by a lookup, caching any parsed per pod/container resources and the remaining unparsed slice of the resource list. However, since we miss purging cached items (when a pod is removed), case we fail to refresh the cached list in the aforementioned case.

This patch set adds support for purging cached entries and actively purge pod entries when a pod is stopped or removed.

### DIFF
--- a/pkg/agent/podresapi/resources.go
+++ b/pkg/agent/podresapi/resources.go
@@ -114,6 +114,16 @@ func (l *PodResourcesList) GetContainer(ns, pod, ctr string) *ContainerResources
 	return l.GetPodResources(ns, pod).GetContainer(ctr)
 }
 
+func (l *PodResourcesList) PurgePodResources(ns, pod string) {
+	if l == nil {
+		return
+	}
+
+	if podMap, ok := l.m[ns]; ok {
+		delete(podMap, pod)
+	}
+}
+
 // GetDeviceTopologyHints returns topology hints for the given container. checkDenied
 // is used to filter out hints that are disallowed.
 func (r *ContainerResources) GetDeviceTopologyHints(checkDenied func(string) bool) topology.Hints {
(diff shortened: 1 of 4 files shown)


pkg/agent/podresapi/resources.go
```diff
@@ -114,6 +114,16 @@ func (l *PodResourcesList) GetContainer(ns, pod, ctr string) *ContainerResources
...
 }
 
+func (l *PodResourcesList) PurgePodResources(ns, pod string) {
+	if l == nil {
+		return
+	}
+
+	if podMap, ok := l.m[ns]; ok {
+		delete(podMap, pod)
+	}
```
fmuyassarov, line 124:
```suggestion
	if podMap, ok := l.m[ns]; ok {
		if _, exists := podMap[pod]; exists {
			delete(podMap, pod)
		}
	}
```
nit, perhaps it is safer to check of the Pod actually exists in a given ns before trying to delete from the map?

fmuyassarov (commented):
Have a tiny question, otherwise LGTM

pkg/agent/podresapi/resources.go
klihub: Considering the golang docs/specs explicitly state for delete that ' If m is nil or there is no such element, delete is a no-op.', it is always better to do it this way. Otherwise you would always end up doing two map key lookups for a single successful delete and an failed key lookup attempt for a deletion attempt of a non-existing key, which is a clearly a net loss (compared to the current single lookup and a single lookup attempt).

fmuyassarov: fair enough

fmuyassarov (approved):

askervin (approved):
LGTM

PR #514, opened 2025-05-23
mempolicy: go interface for set_mempolicy and get_mempolicy syscalls

askervin:


pkg/mempolicy/mempolicy.go
```diff
@@ -0,0 +1,128 @@
...
+	maxNode := uint64(MAX_NUMA_NODES)
+	nodeMask := make([]uint64, maxNode/64)
+	nodeMaskPtr := unsafe.Pointer(&nodeMask[0])
+	_, _, errno := syscall.Syscall(SYS_GET_MEMPOLICY, uintptr(unsafe.Pointer(&mpol)), uintptr(nodeMaskPtr), uintptr(maxNode))
+	if errno != 0 {
+		return 0, []int{}, syscall.Errno(errno)
+	}
+
+	nodes := make([]int, 0)
+	for i := range int(maxNode) {
```
klihub, line 111: nit: if we have nodesToMask why not symmetrically split this out to a maskToNodes ?

klihub (commented):
Small nit. Otherwise looks good.

pkg/mempolicy/mempolicy.go
askervin: A just question. Very much nicer to have it symmetric. Fixed.

klihub (approved):

klihub (commented):
Is there any real benefit of not having this as (potentially separate commit but) part  of #515 ?

askervin:
> Is there any real benefit of not having this as (potentially separate commit but) part of #515 ?

This commit is in a separate PR to limit the lines of code for easier review. So if there was a change that would be reasonable stand-alone change as is, I filed a separate PR. This is comparable to PR #513 that adds a function to an interface, yet there is no user for the function in the same PR.

klihub (approved):

PR #515, opened 2025-05-23
mpolset: get/set memory policy and exec a command

askervin:
mpolset is a standaone executable that can be compiled into a static binary. In this form, it can be injected into containers, where the original command of a container can be prefixed with mpolset and arguments that define the default memory policy for the original command in the container.

If command is not given, mpolset prints the default memory policy. This offers an easy way to test how kernels handle missing nodes and various flag combinations, for instance.

For example: set, execute and get memory policy:

./mpolset -nodes 0-2 -mode MPOL_INTERLEAVE -- ./mpolset

### DIFF
--- a/cmd/plugins/memory-policy/mpolset/mpolset.go
+++ b/cmd/plugins/memory-policy/mpolset/mpolset.go
@@ -1,0 +1,164 @@
+// Copyright The NRI Plugins Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// mpolset is an executable that sets the memory policy for a process
+// and then executes the specified command.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/containers/nri-plugins/pkg/mempolicy"
+	"github.com/containers/nri-plugins/pkg/utils/cpuset"
+	"github.com/sirupsen/logrus"
+)
+
+type logrusFormatter struct{}
+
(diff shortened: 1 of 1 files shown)


cmd/plugins/memory-policy/mpolset/mpolset.go
```diff
@@ -0,0 +1,164 @@
...
+
+var (
+	log *logrus.Logger
+)
+
+func modeToString(mode uint) string {
+	// Convert mode to string representation
+	flagsStr := ""
+	for name, value := range mempolicy.Flags {
+		if mode&value != 0 {
```
klihub, line 47: Out of curiosity, can flag values have overlapping bits ? And why the != 0 test and masking out value on each iteration, instead of a == value test ?

klihub (commented):
I have a (probably stupid) minor question. But LGTM.

cmd/plugins/memory-policy/mpolset/mpolset.go
askervin: Non-flag (least significant) bits of `mode` returned by the kernel form a natural number that specifies the memory policy mode. Most significant bits are flags, each bit corresponding separate flag without overlapping bits. (In reality, most flag bit combinations are invalid, and validity of a combination depends on then natural number, but that's semantics of memory policies and does not affect this function.)

In order to separate the natural number from the bits, the bits need to be masked away somewhere. They could be masked out all at once, if we'd use a constant that includes bits of all flags or correspondingly bits available for the natural number (or if we'd hardcode all flags here). I don't really want to introduce a constant and hardcoding "all flag bits" here sounds even worse, so I prefer clearing each known flag that happens to be set.

Why comparing `mode&flag != 0` instead of `mode&flag == flag` (if this is what you asked) is just a habit with single-bit flags... I find it more readable -- and there at least were the days when comparing against zero was more efficient than comparing equality of two values. But my efficiency argument does not make much sense here, as I'm masking out flag bits one-by-one.

klihub: Okay, so semantically mode here is a mode and (potentially mode-specific) flags. I missed/misread/did not understand the final map-lookup based mode validity check (modeStr == ""). I guess I'm just more used to almost always doing an explicit key existence check (`if val, ok := tbl[key]; !ok ...`) instead of testing for zero-value of val.

Makefile
```diff
@@ -82,7 +82,8 @@ PLUGINS ?= \
         nri-sgx-epc
 
 BINARIES ?= \
-	config-manager
+	config-manager \
```
fmuyassarov, line 85: hopefully we can get rid of this soon...

klihub (approved):

PR #517, opened 2025-05-23
memory-policy: NRI plugin for setting memory policy

askervin:
nri-memory-policy plugin sets memory policy when an annotated container is getting created. The plugin supports two ways for setting memory policy. First, using MemoryPolicy NRI adjustment and in OCI spec, and second, by injecting mpolset into the container and prefixing container command with it.

The first version of the plugin configures memory policy based on pod annotations: either a class or a custom annotated policy. Memory policy classes can be defined in the plugin configuration.

### DIFF
--- a/cmd/plugins/memory-policy/main.go
+++ b/cmd/plugins/memory-policy/main.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Inter Corporation. All Rights Reserved.
+// Copyright 2025 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/plugins/memory-qos/main.go
+++ b/cmd/plugins/memory-qos/main.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Inter Corporation. All Rights Reserved.
+// Copyright 2023 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


go.mod
```diff
@@ -4,16 +4,16 @@ go 1.23.4
 
 require (
 	github.com/askervin/gofmbt v0.0.0-20250119175120-506d925f666f
-	github.com/containerd/nri v0.6.0
+	github.com/containerd/nri v0.9.1-0.20250530003506-6120e633d4ad
```
klihub, line 7: Have you tested if this has some unwanted side-effect on an otherwise unmodified existing balloons or t-a policy in a cluster which runs an stock runtime using an older version of tagged NRI ?

klihub: @askervin Thanks for the extra tests you ran. Perhaps we should then be brave enough and merge this. We can deal with any problems if it turns out we run into something unexpected.

askervin: All balloons and topology-aware n4c16 e2e tests pass on vm with containerd v2.1.1.

Tested balloons in real use cases on a bare-metal single-node cluster with containerd v1.7.24. No problems encountered.

cmd/plugins/memory-policy/Dockerfile
```diff
@@ -0,0 +1,27 @@
+ARG GO_VERSION=1.23
```
fmuyassarov: Since [this](https://github.com/containers/nri-plugins/pull/528) change, we better bump the go version here too ?

cmd/plugins/memory-policy/main.go
```diff
@@ -0,0 +1,595 @@
+// Copyright 2025 Inter Corporation. All Rights Reserved.
```
fmuyassarov:
```suggestion
// Copyright 2025 Intel Corporation. All Rights Reserved.
```

fmuyassarov (commented):
nits, otherwise LGTM, though haven't tested it.

cmd/plugins/memory-policy/Dockerfile
askervin: Good catch! Thanks @fmuyassarov!

PR #518, opened 2025-05-23
memory-policy: Dockerfile and a sample config for the plugin image

askervin:


cmd/plugins/memory-policy/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+COPY go.mod go.sum ./
+COPY pkg/topology/ pkg/topology/
+RUN go mod download
+
+# Build nri-resmgr
+COPY . .
+
+RUN make clean
+RUN make IMAGE_VERSION=${IMAGE_VERSION} BUILD_VERSION=${BUILD_VERSION} BUILD_BUILDID=${BUILD_BUILDID} PLUGINS=nri-memory-policy build-plugins-static
+RUN make -C cmd/plugins/memory-policy/mpolset
```
klihub, line 20: So far I think we tried to keep a single monolithic Makefile for all plugins, so I wonder if this extra Makefile is really necessary. I'm guessing that you want to force compiling to a static binary this way and maybe it is the most straightforward one, but I wonder if that could be done in the top level Makefile for this plugin somehow using the existing related plumbing... anyway, not a blocker.

klihub (approved):
I'm having a nit-streak, I guess... but overall LGTM.

klihub (changes requested):
This is basically a single commit PR because the other commit here is a (bit odd) merge commit. The real commit with genuine content here declares the new plugin in the Makefile, adds a sample config for it, and a Dockerfile for building images for it.

So I think this would be much better to have as part of #515, which is the original PR that implements the new plugin which every addition here is directly related to.

PR #519, opened 2025-05-23
helm: add Chart for memory-policy deployment

askervin:


klihub (approved):

deployment/helm/memory-policy/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Prerequisites
+
+- Kubernetes 1.24+
+- Helm 3.0.0+
+- Container runtime:
+  - containerD:
+    - At least [containerd 2.X](https://github.com/containerd/containerd/releases/tag/not yet released).
+  - CRI-O
+    - At least [vX.X.X](https://github.com/cri-o/cri-o/releases/tag/not yet released).
```
fmuyassarov: These links seem to be broken...

fmuyassarov: If we are asking for containerd at least 2.X version, it would make sense to drop the `patchConfig` field from this chart. Not necessarily on this PR and can be done as a follow up.

fmuyassarov (commented):
one more nit, otherwise LGTM

deployment/helm/memory-policy/README.md
klihub: @fmuyassarov Before NRI goes to 1.0/first stable version, there is no guarantee that we wouldn't hit some snafu which then caused it to be flipped off by default in the runtime. So I would wait with that until we reach a first stable release.

fmuyassarov: Fair enough

fmuyassarov: This looks great to be merged apart from broken links.

askervin: Fixed with a working link to the commit in NRI that container runtimes will need to include in order to get command injection working.

PR #522, opened 2025-05-23
e2e: fix CNI directory in recent containerd versions

askervin:
Due to containerd configuration change in recent versions, that is switching from bin_dir to bin_dirs, our configuration modification for Fedora stopped working. This patch makes the CNI location fix more robust, and it works with old and new containerd versions.


klihub (approved):

askervin:
> Did this change appear starting from containerd `v2.*` ?

Building old shims seems to be dropped in https://github.com/containerd/containerd/commit/c50a3ef0434cc1961311003e7ec6ab08d4bebdc7 that has been merged to master right after tagging v1.7.

fmuyassarov (approved):

PR #523, opened 2025-05-23
e2e: test memory-policy plugin

askervin:
This is first ansible deployment file for an NRI non-resource-policy plugin. It comes with some technical debt for refactoring duplicate tasks in deployment files with NRI resource policy plugins. This version is deliberately separate from them in order to minimize conflicts.


askervin:
Re-stacked on top of updated PR #519.

PR #524, opened 2025-05-23
e2e: helm-launch supports launching the memory-policy plugin

askervin:


test/e2e/run.sh
```diff
@@ -415,13 +419,15 @@ helm-launch() { # script API
         fi
     fi
 
-    timeout=$(timeout-for-deadline $deadline)
-    timeout=$timeout wait-config-node-status $cfgresource
+    if [[ -n "$cfgresource" ]]; then
```
klihub, line 422: nit: does this really need [[. If not I'd go with [ for better local contextual consistency with the rest of the function.

klihub (commented):
I have a nit-question, but otherwise looks good.

klihub (changes requested):
Could we please have this (together with #525) as part of #523  and retitle it 'e2e: add memory policy plugin e2e tests.' ?

test/e2e/run.sh
askervin: Fixed in #523.

PR #525, opened 2025-05-23
e2e: test the memory-policy plugin

askervin:


askervin:
Added as part of #523.

PR #533, opened 2025-06-10
sysfs: fix CPU.GetCaches() to not return empty slice.

klihub:
Fix `pkg/cache.GetCaches()` always incorrectly returning an empty slice.


marquiz (approved):
lgtm

askervin (approved):
LGTM

PR #534, opened 2025-06-10
sysfs: export CPUFreq.{Min,Max}.

klihub:
Make sysfs.CPUFreq.{Min,Max} exported members. Without this a returned CPUFreq is useless. Also, fold baseFreq into the same struct.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -241,7 +241,6 @@ type cpu struct {
 	node     idset.ID    // node id
 	core     idset.ID    // core id
 	threads  idset.IDSet // sibling/hyper-threads
-	baseFreq uint64      // CPU base frequency
 	freq     CPUFreq     // CPU frequencies
 	epp      EPP         // Energy Performance Preference from cpufreq governor
 	online   bool        // whether this CPU is online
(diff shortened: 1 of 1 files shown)


askervin (approved):
LGTM

marquiz (approved):
Looks good. Just a question do we need/want getters? What if just export the fields?

klihub:
> Looks good. Just a question do we need/want getters? What if just export the fields?

That's a clear alternative and I don't know which one is better here. I simply chose to go with pure additions instead of changes but did not really think about it much/almost at all.

klihub:
> Looks good. Just a question do we need/want getters? What if just export the fields?

@marquiz I updated it to just use exported members in CPUFreq.

marquiz (approved):
LGTM

klihub:
> > I updated it to just use exported members in CPUFreq.
> 
> I didn't have strong opinions and didn't mean to cause unnecessary extra churn because of this small detail 🙈 But prolly feels more logical this way (the CPUFreq is not a complex object, basically just carrying data so why not just export the fields).

Yeah, I agree. Much better this way for such a trivial struct. It is anyway returned by copy and not as a pointer, so having its fields exported should be totally fine.

PR #535, opened 2025-06-10
Drop tools.go in favor of native tool directive support in go 1.24

fmuyassarov:


fmuyassarov:
> @fmuyassarov I'm not against this at all, but could we delay this until we manage to get most of the other functional PR's merged ? We have a bit of a crunch going on ATM, so I'd prefer to focus on getting those merged first.

Absolutely. You can ping me here whenever you will ready for this change. Thanks.

klihub:
> > @fmuyassarov I'm not against this at all, but could we delay this until we manage to get most of the other functional PR's merged ? We have a bit of a crunch going on ATM, so I'd prefer to focus on getting those merged first.
> 
> Absolutely. You can ping me here whenever you will ready for this change. Thanks.

@fmuyassarov Then again, if those CI failures would go away by simply bumping golang to 1.24.x in go.mod, then there is no need to delay this. We can do so and merge this once it is updated.

klihub:
> > > @fmuyassarov I'm not against this at all, but could we delay this until we manage to get most of the other functional PR's merged ? We have a bit of a crunch going on ATM, so I'd prefer to focus on getting those merged first.
> > 
> > 
> > Absolutely. You can ping me here whenever you will ready for this change. Thanks.
> 
> @fmuyassarov Then again, if those CI failures would go away by simply bumping golang to 1.24.x in go.mod, then there is no need to delay this. We can do so and merge this once it is updated.

I pulled this in and it looks like the problems go away by a
```shell
go get k8s.io/client-go/tools/clientcmd@v0.31.2 && \
    go mod tidy && \
    git commit -m -s "go.{mod,sum} update and tidy deps."
```

After that `make verify-generate verify-build verify-godeps && echo PASS` passes locally.

klihub (approved):

marquiz (approved):
Thanks @fmuyassarov. The PR also updates the deps (especially k8s.io/code-generator). That's fine assuming you've tested that code-generation still works :) Suggestion: that could be mentioned in the commit message. 

LGTM from me

PR #539, opened 2025-06-16
doc: how to allow using all CPUs and memories with a balloon

askervin:


docs/resource-policy/policy/balloons.md
```diff
@@ -534,6 +534,34 @@ metadata:
     memory-type.resource-policy.nri.io/container.LLM: HBM,DRAM
 ```
 
+## Resetting CPU and memory pinning
+
+The balloons effectively clears cgroups CPU and memory pinning of all
+containers if a policy pins all containers to all CPUs and memories in
+the system. We can extend the built-in "reserved" balloon instance to
```
klihub: @askervin Hmm, I'm not sure I understand (or believe) what this states. Is this really accurate ?

askervin: Thanks @klihub. Paragraph rewritten to be clear.

klihub (approved):

PR #540, opened 2025-06-16
doc: an example on preventing creating a scheduled container

askervin:


klihub:
@askervin There is a codespell-flagged typo in the docs.

PR #541, opened 2025-06-16
resmgr,config: allow configuring cache allocation via goresctrl.

klihub:
This patch adds
- cache class/partitioning configuration (RDT)
- updates resmgr to apply RDT configuration using `goresctrl` if RDT control is enabled
- updates helm charts with a simple sample rdt control config block in the default configuration
- updates helm charts to allow access to resctrlfs if `config.control.rdt.enable == true`
- adds brief documentation about RDT control


askervin (approved):
This looks good to me.

PR #545, opened 2025-07-01
topology-aware: try picking resources by hints first

klihub:
This PR updates how topology hints are taken into account during resource allocation, especially when there are multiple devices with different HW locality allocated to a single container. In particular the PR

- gives more priority for topology hints, bringing them just below explicitly annotated affinity
- updates CPU allocation to try pick CPUs by pod resource API hints first
- updates memory allocation to pick memory nodes by pod resource API hints first
- updates the helm charts values to include agent setting defaults for enabling NRT and pod resource API

Notes: This behavioral change is currently not put behind a config option or an annotation yet. I intend to do that however, by pushing an additional commit.

### DIFF
--- a/docs/resource-policy/policy/topology-aware.md
+++ b/docs/resource-policy/policy/topology-aware.md
@@ -540,6 +540,68 @@ locality to a NUMA node is advertised by the API. Annotated allow and deny
 lists can be used to selectively disable or enable per-resource hints, using
 `podresapi:$RESOURCE_NAME` as the path for the resource.
 
+### Picking CPU And Memory By Topology Hints
+
+Normally topology hints are only used to pick the assigned pool for a workload.
+Once a pool is selected the available resources within the pool are considered
+equally good for satisfying the topology hints. When the policy is allocating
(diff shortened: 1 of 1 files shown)


klihub:
> With a very quick pass on this, looks sane to me 😅 With a potentially "might not be a good idea" I'd suggest to start with an off-by-default configuration option.

Yes, definitely. I'll put this behind an annotation so that it will be in effect only for containers annotated for it. And only take of the draft status once that commit is in place.

> 
> Spotted one typo in one commit message `With multiple devices allocated to a since container`

Thanks for spotting that! Fixed.

fmuyassarov (commented):
If NRT is now enabled by default on the agent, I wonder if we still need the Helm based gate for toggling it. Since exposing resources via NRT seems to be the expected default (at least in Helm deployments), maybe we can simplify things by removing that extra switch, unless we're still concerned about generating large CRs? Just a thought.

fmuyassarov:
but otherwise LGTM, as you already mentioned the annotation name is something that can be improved but I can't really suggest anything better since naming is always hard.

PR #551, opened 2025-08-07
rdt,resmgr: allow running RDT in discovery mode.

klihub:
Run RDT in "discovery" mode if it is enabled but without any `goresctrl`-specific configuration (IOW options or partitions). In this case we initialize goresctrl without setting any RDT configuration. This effectively puts `goresctrl` into discovery mode, where containers can still be assigned to any RDT class already present in the system, typically having been created by external means.


pkg/apis/config/v1alpha1/resmgr/control/rdt/config.go
```diff
@@ -160,6 +160,10 @@ func (c *Config) ToGoresctrl() (*rdt.Config, bool, error) {
 		return nil, false, nil
 	}
 
+	if c.Options == nil && c.Partitions == nil {
```
marquiz, line 163: Would it be more logical if just?

```
if c.Partitions == nil {
```

Moreover, in the future, the option might have something related to the "discovery mode". WDYT?

klihub: > Would it be more logical if just?
> 
> ```
> if c.Partitions == nil {
> ```

That's what I had first. The reason why I changed it is, because with that approac the user might think that if options are given without partitions, the options are still passed to goresctrl and somehow taken into account, which is not the case. That said, I am fine with changing it back (and maybe log a warning if some options are given without partitions).

> Moreover, in the future, the option might have something related to the "discovery mode". WDYT?

Yes, that can happen. And actually already for this PR one alternave I considered was to add an explicit 'DiscoveryMode bool` to Options and only run in "discovery mode" if it is set to true, because the current way of enabling it in this PR might feel a bit implicit (enable, but omit any configuration => discovery mode). Then again, with that setup it would be possible to give an inconsistent setup by setting both discovery mode and giving a partition configuration, so I did not go with that.

marquiz: K, I rely on your reasoning. LGTM

fmuyassarov (approved):

PR #552, opened 2025-08-13
docs: update balloons debugging guidance and add examples

askervin:
Fixing configuration sample that does not work.

Offering logs and metrics reading commands that work with direct copy-paste.

### DIFF
--- a/docs/resource-policy/policy/balloons.md
+++ b/docs/resource-policy/policy/balloons.md
@@ -679,6 +679,21 @@ instrumentation:
   metrics:
     enabled: # use '*' instead for all available metrics
     - policy
-logger:
-  Debug: policy
+log:
+  debug:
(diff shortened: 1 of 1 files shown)


docs/resource-policy/policy/balloons.md
```diff
@@ -679,6 +679,21 @@ instrumentation:
...
-logger:
-  Debug: policy
+log:
+  debug:
+  - policy     # to follow policy's internal logic
+  - nri-plugin # to see what the policy saw and to hear what it said (there is a lot of it)
+  source: true
+```
+
+Read metrics:
```
marquiz: Maybe a note that this only works "locally" i.e. you need direct connectivity to the pod ip(?)

askervin: Thanks @marquiz, note added.

PR #554, opened 2025-08-18
sysfs: patch up asymmetric NUMA distances.

klihub:
Hide asymmetric NUMA distances by replacing asymmetric distance pairs by their average.

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -1460,6 +1460,26 @@ func (sys *system) discoverNodes() error {
 		}
 	}
 
+	// Patch any asymmetric NUMA matrix using the average distance.
+	// Notes:
+	//   Instead of hiding asymmetricity completely, we might want to keep
+	//   the pristine distances and instead update System.NodeDistance to
+	//   return an average for asymmetric distance pairs.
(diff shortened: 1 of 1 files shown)


pkg/sysfs/system.go
```diff
@@ -1460,6 +1460,26 @@ func (sys *system) discoverNodes() error {
...
+	//   Instead of hiding asymmetricity completely, we might want to keep
+	//   the pristine distances and instead update System.NodeDistance to
+	//   return an average for asymmetric distance pairs.
+	for id1, n1 := range sys.nodes {
+		for id2, n2 := range sys.nodes {
+			if id1 == id2 {
+				continue
+			}
+			if n1.distance[id2] != n2.distance[id1] {
+				avg := int((float64(n1.distance[id2]+n2.distance[id1]) + 0.5) / 2.0)
```
askervin: Why not simply: `avg := (n1.distance[id2] + n2.distance[id1]) / 2`? If I'm not mistaken, it's equivalent...

klihub: Updated to use integer arithmetic.

PR #561, opened 2025-08-23
e2e: fix VM disk sizing, provision rootfs with max possible size.

klihub:
This PR
- fixes VM disk size setting to really take effect
- updates provisioning to resize rootfs to max. possible size

This is only done for fedora (blindly assuming cloud images, which is what we use via Vagrant and) taking into account that the fedora cloud image layout is known a priory and deliberately designed to make this straightforward. For Ubuntu (22.04LTS) this is not necessary, as this seems to be done automatically during setup. Vagrant support for Ubuntu 24.04 and later has been discontinued, so there is no possibility or need for testing it.

### DIFF
--- a/test/e2e/playbook/provision.yaml
+++ b/test/e2e/playbook/provision.yaml
@@ -145,6 +145,8 @@
             pkg:
               - iproute-tc
               - grubby
+              - cloud-utils-growpart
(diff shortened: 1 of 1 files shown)


askervin (approved):
LGTM

PR #563, opened 2025-08-26
images: save licenses in generated images.

klihub:
Add Makefile target for license extraction and save extracted licenses to built images.


Makefile
```diff
@@ -263,6 +267,35 @@ image.%:
...
+	        ;; \
+	    *.*) \
+		dir=$(patsubst licenses.%,./cmd/%,$@); \
+	        ;; \
+	esac; \
+	bin="$(patsubst licenses.%,%,$@)"; \
+	out="$(LICENSE_PATH)/$$bin"; \
+	rm -fr "$$out" && \
+	go-licenses save "$$dir" \
+	    --ignore github.com/containers/nri-plugins \
```
pfl: I'm scratching my head wondering why this directory was excluded.

.github/workflows/verify.yaml
```diff
@@ -89,3 +89,17 @@ jobs:
           git config user.name "Github"
           git config user.email "no-reply@github.com"
           ./scripts/build/update-gh-pages.sh
+
+  verify-licenses:
+    name: Verify Licenses
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
```
marquiz:
```suggestion
      - uses: actions/checkout@v5
```

Makefile
marquiz: It's not about the directory, it's about skipping the nri-plugins package itself. But should we include this as well(?)

Makefile
```diff
@@ -263,6 +267,35 @@ image.%:
...
+	    --save_path "$$out" && \
+	go-licenses report "$$dir" \
+	        --ignore github.com/containers/nri-plugins \
+	        > "$$out"/licenses.csv && \
+	cat "$$out"/licenses.csv | sed 's/^.*,//g' | sort -u > "$$out"/licenses-summary.txt
```
marquiz: suggestion: could we put this longish script as a separate script under (`scripts/`, for example)? Would make it so much easier to read and maintain. WDYT?

Makefile
```diff
@@ -423,13 +456,19 @@ verify-docs: site-build clean-html
...
+
 #
 # targets for installing dependencies
 #
 
 install-ginkgo:
 	$(Q)$(GO_INSTALL) -mod=mod github.com/onsi/ginkgo/v2/ginkgo
 
+install-go-licenses:
+	$(Q)$(GO_INSTALL) github.com/google/go-licenses@latest
```
marquiz: suggestion: pin to a specific version. Prevents possible CI failures just because something changed in go-licenses

marquiz (commented):
Thanks @klihub. A few quick comments below.

Makefile
klihub: > I'm scratching my head wondering why this directory was excluded.

Originally it was (I think) because pkg/topology lacked a license but was a go module of its own, which go-licenses could not handle. I fixed that locally, but forgot to (fully) commit it...

Makefile
klihub: Okay, I tried to create a more civilized/easier to maintain version of this as `scripts/build/extract-licenses.sh`.

.github/workflows/verify.yaml
klihub: Okay, I'll then add a bottom commit to this PR which bumps all action/checkout-instances to v5. An alternative would be to go with v4 here for consistency, and then file a separate PR for bumping all instances of actions/checkout to v5 in a single go.

Makefile
klihub: Pinned to v1.6.0 which is the latest tagged release.

Makefile
klihub: Ok, I ended up adding an ignore for `github.com/containers/nri-plugins/pkg/topology`. It looks like go-license can't handle/resolve a replace directive within the same git repo, so it ends up trying to do a git http fetch for `https://./pkg/topology` which of course fails but shows up as weird extra errors in the logs.

cmd/config-manager/Dockerfile
```diff
@@ -1,19 +1,23 @@
...
 RUN go mod download
 
-# Build the config-manager
-COPY cmd/config-manager/*.go ./
-RUN go build -tags osusergo,netgo -ldflags "-extldflags=-static" -o config-manager
+# Build config-manager binary
+COPY . .
+
+RUN make clean install-go-licenses
+RUN make IMAGE_VERSION=${IMAGE_VERSION} BUILD_VERSION=${BUILD_VERSION} BUILD_BUILDID=${BUILD_BUILDID} PLUGINS= BINARIES=config-manager build-binaries-static licenses
```
marquiz, line 16: We have complex rules for licenses.% targets but use the top-level licenses target in the dockerfiles. Unintended?

klihub: No, not unintended. Intentionally identical to images.

The seeming complexity is the case-switch which decides what and where we are dealing with. Could be reduced slightly, with a questionable tradeoff to duplicating the target and have one of them without a case at all. IMO, not a clear winner here.

PR #565, opened 2025-08-29
build: speed up image building significantly.

klihub:
**Note: this PR is stacked on top of #563.**

Use shared golang module and build docker cache mounts to speed up (especially repeated or multiple) image builds significantly.
    
Also, further unify Dockerfile to be as identical as possible. One exception to this is extra debugging collateral which only gets installed for DEBUG=1 in resource management plugin images at the moment.

Add SKIP_LICENSES[=0] to allow skipping license extraction, checking and the installation of go-licenses altogether. This comes handy to speed up image building for local or end-to-end testing, among others.


marquiz (approved):
LGTM

PR #569, opened 2025-09-01
e2e: pre-download and cache custom vagrant images.

klihub:
Some versions of vagrant seem to behave pretty badly when downloading images from custom URLs, giving up easily on any networking problems. For instance, `2.2.18` seems to be bad while `2.3.4` seems to be fine. So to avoid version dependent failures, we now download the images first using `wget`, then feed the local file to vagrant as a URL. I hope this gets us rid of the flaky behavior...

### DIFF
--- a/test/e2e/run.sh
+++ b/test/e2e/run.sh
@@ -27,10 +27,6 @@ export TEST_OUTPUT_DIR=${test_outdir:-"$OUTPUT_DIR"}
 export COMMAND_OUTPUT_DIR="$TEST_OUTPUT_DIR"/commands
 
 distro=${distro:-$DEFAULT_DISTRO}
-image_url=${distro_images[$distro]}
-if [ -n "$image_url" ]; then
-    export distro_img="$image_url"
-fi
 
 export k8scri=${k8scri:-"containerd"}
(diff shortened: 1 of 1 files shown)


marquiz (commented):
This is on the host context? If so, should we have "nri-plugins" somewhere in the cache dir name?

klihub:
> This is on the host context? If so, should we have "nri-plugins" somewhere in the cache dir name?

@marquiz We have `e2e` in the cache dir name. If that is not good/distinctive enough, please suggest something better and I'll update it...

fmuyassarov (commented):
lgtm

klihub:
> > This is on the host context? If so, should we have "nri-plugins" somewhere in the cache dir name?
> 
> @marquiz We have `e2e` in the cache dir name. If that is not good/distinctive enough, please suggest something better and I'll update it...

@marquiz Updated the cache location to be `$HOME/.cache/nri-plugins/e2e`.

klihub:
> ... I hope this gets us rid of the flaky behavior...

#571 is needed to get rid of the experienced flaky behavior, which is most prevalent on OpenSUSE but can be triggered on other distro's when trying persistently enough.

PR #570, opened 2025-09-02
topology-aware: consider burstable CPU limit when picking a pool

klihub:
This PR improves CPU resource allocation for burstable QoS class containers.

Currently CPU burstability is not taken into account when a pool is picked for a container. For containers with high burstability (large CPU limit - CPU request), this can result in an assignment which cannot have any or enough excess CPU capacity for bursting.

This PR changes that behavior. When picking a pool for a burstable QoS class container, take into account the CPU limit, too. For limited burstable containers, prefer pools with enough free capacity for the limit. For unlimited burstable containers, prefer pools at a topology level indicated by the new `unlimitedBurstable` configuration option, with a default of `package` indicating sockets. This can be overridden on a per pod or container level using the new `unlimited-burstable.resource-policy.nri.io` annotation. Unlimited containers will typically end up in a pool at the configured or annotated topology level, unless affinity, memory types, or topology hints dictate otherwise.

### DIFF
--- a/docs/resource-policy/policy/topology-aware.md
+++ b/docs/resource-policy/policy/topology-aware.md
@@ -145,6 +145,11 @@ behavior. These options can be supplied as part of the effective
     `normal`, `low`, and `none`. Currently this option only affects exclusive
     CPU allocations. For a more detailed discussion of CPU prioritization see
     the [cpu allocator](../developers-guide/cpu-allocator.md) documentation.
+- `unlimitedBurstable`
+  - is the default topology level preference for containers with unlimited
+    burstability. The policy will try to allocate Burstable containers with
+    no CPU limit to a pool at this topology level. The possible values are:
+    `system`, `package`, `die`, `numa`.
 
 Additionally, the following sub-configuration is available for instrumentation:
 
@@ -280,6 +285,29 @@ metadata:
 
 These Pod annotations have no effect on containers which are not eligible for
 exclusive allocation.
+
+### Preferred Topology Level for Burstable Containers Without CPU Limit
+
+CPU-unlimited burstable containers are by default preferred to allocate to a
+pool at the topology level specified by the `unlimitedBurstable` configuration
+option. This global default can be overridden by pod or container using the
+`unlimited-burstable.resource-policy.nri.io` annotation. This annotation can
+have the same values as the `unlimitedBurstable` configuration option.
+
+```yaml
+metadata:
+  annotations:
+    # prefer to allocate container within the pod by default to a die
+    unlimited-burstable.resource-policy.nri.io/pod: "die"
+    # prefer to allocate C1 to a single NUMA node
+    unlimited-burstable.resource-policy.nri.io/container.C1: "numa"
+    # prefer to allocate C2 to a single socket
+    unlimited-burstable.resource-policy.nri.io/container.C2: "package"
+    # prefer to allocate C3 to all sockets in the system
+    unlimited-burstable.resource-policy.nri.io/container.C3: "system"
+    # any other containers in the pod will prefer allocation to a single die
+```
+
 
 ### Selectively Disabling Hyperthreading
 


klihub:
@askervin @fmuyassarov PTAL

klihub:
@askervin @fmuyassarov I added configurability for what unlimited burstability means. There is a global configuration option and a per-pod and container annotation to override the global default. I think this would be now ready for review.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -381,52 +386,68 @@ func cpuAllocationPreferences(pod cache.Pod, container cache.Container) (int, in
...
 	prio := defaultPrio // ignored for fractional allocations
+	limit := 0
+
+	switch qosClass {
+	case corev1.PodQOSBestEffort:
+	case corev1.PodQOSBurstable:
+		if lim, ok := reqs.Limits[corev1.ResourceCPU]; ok {
+			limit = int(lim.MilliValue())
+		} else {
+			limit = unlimitedCPU
```
askervin, line 397: This suggests that limit for BestEffort pods is unlimited, but on line 417 it seems that cpuAllocationPreferences() returns limit==0 for BestEfforts anyway. If this is the case, perhaps it would be clearer to assign/keep limit=0 here.

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -673,7 +674,9 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
 	//   - if we have topology hints
 	//       * better hint score wins
 	//       * for a tie, prefer the lower node then the smaller id
-	//   - if we have a better matching or tighter fitting memory offer, it wins
+	//   - if we have a better matching memory offer, it wins
+	//   - if we have a burstable container, sufficient capacity for the limit wins
+	//   - if we have or tighter fitting memory offer, it wins
```
askervin, line 679: s/have or/have a

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -809,22 +812,99 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
...
+
+		if limit != unlimitedCPU {
+			// prefer pool with enough burstable capacity
+			switch {
+			case r1 >= 0 && r2 < 0:
+				log.Debug("  - %s loses on insufficient CPU burstability (%d vs. %d for limit %d)",
+					node2.Name(), b1, b2, limit)
+				return true
+			case r2 >= 0 && r1 < 0:
+				log.Debug("  - %s loses on insufficient CPU burstability", node1.Name())
```
askervin, line 837: Symmetric debug logs would be nice here. I'd prefer the one with more details, that is, "<node2> loses...".

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -809,22 +812,99 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
...
+				if b1 > b2 {
+					log.Debug("  - %s WINS on more CPU burstability", node1.Name())
+					return true
+				}
+				if b2 > b1 {
+					log.Debug("  - %s WINS on more CPU burstability", node2.Name())
+					return false
+				}
+				log.Debug("  - CPU burstability is a TIE")
+				return id1 < id2
```
askervin, line 874: Shouldn't we continue to next comparisons if CPU burstability is a tie, instead of just returning comparison based on ids? Time for this could be only on line 1054.

That said, we do have "return id1 < id2" in following lines before line 1054, too, making me wonder if it would be sometimes good to cut-off comparisons early if they do not look promising in any case. Anyway, continuing to pool memory comparisons for Burstable containers, in case of tied pools, would make sense to me.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -145,6 +145,11 @@ behavior. These options can be supplied as part of the effective
     `normal`, `low`, and `none`. Currently this option only affects exclusive
     CPU allocations. For a more detailed discussion of CPU prioritization see
     the [cpu allocator](../developers-guide/cpu-allocator.md) documentation.
+- `unlimitedBurstable`
+  - is the default topology level preference for containers with unlimited
+    burstability. The policy will try to allocate Burstable containers with
+    no CPU limit to a pool at this topology level. The possible values are:
+    `system`, `package`, `die`, `numa`.
```
askervin, line 152: For a very little while I was thinking should we enable l2cache as a bursting level, too, but I think that can wait for the future.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Uhmmm, why ? In a golang switch, a break is implied between cases and a fallthrough needs to be explicit, the opposite as in C. So for this to suggest limit for `BestEffort` pods to be unlimited, it'd need to be

```golang
	// use an explicit fallthrough
...
	switch qosClass {
	case corev1.PodQOSBestEffort:
		fallthrough // handle BestEffort and Burstable usingby the same chunk of code
	case corev1.PodQOSBurstable:
...
	// or use a multi-value case
	switch qosClass {
	case corev1.PodQOSBestEffort, corev1.PodQOSBurstable:
...
```

cmd/plugins/topology-aware/policy/pools.go
klihub: Hmm, yes we probably should continue evaluation instead short-circuiting based on the ID.

cmd/plugins/topology-aware/policy/pod-preferences.go
askervin: Oh god. Sorry about that. I was indeed reviewing Go like C. 🙀

docs/resource-policy/policy/topology-aware.md
klihub: In T-A, we don't have tree nodes for LLC/L2C clusters at the moment. Instead, we let the common cpuallocator handle cache-clustered allocation. As a side effect, we don't do shared cache-clustered allocation ATM, so an l2cache topology level for unlimited burstability would not make sense here for now.

cmd/plugins/topology-aware/policy/pools.go
klihub: Update logs to be symmetric.

cmd/plugins/topology-aware/policy/pools.go
klihub: Fixed.

cmd/plugins/topology-aware/policy/pools.go
klihub: Updated accordingly.

klihub:
> > I was also thinking about adding a mention of the (current) default to the documentation, because it makes sense to document it somewhere. The counter-argument for putting it into the documentation would be that if we ever to decide changing it then there is a danger of stale/incorrect documentation, if we forget to update them.
> 
> What would you think about a solution somewhere in the middle ground, that is, "make generate" or something would actually generate `docs/resource-policy/policy/topology-aware-defaults.md` that we could refer to in topology-aware.md? (And probably the same for balloons....)

If we want to document all the configuration defaults in one place, I think we would need something like that. I'm just not sure if there is enough existing tooling to do it without actually installing the CRD on a cluster, creating a CR instance, then reading out how it came out to be...

PR #571, opened 2025-09-02
e2e: adjust cloud image URLs, allow image override, fix ssh connectivity problems.

klihub:
This PR fixes a number recent/remaining problems with vagrant images in e2e tests. In particular,

- adjusts fat-fingered or otherwise wrong cloud image URLs
- allows overriding distro images (for instance to local caches for environments behind proxies)
- fixes post-provisioning ssh connectivity problems with Fedora >= 41 images
- changes default distro consistently to fedora/42.

### DIFF
--- a/test/e2e/playbook/provision.yaml
+++ b/test/e2e/playbook/provision.yaml
@@ -113,6 +113,12 @@
         path: /tmp/get_helm.sh
         state: absent
 
+    - name: Install ansible.builtin.dnf dependencies
+      ansible.builtin.shell: |
+        dnf install -y python3-libdnf5
+      when: (ansible_facts['distribution'] == "Fedora") and
+            (ansible_facts['distribution_major_version'] | int == 41)
+
     - name: Install common packages
       ansible.builtin.package:
         name:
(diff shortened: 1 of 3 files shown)


test/e2e/playbook/provision.yaml
```diff
@@ -113,6 +128,12 @@
         path: /tmp/get_helm.sh
         state: absent
 
+    - name: Install ansible.builtin.dnf dependencies
+      ansible.builtin.shell: |
+        dnf install -y python3-libdnf5
+      when: (ansible_facts['distribution'] == "Fedora") and
+            (ansible_facts['distribution_major_version'] | int == 41)
```
marquiz, line 135: Is this only for Fedora 41 (not 42 and later?)

klihub: Only 41, not needed from 42 on, neither before 41.

fmuyassarov, line 135: Don't we need python3-libdnf5 for fedora 42?

klihub: This is the same resolved question @marquiz asked. No, it needs to be explicitly installed for Fedora 41 only. It is not needed before 41 and it is part of the base distro image from 42 on.

fmuyassarov: I see. I missed that conversation then.

fmuyassarov (approved):

PR #576, opened 2025-09-10
e2e: always configure systemd cgroup driver for containerd.

klihub:
Containerd 2.1.x omits the cgroup driver default setting (`SystemdCgroup = false`) from the config file altogether. Our current e2e test provisioning therefore fails to flip the setting to on. Use a bigger hammer to always force it on.

### DIFF
--- a/test/e2e/playbook/provision.yaml
+++ b/test/e2e/playbook/provision.yaml
@@ -339,13 +339,34 @@
     - name: Configure containerd
       when: is_containerd
       block:
-        - name: Create containerd configuration
-          ansible.builtin.shell: "{{ item }}"
-          with_items:
-            - mkdir -p /etc/containerd
-            - containerd config default > /etc/containerd/config.toml
-            - sed -i 's/^.*disabled_plugins *= *.*$/disabled_plugins = []/' /etc/containerd/config.toml
-            - sed -i 's/SystemdCgroup = false/SystemdCgroup = true/g' /etc/containerd/config.toml
+        - name: Create drop-in directory (/etc/containerd/conf.d)
+          ansible.builtin.file:
+            path: /etc/containerd/conf.d
+            state: directory
+
+        - name: Generate default configuration (/etc/containerd/config.toml)
+          ansible.builtin.shell: containerd config default > /etc/containerd/config.toml
+
+        - name: Enable drop-in directory in configuration
+          ansible.builtin.lineinfile:
+            path: /etc/containerd/config.toml
+            regexp: '^imports *=.*'
+            line: 'imports = [ "/etc/containerd/conf.d/*.toml" ]'
+            state: present
+
+        - name: Enable all plugins
+          ansible.builtin.lineinfile:
+            path: /etc/containerd/config.toml
+            regexp: '^disabled_plugins *=.*'
+            line: 'disabled_plugins = []'
+            state: present
+
+        - name: Enable systemd cgroup driver
+          ansible.builtin.copy:
+            dest: /etc/containerd/conf.d/systemd-cgroup.toml
+            content: |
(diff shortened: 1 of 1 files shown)


marquiz (approved):
LGTM

That missing from containerd is a bug? Sounds strange that it doesn't show the config as part of default config (which I thought displays all possible knobs) but then obeys it if somebody sets it.

marquiz:
> Fix merged but will take a while to hit the releases: [containerd/containerd#12244](https://github.com/containerd/containerd/pull/12244)

Should we put a note around our hack so that we don't so easily forget what we did and why (and possibly remove the hack when it's fixed upstream)? WDYT?

klihub:
> > Fix merged but will take a while to hit the releases: [containerd/containerd#12244](https://github.com/containerd/containerd/pull/12244)
> 
> Should we put a note around our hack so that we don't so easily forget what we did and why (and possibly remove the hack when it's fixed upstream)? WDYT?

We can. I accidentally spotted something which I think would be nicer alternative, both to this workaround and to the original sediting. It would be something like this:

```yaml
...
    - name: Configure containerd
      when: is_containerd
      block:
        - name: Create containerd configuration
          ansible.builtin.shell: |
            mkdir -p /etc/containerd
            containerd config default > /etc/containerd/config.toml
            sed -i 's/^.*disabled_plugins *= *.*$/disabled_plugins = []/' /etc/containerd/config.toml
            sed -i 's|^imports *=.*$|imports = [ "/etc/containerd/conf.d/*.toml" ]|' /etc/containerd/config.toml
            cat <<EOF > /etc/containerd/conf.d/systemd-cgroup.toml
            [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc.options]
            SystemdCgroup = true
            EOF
...
```

askervin (approved):
LGTM

test/e2e/playbook/provision.yaml
```diff
@@ -340,12 +340,22 @@
...
+            sed -i 's|^imports *=.*$|imports = [ "/etc/containerd/conf.d/*.toml" ]|' $CONFIG
+            cat <<EOF > /etc/containerd/conf.d/enable-all-plugins.toml
+            disabled_plugins = []
+            EOF
+            cat <<EOF > /etc/containerd/conf.d/systemd-cgroup.toml
+            [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc.options]
+            SystemdCgroup = true
+            EOF
+          register: containerd_configuration
+          failed_when: containerd_configuration.rc != 0
```
marquiz: Taking inspiration from the ansible.builtin.replace below...

Now that we got into this, I wonder would it be possible to make this a bit more ansible'ish, splitting the operations into small more readable/maintainable chunks, like


```yaml
- name: Create /etc/containerd/conf.d/ 
  ansible.builtin.file:
    path: /etc/containerd/conf.d
    state: directory

- name: Generate containerd default configuration
  ansible.builtin.shell: containerd config default > /etc/containerd/config.toml

- name: Enable all containerd plugins
  ansible.builtin.lineinfile:
    path: /etc/containerd/config.toml
    regexp: '^disabled_plugins'
    state: absent

- name: Set imports in containerd config
  ansible.builtin.lineinfile:
    path: /etc/containerd/config.toml
    regexp: '^imports ='
    line: 'imports = [ "/etc/containerd/conf.d/*.toml"]'
    state: present

- name: Create containerd config snippet
  ansible.builtin.copy:
    dest: /etc/containerd/conf.d/systemd-cgroup.toml
    content: |
      [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc.options]
      SystemdCgroup = true
```

WDYT?

klihub: Updated.

askervin:
Thanks for clarifications!

PR #578, opened 2025-09-12
e2e: alternate test runtimes in tests.

klihub:
- alternate the CRI runtime tests are being run on every other day
- only build the full set of binaries and images once per week, otherwise the minimal set we're going to test


askervin (approved):
LGTM

PR #579, opened 2025-09-19
Balloons: enabling disabling c-states in latency-critical balloons

askervin:


askervin:
Depends on https://github.com/intel/goresctrl/pull/158

klihub:
Haven't really looked at this, but as a nit you seem to lack a sign-off.

askervin:
> Haven't really looked at this, but as a nit you seem to lack a sign-off.

@klihub, this is the first time ever when I did it by purpose... It's the HACK commit that switches to use goresctrl from askervin's github branch with cstates support in it. I did not want to get that commit merged even accidently, so I thought that one way to get red flags would be leaving it unsigned. I'll fix it when the goresctrl PR is merged.

askervin:
@marquiz, @klihub , balloons c-states is now on top of @marquiz's c-states filter API. This PR is now ready for review.

pkg/apis/config/v1alpha1/resmgr/control/cpu/config.go
```diff
@@ -32,4 +32,19 @@ type Class struct {
 	UncoreMaxFreq uint `json:"uncoreMaxFreq,omitempty"`
 	// CPUFreq Governor for this class.
 	FreqGovernor string `json:"freqGovernor,omitempty"`
+	// DisabledCstates lists C-states disabled for this class.
+	DisabledCstates []string `json:"disabledCstates,omitempty"`
+}
+
+// DeepCopy makes a deep copy of the Class.
```
klihub: Since we're generating code here all over the map anyway, and since AFAICT there is nothing tricky about `Class` here wrt. a `Class.DeepCopy()`, I'd suggest to annotate it for `DeepCopy`-generation and let the toolchain take care of it instead of manually writing it.

```golang
// +k8s:deepcopy-gen=true
type Class struct {
	// MinFreq is the minimum frequency for this class.
	MinFreq uint `json:"minFreq,omitempty"`
	// MaxFreq is the maximum frequency for this class.
	MaxFreq uint `json:"maxFreq,omitempty"`
	// EnergyPerformancePreference for CPUs in this class.
	EnergyPerformancePreference uint `json:"energyPerformancePreference,omitempty"`
	// UncoreMinFreq is the minimum uncore frequency for this class.
	UncoreMinFreq uint `json:"uncoreMinFreq,omitempty"`
	// UncoreMaxFreq is the maximum uncore frequency for this class.
	UncoreMaxFreq uint `json:"uncoreMaxFreq,omitempty"`
	// CPUFreq Governor for this class.
	FreqGovernor string `json:"freqGovernor,omitempty"`
	// DisabledCstates lists C-states disabled for this class.
	DisabledCstates []string `json:"disabledCstates,omitempty"`
}
```

It should result in this being generated as part of `pkg/apis/config/v1alpha1/resmgr/control/cpu/zz_generated.deepcopy.go`:

```shell
$ git diff pkg/apis/config/v1alpha1/resmgr/control/cpu
diff --git a/pkg/apis/config/v1alpha1/resmgr/control/cpu/zz_generated.deepcopy.go b/pkg/apis/config/v1alpha1/resmgr/control/cpu/zz_generated.deepcopy.go
index d38afcb0..5dfb1b1a 100644
--- a/pkg/apis/config/v1alpha1/resmgr/control/cpu/zz_generated.deepcopy.go
+++ b/pkg/apis/config/v1alpha1/resmgr/control/cpu/zz_generated.deepcopy.go
@@ -20,6 +20,26 @@ package cpu
 
 import ()
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *Class) DeepCopyInto(out *Class) {
+       *out = *in
+       if in.DisabledCstates != nil {
+               in, out := &in.DisabledCstates, &out.DisabledCstates
+               *out = make([]string, len(*in))
+               copy(*out, *in)
+       }
+}
+
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new Class.
+func (in *Class) DeepCopy() *Class {
+       if in == nil {
+               return nil
+       }
+       out := new(Class)
+       in.DeepCopyInto(out)
+       return out
+}
+
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *Config) DeepCopyInto(out *Config) {
        *out = *in
```

askervin: Thanks @klihub! Fixed as you suggested, works fine.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -153,6 +155,43 @@ func (ctl *cpuctl) enforceCpufreq(class string, cpus ...int) error {
 	return nil
 }
 
+// enforceStates enforces a class-specific C-state configuration to a cpuset
```
marquiz: nit:
```suggestion
// enforceCStates enforces a class-specific C-state configuration to a cpuset
```

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -251,6 +290,7 @@ func (ctl *cpuctl) configure(cfg *cfgapi.Config) error {
 	log.Debug("applying cpu controller configuration:\n%s", utils.DumpJSON(ctl.classes))
 
 	// Sanity check
+	cstatesNeeded := map[string]struct{}{}
```
marquiz: Wouldn't a simple bool be enough? Where do we need the map?

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -260,6 +300,21 @@ func (ctl *cpuctl) configure(cfg *cfgapi.Config) error {
...
 		}
+		for _, cstate := range conf.DisabledCstates {
+			cstatesNeeded[cstate] = struct{}{}
+		}
+	}
+	if len(cstatesNeeded) != 0 {
+		var err error
+		filter := cstates.NewBasicFilter().SetAttributes(cstates.AttrDisable)
+		if cstatesEnvOverridesJson != "" {
+			ctl.cstates, err = NewCstatesFromOverride(filter)
```
marquiz, line 315: I suppose this is mean only for (e2e) testing? Maybe a comment/note clarifying this?

pkg/resmgr/control/cpu/cpu.go
askervin: Thanks, sharp-eye! Fixed to match the function name.

I also considered if it would be better to switch using *CStates instead of *Cstates, but ended up leaving capitalization as is at the moment. (If switching from xxxCstates to xxxCStates, should variable "cstates*" be renamed to "cStates*", too?)

If you wish, I can change camel case in another update to this PR.

pkg/resmgr/control/cpu/cpu.go
askervin: Definitely. Fixed.

I spotted yet another map-xxx-struct{} also in cstates override fs, and fixed that to map-xxx-bool, too.

pkg/resmgr/control/cpu/cpu.go
askervin: Yes, it's much better when clearly stated that this is not part of the production control flow. Fixed.

PR #580, opened 2025-10-03
scripts: add sleep-accuracy measurement tool

askervin:


askervin:
@klihub, I think you're right. Marked this now ready for review. Merging it will simplify getting numbers from #618, and help others to do the same.

scripts/testing/sleep-accuracy/sleep-accuracy.c
```diff
@@ -0,0 +1,571 @@
...
+}
+
+// set_cpuidle_minmax - enable/disable cpuidle/stateX's
+void set_cpuidle_minmax(int cpu, int min, int max) {
+    char disable_filename[1024];
+    int state = 0;
+    FILE *f = NULL;
+    while (1) {
+        sprintf(disable_filename, "/sys/devices/system/cpu/cpu%d/cpuidle/state%d/disable", cpu, state);
+        FILE *f = fopen(disable_filename, "w");
```
klihub, line 167: nit: Since we're _GNU_SOURCE, I think we could avoid buffered I/O altogether and just use `open` and `dprintf` here.

scripts/testing/sleep-accuracy/sleep-accuracy.c
```diff
@@ -0,0 +1,571 @@
...
+    }
+}
+
+// set_cpufreq_minmax - set min and max cpufreq for the CPU in sysfs
+void set_cpufreq_minmax(int cpu, int min, int max) {
+    char freq_filename[1024];
+    FILE *f = NULL;
+
+    sprintf(freq_filename, "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_max_freq", cpu);
+    f = fopen(freq_filename, "w");
```
klihub, line 213: nit: ditto here... we could avoid buffered I/O with _GNU_SOURCE.

scripts/testing/sleep-accuracy/sleep-accuracy.c
```diff
@@ -0,0 +1,571 @@
...
+        }
+        if (toggle_cpu1 != -1) {
+            set_cpu_affinity(toggle_cpu1);
+            delay(toggle_cpu_interval_ns);
+        } else {
+            // Wait until there is more than single CPU to toggle again.
+            // This prevents keep setting main thread's CPU affinity to same CPU in a loop.
+            while (toggle_cpu0 != -1) {
+                delay(toggle_cpu_interval_ns);
+            }
```
klihub: Is the idea here that if we only have a single CPU (cpu0) then at the end of the first iteration, we stay looping here forever. So basically doing the same as we would do without having this extra while, but avoiding the now unnecessary set_cpu_affinity() ?

scripts/testing/sleep-accuracy/sleep-accuracy.c
```diff
@@ -0,0 +1,571 @@
...
+            char *token = strtok(argv[++i], ",");
+            while (token && options.cpu_count < MAX_COMB) {
+                char *slash = strchr(token, '/');
+                if (slash) {
+                    *slash = '\0';
+                    options.cpus[options.cpu_count][0] = atoi(token);
+                    options.cpus[options.cpu_count++][1] = atoi(slash + 1);
+                } else {
+                    options.cpus[options.cpu_count++][0] = atoi(token);
+                    options.cpus[options.cpu_count - 1][1] = -1; // indicate single CPU pinning
```
pfl, line 806: This looks a bit confusing. Isn't the intention to set cpus[options.cpu_count][0] to the first part to value A and cpus[options.cpu_count][1] to either the value following '/' or -1 if none? So the latter could also be [options.cpu_count][0] = atoi(token) and [options.cpu_count++][1] = -1? BTW, should it also be checked that atoi() returns useful values and not perhaps -42, and would strtoul() be a better choice?

scripts/testing/sleep-accuracy/sleep-accuracy.c
```diff
@@ -0,0 +1,571 @@
...
+                }
+                token = strtok(NULL, ",");
+            }
+        } else if (strcmp(argv[i], "-p") == 0 && i + 1 < argc) {
+            options.polprio_count = 0; // Reset defaults
+            char *token = strtok(argv[++i], ",");
+            while (token && options.polprio_count < MAX_COMB) {
+                char *slash = strchr(token, '/');
+                if (slash) {
```
pfl, line 815: If there is no slash, the option is silenty ignored?

scripts/testing/sleep-accuracy/sleep-accuracy.c
```diff
@@ -0,0 +1,1010 @@
...
+        fclose(f);
+    } else {
+        perror("cannot open for writing: cpufreq/scaling_max_freq");
+    }
+
+    sprintf(freq_filename, "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_min_freq", cpu);
+    f = fopen(freq_filename, "w");
+    if (f) {
+        fprintf(f, "%d\n", min);
+        fflush(f);
```
ozhuraki, line 227: Matter of taste: setlinebuf() might save on fflush()-es.

ozhuraki (approved):
LGTM

PR #581, opened 2025-10-15
e2e: EFI BIOS support for e2e test VMs.

klihub:
Add EFI BIOS support for our e2e test VMs.
- use `efi=/usr/share/OVMF`, if `efi=1` is set
- create/boot VM using `$efi/OVMF_{CODE,VARS}[_4M].fd` if `$efi` is set
- dump extra `vm-setup()` info, if `vagrant_debug=1` is set

### DIFF
--- a/test/e2e/lib/vm.bash
+++ b/test/e2e/lib/vm.bash
@@ -100,6 +100,7 @@ vm-setup() {
     local files="$nri_resource_policy_src/test/e2e/files"
     local distro_name=$(printf '%s\n' "$distro" | sed -e 's/[\/&]/\\&/g')
     local qemu_dir="${qemu_dir:-/usr/share/qemu}"
+    local efi_code efi_vars kind
 
     mkdir -p "$inventory"
     if [ ! -f "$inventory/vagrant.ini" ]; then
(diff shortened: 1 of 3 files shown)


test/e2e/lib/vm.bash
```diff
@@ -117,9 +117,26 @@ vm-setup() {
...
 
+    if [ "$efi" = "1" ]; then
+        if [ ! -f "$vagrantdir/OVMF_CODE.fd" -o ! -f "$vagrantdir/OVMF_VARS.fd" ] ; then
+            if [ ! -e /usr/share/OVMF/OVMF_CODE.fd -o ! -e /usr/share/OVMF/OVMF_VARS.fd ]; then
+                error "EFI requested but OVMF files not found in /usr/share/OVMF"
+            fi
+
+            echo "copying OVMF files to $vagrantdir..."
+            rm -f "$vagrantdir/OVMF_*.fd"
+            cp /usr/share/OVMF/OVMF_CODE.fd $vagrantdir/OVMF_CODE.fd || \
```
askervin: New deb packages have named EFI files as OVMF_{CODE,VARS}_4M.fd
https://packages.debian.org/sid/all/ovmf/filelist

Do you think we could fallback trying to use those from /usr/share (and just drop _4M from the name when copied to $vagrantdir)? Or should we even allow user define full filepaths where to find them, if they are located elsewhere in some other distribution or users' own buids... (probably not necessary to go that far in the first version)...

klihub: Yes, I we definitely should. Just haven't realized they made that change.

klihub: Okay, updated so that
- we try OVMF_{CODE,VARS}{,_4M}.fd
- use /usr/share/OVMF as the EFI BIOS directory if `efi=1` is set
- use the given `$efi` directory if it is set to any other absolute path

test/e2e/lib/vm.bash
```diff
@@ -117,9 +118,39 @@ vm-setup() {
     EXTRA_ARGS=$(echo $VM_QEMU_CPUMEM | sed 's/MACHINE:.*CPU:.*MEM:.*EXTRA:\([^|]*\).*/\1/g')
     EXTRA_ARGS+="${EXTRA_ARGS:+,} \"-monitor\", \"unix:monitor.sock,server,nowait\""
 
+    case $efi in
+        "") ;;
+        1) efi=/usr/share/OVMF;;
+        /*) ;;
+        *) error "invalid efi value: $efi, should be 1 or absolute path to OVMF";;
```
askervin, line 125: Before this case: $efi is 0, 1 or a path.

After this case: non-empty $efi is a path, empty $efi means no EFI.

So probably adding one more case, `0) efi="";;` would be good to prevent $efi==0 causing the error on line 125.

askervin: Otherwise this looks great! Elegant solution to non-elegant problem.

klihub: Yes, I just just noticed the same and decided to fix the inconsistent efi=0 default to efi=...

askervin (approved):
Looks great, thanks @klihub!

Footnote: switching the order of the first two cases to:
```
 case $efi in
        1) efi=/usr/share/OVMF;;
        "") ;;
        /*) ;;
        *) error "invalid efi value: $efi, should be 1 or absolute path to OVMF";;
    esac
```
would give us 14 consecutive characters from the set `"*/);` (not counting whitespace), which I find somewhat bashilarious.

PR #583, opened 2025-10-16
e2e: cache and reuse release tarballs.

klihub:
This PR
- adds global caching of discovered variables
- implements caching and reuse of release tarballs
- switches Helm installation to use release tarballs for caching

### DIFF
--- a/test/e2e/files/Vagrantfile.in
+++ b/test/e2e/files/Vagrantfile.in
@@ -17,6 +17,7 @@ N = 1
 
 K8S_RELEASE = "#{ENV['k8s_release']}"
 K8S_VERSION = "#{ENV['k8s_version']}"
+HELM_RELEASE = "#{ENV['helm_release']}"
 CRI_RUNTIME = "#{ENV['k8scri']}"
 CRIO_RELEASE = "1.28.1"
 CRIO_SRC = ""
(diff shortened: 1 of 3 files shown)


klihub:
> A tangentially related note on caching... a couple of days ago I needed faster feedback loop with less network traffic, I hacked caching into local registry and http proxies that are running in containers on the host:
> 
> [intel/memtierd@c40b062](https://github.com/intel/memtierd/commit/c40b0624838a11e240b49285ba88bfc86073e82e)
> 
> However, so far that has been only hacking, and needs improvements in host directory naming, port numbering, and especially testing in environments where the host would be behind other proxies.
> 
> One problem that I already ran into with these proxies running is that when I close my laptop and wake it up, proxy services do not answer anymore to e2e vms, and I need to stop and restart them.

This (HTTP proxy) is exactly what I had in mind for the next thing in speeding up repeated tests. Not necessarily in a container, but as an IP address end point to point to, so we'd set up the VMs to use. Then we could provide a separate script to set up the proxies, for instance in a container. But if you happened to already have one, for instance in your home network, then you could simply point the VMs at the already existing one.

Once I have some more time, I'll take a look at what you cooked up in that commit and see if I could easily lift it over.

fmuyassarov (approved):
LGTM

klihub:
> I wonder if [force](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-force) parameter could be somehow helpful for caching...
> 

Yeah, good point. It looks like we could avoid the explicit cache check and let get_url realize we already have the file locally, if force is (left at the default) false. I'll check and update the PR if everything works as expected

PR #589, opened 2025-11-14
e2e: fix helm install after helm v4.0 cli flag rename

askervin:
Helm 4 renamed --atomic to --rollback-on-failure.

https://helm.sh/docs/overview/#cli-flags-renamed


klihub (approved):
LGTM

PR #590, opened 2025-11-14
balloons: allow allocation of online CPUs only

askervin:
Fixes balloons NRT report and internal behavior when user-specified allowed resources cpuset contains CPUs that are not online. Previously used difference with OfflineCPUs() does not work in this context anymore because it contains only present CPUs not online, while user-specified set may extend even beyond all possible CPUs.

### DIFF
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -1532,10 +1532,11 @@ func (p *balloons) setConfig(bpoptions *BalloonsOptions) error {
 	case cfgapi.AmountQuantity:
 		return balloonsError("can't handle CPU resources given as resource.Quantity (%v)", amount)
 	case cfgapi.AmountAbsent:
-		// Available CPUs not specified, default to all on-line CPUs.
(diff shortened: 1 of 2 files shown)


cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1532,10 +1532,11 @@ func (p *balloons) setConfig(bpoptions *BalloonsOptions) error {
 	case cfgapi.AmountQuantity:
 		return balloonsError("can't handle CPU resources given as resource.Quantity (%v)", amount)
 	case cfgapi.AmountAbsent:
-		// Available CPUs not specified, default to all on-line CPUs.
-		availableCpus = p.options.System.CPUSet().Difference(p.options.System.Offlined())
```
klihub, line 1536: I think/looks like we have the same problem in the topology-aware policy, too. Could you push a fix for that one as well to this PR ?

askervin: Great point! Done.

klihub (approved):
LGTM.

PR #587, opened 2025-11-13
e2e: enable creating VMs up to 4096 CPUs and CPU hot-plugging/removing from test scripts

askervin:

PR #591, opened 2025-11-14
e2e: add 8-socket 4k-CPU e2e vm and related test for balloons policy

askervin:
Depends on
- #587
- #590
  to pass the test.

Best reviewed in separate commits. The first commit only refactors (move and rename starting with "nrt-") NRT verification functions from NRT test to a common source file. These functions are then used by the original test script and the test added here.

### DIFF
--- a/test/e2e/policies.test-suite/balloons/s8c4k/test01-sparse-4kcpus/code.var.sh
+++ b/test/e2e/policies.test-suite/balloons/s8c4k/test01-sparse-4kcpus/code.var.sh
@@ -1,0 +1,59 @@
+# Prepare virtual machine before installing balloons.
+min_kernel_version=6.14
+vm-command "uname -r"
+if [ "$( ( echo $min_kernel_version; echo $COMMAND_OUTPUT ) | sort --version-sort | tail -n 1 )" == "$min_kernel_version" ]; then
+    error "quest OS runs too old kernel, hot-plugged CPU node topology may not work. Required: $min_kernel_version"
+fi
+
+# Hot-plug CPUs.
+vm-command 'grep 511,1535,4095 /sys/devices/system/cpu/enabled' || {
+    vm-cpu-hotplug 0 511 0
+    vm-cpu-hotplug 2 511 0
+    vm-cpu-hotplug 7 511 0
+
+    # Wait for the kernel to expose all hot-plugged CPUs in sysfs.
+    vm-run-until '[ -d /sys/devices/system/cpu/cpu511 ] && [ -d /sys/devices/system/cpu/cpu1535 ] && [ -d /sys/devices/system/cpu/cpu4095 ]'
+
+    # Online all CPUs.
+    vm-command 'for cpuX in /sys/devices/system/cpu/cpu[1-9]*; do
+            echo onlining $cpuX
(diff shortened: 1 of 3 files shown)


test/e2e/policies.test-suite/balloons/s8c4k/test01-sparse-4kcpus/code.var.sh
```diff
@@ -0,0 +1,61 @@
...
+    # Restart kubelet to let it detect new enabled CPUs.
+    vm-command "systemctl restart kubelet"
+}
+
+# Wait until kubelet has reported all enabled CPUs in node capacity.
+vm-run-until 'kubectl get node -o jsonpath="{.items[0].status.capacity.cpu}" | grep 6' ||
+    command-error "Unexpected node CPU capacity"
+
+# Make sure that k8s root cpuset.cpus contains hot-plugged CPUs.
+# containerd:  kubepods/cpuset.cpus
```
klihub: Hmm, this looks off for containerd. It should not happen after commit 3116747f534fc182ea40ac67146355e235cac340. If it does the containerd still get sometimes incorrectly set up in the e2e tests with cgroupfs instead of systemd for cgroup control.

askervin: Do you think it would be still reasonable to keep the cpuset.cpus check, as it may save a lot of time from debugging if there is an issue? I mean, it would indicate a problem in the system rather than in our plugins and/or OCI runtime. This would enable failing fast rather than next phase, that is verifying that containers do not have proper CPU affinity. At next phase we might start debugging wrong stack.

The spirit of these checks is the same as in the sysfs checks in the beginning. There it was about kernel actually finding all CPUs that where hot-plugged, and here about actually getting them to kubernetes to use. Our responsibilities start only after that's clear.

I'm sort of creating a minefield here for regressions like there was in runc v1.3.1 onwards, where suddenly CPUs beyond 1024 just disappeared yet they worked just fine before that. I'm pessimistic about hot-plugged CPUs surviving without regressions in multiple levels of this stack.

klihub: @askervin Sure, I think it is reasonable.

What I meant is that the comment says that you had to use a different entry for cri-o vs. containerd, and the one for containerd looks like a libcgroupfs-driver entry, which suggests that at some point we were unintentionally running containerd with the cgroupfs driver, instead of the systemd one. This was expected before the above mentioned containerd commit, if during provisioning we somewhere generate a config file using `containerd config default` (which a quick `git grep` confirms), which used to incorrectly omit emitting the systemd cgroup driver setting (before that commit).

My indirect suggestion was that we should remove that (IMO misleading) comment and make sure that containerd is not run in our test setup with the cgroupfs driver any more.

klihub: @askervin So at least the comment ` # containerd:  kubepods/cpuset.cpus` should be removed. It was never an intentional setup, and also it has not been true since 3116747f534fc182ea40ac67146355e235cac340.

askervin: Thanks @klihub! Removed tmisleading comment lines.

PR #595, opened 2025-11-21
e2e: fix helm-launch when error is expected

askervin:
- helm install --rollback-on-failure causes long delays when container crashes. Drop the parameter when expecting failures. (Negative tests do not expect rollbacks.)

- Add expect_error=1 on negative test helm-launches where missing.

- More deterministic return 1 from helm-launch when expecting errors.

Fixes failing topology-aware/test12 and random failures in test05 and test13.


klihub (approved):
LGTM

PR #596, opened 2025-11-25
e2e: speed up test execution

askervin:
- Do not use proxy plugin if not needed. Skips repeating long "Configuring ...." steps.
- Use synchronize instead of copy for helm charts. Runs single rsync instead of multiple scp transfers.
- Use ssh connection multiplexing in case of using .ssh-config, too.


askervin:
If a vagrant vm is already running after previous tests, these changes speed-up the test start up time: from launching "run_tests.sh" until finishing "TASK [copy helm charts]":
- without patches: 37 seconds
- with patches: 4.8 seconds

PR #598, opened 2025-11-25
topology-aware: allow slicing 5 CPUs when 5 CPUs available

askervin:


cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -432,7 +432,7 @@ func (cs *supply) AllocateCPU(r Request) (Grant, error) {
 				cs.node.Name(), full, cs.isolated, err)
 		}
 
-	case full > 0 && cs.AllocatableSharedCPU() > 1000*full:
+	case full > 0 && cs.AllocatableSharedCPU() >= 1000*full:
```
klihub, line 435: I think this was deliberately disallowed, to prevent the shared pool from being used up fully. If we allowed it, then there would be a corresponding attempt to set the cpusets of any burstable or guaranteed containers in the same pool to the empty set, which will fail.

PR #600, opened 2025-11-28
metrics: switch to OpenTelemetry based metrics collection.

klihub:
This PR
- removes the original prometheus based metrics collection,
- removes any and related instrumentation bits,
- reworks these using OpenTelemetry,
- removes the original prometheus metrics collectors, and
- reimplements them using OpenTelemetry

### DIFF
--- a/config/crd/bases/config.nri_balloonspolicies.yaml
+++ b/config/crd/bases/config.nri_balloonspolicies.yaml
@@ -784,9 +784,12 @@ spec:
                       Endpoints are specified as full URLs, or as plain URL schemes which then
                       imply scheme-specific defaults. The supported schemes and their default
                       URLs are:
-                        - otlp-http, http: localhost:4318
-                        - otlp-grpc, grpc: localhost:4317
-                    example: otlp-http://localhost:4318
+                        - otlp-http: use OpenTelemetry HTTP[S] tracing exporter/collector
+                        - otlp-grpc: use OpenTelemetry gRPC tracing exported/collector
+                    enum:
+                    - otlp-http
+                    - otlp-grpc
+                    example: otlp-http
                     type: string
                 type: object
               loadClasses:
--- a/config/crd/bases/config.nri_templatepolicies.yaml
+++ b/config/crd/bases/config.nri_templatepolicies.yaml
@@ -508,9 +508,12 @@ spec:
                       Endpoints are specified as full URLs, or as plain URL schemes which then
                       imply scheme-specific defaults. The supported schemes and their default
                       URLs are:
-                        - otlp-http, http: localhost:4318
-                        - otlp-grpc, grpc: localhost:4317
-                    example: otlp-http://localhost:4318
+                        - otlp-http: use OpenTelemetry HTTP[S] tracing exporter/collector
+                        - otlp-grpc: use OpenTelemetry gRPC tracing exported/collector
+                    enum:
+                    - otlp-http
+                    - otlp-grpc
+                    example: otlp-http
                     type: string
                 type: object
               log:
--- a/config/crd/bases/config.nri_topologyawarepolicies.yaml
+++ b/config/crd/bases/config.nri_topologyawarepolicies.yaml
@@ -535,9 +535,12 @@ spec:
                       Endpoints are specified as full URLs, or as plain URL schemes which then
                       imply scheme-specific defaults. The supported schemes and their default
                       URLs are:
-                        - otlp-http, http: localhost:4318
-                        - otlp-grpc, grpc: localhost:4317
-                    example: otlp-http://localhost:4318
+                        - otlp-http: use OpenTelemetry HTTP[S] tracing exporter/collector
+                        - otlp-grpc: use OpenTelemetry gRPC tracing exported/collector
+                    enum:
+                    - otlp-http
+                    - otlp-grpc
+                    example: otlp-http
                     type: string
                 type: object
               log:
--- a/deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
+++ b/deployment/helm/balloons/crds/config.nri_balloonspolicies.yaml
@@ -784,9 +784,12 @@ spec:
                       Endpoints are specified as full URLs, or as plain URL schemes which then
                       imply scheme-specific defaults. The supported schemes and their default
                       URLs are:
-                        - otlp-http, http: localhost:4318
-                        - otlp-grpc, grpc: localhost:4317
-                    example: otlp-http://localhost:4318
+                        - otlp-http: use OpenTelemetry HTTP[S] tracing exporter/collector
+                        - otlp-grpc: use OpenTelemetry gRPC tracing exported/collector
+                    enum:
+                    - otlp-http
+                    - otlp-grpc
+                    example: otlp-http
                     type: string
                 type: object
               loadClasses:
--- a/deployment/helm/template/crds/config.nri_templatepolicies.yaml
+++ b/deployment/helm/template/crds/config.nri_templatepolicies.yaml
@@ -508,9 +508,12 @@ spec:
                       Endpoints are specified as full URLs, or as plain URL schemes which then
                       imply scheme-specific defaults. The supported schemes and their default
                       URLs are:
-                        - otlp-http, http: localhost:4318
-                        - otlp-grpc, grpc: localhost:4317
-                    example: otlp-http://localhost:4318
+                        - otlp-http: use OpenTelemetry HTTP[S] tracing exporter/collector
+                        - otlp-grpc: use OpenTelemetry gRPC tracing exported/collector
+                    enum:
+                    - otlp-http
+                    - otlp-grpc
+                    example: otlp-http
                     type: string
                 type: object
               log:
--- a/deployment/helm/topology-aware/crds/config.nri_topologyawarepolicies.yaml
+++ b/deployment/helm/topology-aware/crds/config.nri_topologyawarepolicies.yaml
@@ -535,9 +535,12 @@ spec:
                       Endpoints are specified as full URLs, or as plain URL schemes which then
                       imply scheme-specific defaults. The supported schemes and their default
                       URLs are:
-                        - otlp-http, http: localhost:4318
-                        - otlp-grpc, grpc: localhost:4317
-                    example: otlp-http://localhost:4318
+                        - otlp-http: use OpenTelemetry HTTP[S] tracing exporter/collector
+                        - otlp-grpc: use OpenTelemetry gRPC tracing exported/collector
+                    enum:
+                    - otlp-http
+                    - otlp-grpc
+                    example: otlp-http
                     type: string
                 type: object
               log:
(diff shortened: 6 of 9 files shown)


pkg/instrumentation/instrumentation.go
```diff
@@ -104,19 +104,32 @@ func start() error {
...
+		resource,
 		tracing.WithCollectorEndpoint(cfg.TracingCollector),
 		tracing.WithSamplingRatio(float64(cfg.SamplingRatePerMillion)/float64(1000000)),
 	); err != nil {
 		return fmt.Errorf("failed to start tracing: %v", err)
 	}
 
+	if cfg.PrometheusExport {
+		if cfg.MetricsExporter != "" && cfg.MetricsExporter != "prometheus" {
+			return fmt.Errorf("conflicting metric exporters: %s and %q",
```
askervin: s/metric/metrics

In addition to that, I'd be tempted to help user by providing names of conflicting configuration options in the error message. Especially "(implicit) prometheus" might sound difficult to find and fix in the configuration, yet it is not. Do you think we could spell out in the error message: "'prometheusExporter: true' and 'metricsExporter: %q'?

pkg/instrumentation/metrics/metrics.go
```diff
@@ -15,142 +15,210 @@
...
 	log          = logger.Get("metrics")
 )
 
-func WithExporterDisabled(v bool) Option {
+// WithExporter sets the type of metrics exporter to use.
+func WithExporter(v string) Option {
 	return func() error {
-		disabled = v
+		if v != "" && exporter != "" && v != exporter {
+			return fmt.Errorf("conflicting metrics exporter: %q and %q requested",
```
askervin, line 63: s/exporter/exporters.

pkg/instrumentation/metrics/metrics.go
```diff
@@ -15,142 +15,210 @@
...
-}
+		handlerOpts := promhttp.HandlerOpts{
+			ErrorHandling: promhttp.ContinueOnError,
+		}
+		m.Handle("/metrics", promhttp.HandlerFor(gatherer, handlerOpts))
 
-func Stop() {
-	if mux == nil {
-		return
+	case httpExporter, "otel-http":
```
askervin: Maybe `httpExporterOtel = "otel-http"` could be in consts, too? Or even `httpExporterOtelAlias` if "http" and "otel-http" are always interchangeable.

pkg/instrumentation/metrics/metrics.go
```diff
@@ -15,142 +15,210 @@
...
-	g *metrics.Gatherer
+	return nil
 }
 
-func newMetricsBlock(g *metrics.Gatherer) *MetricsBlock {
-	if g == nil {
-		return nil
+// Stop metrics collection and exporting.
+func Stop() {
+	if provider != nil {
```
askervin, line 212: Without really knowing internals here, would stopping be safer if the provider would be first unregistered from the mux, and shutdown only after that? (Mentally trying to avoid situation where the provider would get a request from the mux while or after being shut down.)

pkg/metrics/metrics.go
```diff
@@ -468,7 +466,8 @@ func (r *Registry) Configure(enabled []string, polled []string) (State, error) {
 	}
 
 	if len(unmatched) > 0 {
-		return r.state, fmt.Errorf("no collectors match globs %s", strings.Join(unmatched, ", "))
+		//return r.state, fmt.Errorf("no collectors match globs %s", strings.Join(unmatched, ", "))
```
askervin: // ...to be removed before merge?

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1588,6 +1610,9 @@ func (p *balloons) setConfig(bpoptions *BalloonsOptions) error {
 			log.Warnf("failed to apply CPU class to balloon %s: %v", bln.PrettyName(), err)
 		}
 	}
+
+	p.NewMeters()
```
marquiz, line 1614: Why do we need to do this on every reconfigure? Wouldn't it be enough to only do it once in Setup()? Would simplify a lot, I think

cmd/plugins/balloons/policy/metrics.go
```diff
@@ -15,53 +15,24 @@
...
-
-// Metrics defines the balloons-specific metrics from policy level.
-type Metrics struct {
-	Balloons []*BalloonMetrics
+// Metrics defines the balloons-specific metric instruments.
+type Meters struct {
+	p       *balloons
+	meter   metric.Meter
+	balloon metric.Int64ObservableGauge
+	cbReg   metric.Registration
```
marquiz: Unused? We can just drop if we're not going to unregister the callback

pkg/metrics/otel.go
```diff
@@ -0,0 +1,245 @@
...
+func (m *meter) Float64UpDownCounter(name string, options ...otelmetric.Float64UpDownCounterOption) (otelmetric.Float64UpDownCounter, error) {
+	return m.Meter.Float64UpDownCounter(m.meterName(name), options...)
+}
+
+// Float64ObservableUpDownCounter returns the corresponding instrument for the meter.
+func (m *meter) Float64ObservableUpDownCounter(name string, options ...otelmetric.Float64ObservableUpDownCounterOption) (otelmetric.Float64ObservableUpDownCounter, error) {
+	return m.Meter.Float64ObservableUpDownCounter(m.meterName(name), options...)
+}
+
+// IsEnabled returns true if the given meter is enabled in the current configuration.
```
askervin: "if the given meter is enabled" looks a bit confusing, because there doesn't seem to be a meter given.

config/crd/bases/config.nri_balloonspolicies.yaml
```diff
@@ -749,8 +749,25 @@ spec:
...
+                  metricsExporter:
+                    description: |-
+                      MetricsExporter defines which exporter is used to export metrics.
+                      The supported exporters are:
+                        - prometheus: use OpenTelemetry prometheus exporter
+                        - otel-http, http: use OpenTelemetry HTTP metrics exporter
+                        - otel-grpc, grpc: use OpenTelemetry gRPC metrics exporter
+                    enum:
+                    - prometheus
+                    - otel-http
```
marquiz: nit: probably a more correct name would be otlp-http (and otlp-grpc)?

pkg/instrumentation/metrics/metrics.go
```diff
@@ -15,142 +15,210 @@
...
+		log.Info("no metrics exporter configured, metrics collection disabled")
+		metrics.SetProvider(nil)
+		metrics.Configure(nil)
 		return nil
 	}
 
-	if disabled {
-		log.Info("metrics exporter disabled")
+	if m == nil {
+		log.Info("no mux provided, metrics collection disabled")
```
marquiz, line 125: We don't need the mux for any other than prometheus(?) So it can be nil in other cases?

pkg/metrics/otel.go
```diff
@@ -0,0 +1,245 @@
...
+
+// WithMeterOptions sets OpenTelemetry options for a meter.
+func WithMeterOptions(options ...otelmetric.MeterOption) *meterOption {
+	return &meterOption{
+		otel: options,
+	}
+}
+
+// Meter returns a meter for the given subsystem with the provided options.
+func (mp *meterProvider) Meter(subsys string, options ...*meterOption) otelmetric.Meter {
```
marquiz, line 102: OK, now I got this "re-create-meter-on-each-reconfigure". Somehow this feels clunky, excessive and "wrong".

I'd like to explore other alternatives, e.g. unregister callbacks of disabled subsystems or use dynamic "views" (WithView) or wrap the reader instead of meter

pkg/resmgr/policy/metrics.go
```diff
@@ -228,11 +196,36 @@ func (s *SystemCollector) Update() {
...
+			),
+		)
 	}
 	for _, c := range s.Cpus {
-		s.Metrics[cpuAllocation].WithLabelValues(c.IdLabel).Set(float64(c.Allocation))
-		s.Metrics[cpuContainers].WithLabelValues(c.IdLabel).Set(float64(c.ContainerCount))
+		s.CpuAllocation.Record(
+			context.Background(),
+			int64(c.Allocation),
+			metric.WithAttributes(
```
askervin: With a huge number of CPUs, it might be cheaper to use static AttributeSets (metric.WithAttributeSet) to avoid creating so many copies... But I think at this point this might be too early optimization.

pkg/instrumentation/instrumentation.go
klihub: Updated error message according to that.

pkg/instrumentation/metrics/metrics.go
klihub: Okay, I simplified this to only accept one of "prometheus", "otlp-http" and "otlp-grpc" (as pointer out by @marquiz, more in line with stock naming and also with our tracing configuration). I added an extra commit to do a similar reduction for tracing which similarly only accepts "otlp-http" and "otlp-grpc" (with everyhing else configurable via extra environment variables).

pkg/instrumentation/metrics/metrics.go
klihub: Fixed.

pkg/instrumentation/metrics/metrics.go
klihub: Changed ordering of Unregister and shutdown.

pkg/metrics/otel.go
klihub: Updated comment aiming for a less confusing description.

cmd/plugins/balloons/policy/balloons-policy.go
klihub: Unfortunately we can't, because the meters are metric-provider specific. So if we flip on metrics from disabled to enabled, nothing would be exported if we didn't recreate the meters here with the potentially new provider.

cmd/plugins/balloons/policy/metrics.go
klihub: Removed.

config/crd/bases/config.nri_balloonspolicies.yaml
klihub: Thanks for pointing that out. Updated to (only) use/support "otlp-http" and "otlp-grpc" (beside "prometheus") which is also consistent now with how we configure tracing exporters.

pkg/instrumentation/metrics/metrics.go
klihub: We have it separately configured, so it could be nil. We use it for other things, for instance as a 'health' endpoint.

pkg/metrics/otel.go
klihub: There is no way around the fact that a meter created by a nop provider can never be turned into one created by another provider. There is no other way than to re-create it.

pkg/resmgr/policy/metrics.go
klihub: Updated to reuse a pre-created `attribute.KeyValue` for oft-used labels (`node.id`, `cpu.id`), readily wrapped in an `attribute.Set`.

kad (approved):
LGTM

PR #601, opened 2025-12-02
topology-aware: avoid slicing busy shared pools empty

klihub:
This PR updates the topology-aware policy CPU accounting and allocation algorithm, to
- account for best-effort and zero-cp-request burstable containers assigned to shared pools
- prevent the policy from slicing a shared CPU pool empty if it is busy
 
A shared pool is considered busy if it has at least one best-effort or zero-cpu-req. burstable container assigned to it or any of its children.

### DIFF
--- a/cmd/plugins/topology-aware/policy/resources.go
+++ b/cmd/plugins/topology-aware/policy/resources.go
@@ -897,7 +897,7 @@ func (cs *supply) GetScore(req Request) Score {
 			score.isolated = cs.isolated.Size() - full
 		}
 
-		// if we don't want isolated or there is not enough, calculate slicable capacity
+		// if we don't want isolated or there is not enough, calculate sliceable capacity
 		if !cr.isolate || score.isolated < 0 {
 			score.shared -= 1000 * full
 		}
--- a/cmd/plugins/topology-aware/policy/pools.go
+++ b/cmd/plugins/topology-aware/policy/pools.go
@@ -722,7 +722,7 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
 	//       * more isolated capacity wins
 	//       * for a tie, prefer the smaller id
 	//   - for (non-reserved) exclusive allocations
-	//       * more slicable (shared) capacity wins
+	//       * more sliceable (shared) capacity wins
 	//       * for a tie, prefer the smaller id
 	//   - for (non-reserved) shared-only allocations
 	//       * fewer colocated containers win
@@ -1043,18 +1043,18 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
 			}
 		}
 
-		// more slicable shared capacity wins
+		// more sliceable shared capacity wins
 		if request.FullCPUs() > 0 && (shared1 > 0 || shared2 > 0) {
 			if shared1 > shared2 {
-				log.Debug("  => %s WINS on more slicable capacity", node1.Name())
+				log.Debug("  => %s WINS on more sliceable capacity", node1.Name())
 				return true
 			}
 			if shared2 > shared1 {
-				log.Debug("  => %s WINS on more slicable capacity", node2.Name())
+				log.Debug("  => %s WINS on more sliceable capacity", node2.Name())
 				return false
 			}
 
-			log.Debug("  => %s WINS based on equal slicable capacity, lower id",
+			log.Debug("  => %s WINS based on equal sliceable capacity, lower id",
 				map[bool]string{true: node1.Name(), false: node2.Name()}[id1 < id2])
 
 			return id1 < id2


cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -78,6 +79,8 @@ type Supply interface {
 	GetScore(Request) Score
 	// AllocatableSharedCPU calculates the allocatable amount of shared CPU of this supply.
 	AllocatableSharedCPU(...bool) int
+	// SlicableCPUs calculates the shared cpuset we can slice exclusive CPUs off of.
+	SlicableCPUs() (cpuset.CPUSet, error)
```
askervin: s/Slicable/Sliceable/g
s/slicable/sliceable/g

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -997,12 +1013,81 @@ func (cs *supply) AllocatableSharedCPU(quiet ...bool) int {
...
+	//
+	// TODO(klihub): We might need to try speeding this up if it gets too slow
+	// on nodes with a lot of CPUs and topology zones as the number of containers
+	// grows. hasZeroCpuReqContainer uses foreachContainer, which in turn loops
+	// through each resource grant calling a callback if the grant is assigned to
+	// a particular node. foreachContainer can stop early based on the callback's
+	// return value, but {Breadth,Depth}First can not ATM. Obvious optimizations
+	// would be to 1) allow {Breadth,Depth}First to stop early if possible, and
+	// 2) store grants per assigned node.
+	hasBestEffort := false
```
askervin: Perhaps hasBestEffort could be renamed as hasContainersWithoutCpuReqs or something similar. Just in case someone would read this piece of code and come up with the bad idea that maybe hasZeroCpuReqContainer() could be rewritten based on QoS classes.

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -433,19 +436,28 @@ func (cs *supply) AllocateCPU(r Request) (Grant, error) {
 		}
 
 	case full > 0 && cs.AllocatableSharedCPU() > 1000*full:
```
askervin, line 438: I'm wondering if this change would actually allow having

`cs.AllocatableSharedCPU() >= 1000*full` instead of `cs.AllocatableSharedCPU() > 1000*full`

because now we can be sure that we can take all full CPUs without drying out a shared pool that has containers in it.

If we have NUMA nodes of size X CPUs, I'd guess that a user expects that a guaranteed container requesting exactly X CPUs would get them all from the same NUMA node, given that there are no other containers lying around.

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -997,12 +1013,81 @@ func (cs *supply) AllocatableSharedCPU(quiet ...bool) int {
...
+		// priority preference of any ongoing allocation, trying to slice
+		// CPUs with a matching preference. We don't do that ATM.
+
+		cset, err := cs.takeCPUs(&cpus, nil, free, nonePrio)
+		if err != nil {
+			errs = append(errs, err)
+			return
+		}
+
+		slicable = slicable.Union(cset)
```
askervin: Not fully sure, so this is more of a question rather than a suggestion:

Assume a parent node P has two child nodes, C1 and C2. C1 has shared CPUs cpu10 and cpu11, and C2 shared CPUs cpu20 and cpu21. Both C1 and C2 host one BestEffort container each. Parent node has no containers.

Does this mean that parent node P has shared CPUs cpu10, cpu11, cpu20, and cpu21? And that depth-first union could first go to C1, make slicable {cpu10}, then to C2 and make slicable {cpu10}.Union({cpu20}), and finally handle parent P, and make slicable {cpu10,cpu21}.Union({cpu10,cpu11,cpu20}), or even {cpu10,cpu21}.Union({cpu10,cpu11,cpu20,cpu21}) because P does not have BestEffort containers?

Perhaps it would be safer to create an upper bound for slicable CPUs from child nodes only, and never add CPUs to that set from parent nodes. Removal might be needed, if C1 and C2 had no containers but there would be a BestEffort container in P...

cmd/plugins/topology-aware/policy/resources.go
klihub: Changed variable name to `hasZeroCpuReqs`.

cmd/plugins/topology-aware/policy/resources.go
klihub: Fixed. And in a bunch of existing printouts and comments and log messages, too.

cmd/plugins/topology-aware/policy/resources.go
klihub: > Perhaps it would be safer to create an upper bound for slicable CPUs from child nodes only, and never add CPUs to that set from parent nodes.

That's what we do (well, we don't artificially limit the number of sliceable containers): we don't add sliceable CPUs from the node itself, unless it is a leaf node. If we did not do it this way, then this fix wouldn't work for cases where we need to prevent slicing empty a busy shared pool in the subtree below the node where we are about to make an exclusive allocation.

cmd/plugins/topology-aware/policy/resources.go
klihub: I was considering that and I think fix should allow us to do that. I decided against doing such a change in this particular PR, to keep it a pure bug fix one. We can file a separate PR for that if we decide to change the current behavior.

klihub: @askervin Filed draft #602 for (playing around with) that (idea).

PR #603, opened 2025-12-10
e2e: early draft of cxl and custom kernel support, possibly to be split in multiple PRs.

askervin:
Qemu parameters:
- topology.var.json supports extra "cxl" definition. It defines a hierarchical structure of:
  host bridge(s) / root port(s) / switch(es) / cxl memories
  where switches may be completely left out and cxl memories connected directly to root ports, too.

ansible playbooks:
- Build kernel packages from any predefined tree or arbitrary "kernel_getsource" command. Example:
  `kernel_getsource="git clone --depth 1 -b v6.18 https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git linux"`
- Cache kernel packages on host. Cache hit takes matching kernel_getsource and kernel config (sha256).
- Upload custom kernel packages needed by a test to the vm but do not install it right away. (Similar logic to container images.)

vm.bash script API:
- Add vm-kernel-pkgs-{install,uninstall} functions for installing and uninstalling custom kernel packages. Uninstall restores original kernel. Install does nothing if kernel is already installed.

### DIFF
--- a/test/e2e/lib/vm.bash
+++ b/test/e2e/lib/vm.bash
@@ -409,17 +409,25 @@ vm-cpu-hotremove() { # script API
     vm-monitor "device_del ${deviceid}"
 }
 
+_vm_cxl_hotplug_count=""
+
 vm-cxl-hw() {
     # Usage: vm-cxl-hw
     #
     # List hotpluggable and removable cxl memory devices
     # See also: vm-cxl-hotplug, vm-cxl-hotremove
+    local plugged plugged_id
+    declare -A plugged
+    for plugged_id in $(vm-monitor "info qtree -b" | awk -F\" '/dev: cxl-type3/{print $2}' | sed 's/\.hp.*//g'); do
+        plugged[$plugged_id]=1
+    done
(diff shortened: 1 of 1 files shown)


test/e2e/lib/vm.bash
```diff
@@ -339,17 +339,35 @@ vm-reboot() { # script API
...
+
+        while (( $(date +%s) < $_deadline )); do
             vagrant up --no-provision
             if [ $? = 0 ]; then
                 break
             fi
+            vagrant halt
+            sleep 5
+            vagrant halt --force
+            sleep 3
```
klihub, line 370: Yes, you often need to kill the right qemu instance in the end to avoid a lingering one. A `(cd $OUTPUT_DIR; kill $(ps axu | grep qemu | grep $(realpath .) | tr -s '\t' ' ' | cut -d ' ' -f2))` should be a big enough hammer to do the trick.

askervin:
Closing this PR as it is now split into hopefully reviewable PRs.

PR #602, opened 2025-12-03
topology-aware: allow slicing idle shared pools empty.

klihub:
Note: this PR is stacked on #601, which should be reviewed first.

This PR updates the topology-aware policy accounting and allocation algorithm to allow slicing an idle shared CPU pool empty for exclusive CPU allocations.

PR #604, opened 2025-12-16
topology-aware: don't (ac)count 0 CPU req. containers in the reserved pool to shared pools 

klihub:
This patch series should fix the observed failure to allocate exclusive CPUs reported here: https://github.com/containers/nri-plugins/pull/602#issuecomment-3617041869.

In particular it updates the topology-aware policy to
- correctly disregard 0 CPU request containers in the reserved pool from normal shared pools
- print grants with reserved CPU pool assignments correctly

Note that due to runc still being broken wrt. allocating CPUs with logical ID > 1024, the above linked test case of #599 still fails. But now only due to runc, with the actual resource allocation and intended policy CPU assignment being as expected, as can be observed by running something like this on the test VM host once the test case has failed:

```bash
$ crictl inspect $(crictl ps | grep pod0c0 | tr -s '\t' ' ' | cut -d ' ' -f1) | jq .info.runtimeSpec.linux.resources.cpu.cpus
"0-2,511,4095"
```

### DIFF
--- a/cmd/plugins/topology-aware/policy/resources.go
+++ b/cmd/plugins/topology-aware/policy/resources.go
@@ -1280,13 +1280,14 @@ func (cg *grant) String() string {
 	if !cg.exclusive.IsEmpty() {
 		exclusive = fmt.Sprintf(", exclusive: %s", cg.exclusive)
 	}
-	if cg.ReservedPortion() > 0 {
+	if cg.ReservedPortion() > 0 || (cg.cpuType == cpuReserved && cg.SharedPortion() == 0) {
 		reserved = fmt.Sprintf(", reserved: %s (%dm)",
 			cg.node.FreeSupply().ReservedCPUs(), cg.ReservedPortion())
-	}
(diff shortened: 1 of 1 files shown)


cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -625,12 +625,12 @@ func (p *policy) updateSharedAllocations(grant *Grant) {
 	}
 }
 
-func (p *policy) foreachContainer(pool Node, fn func(ctr cache.Container) bool) {
+func (p *policy) foreachContainer(pool Node, fn func(g Grant) bool) {
```
askervin: s/foreachContainer/foreachGrant

klihub: Yeah... Actually I had it like that in the original version `foreachContainer`. I'll update it like that.

klihub: Updated.

PR #610, opened 2025-12-18
.github: fix unstable chart publishing with Helm v4.x.

klihub:
Unstable Helm chart publishing has been broken since Helm 4.x came out. This is due to our questionable practice to install and use the latest Helm release in CI, combined with Helm 4.x refusing to parse a registry URL with a path suffix in login. Fix the latter of these in our workflow.


askervin (approved):
LGTM

PR #611, opened 2025-12-18
 e2e: support "cxl" in hardware topology (CXL-1)

askervin:
- topology2qemuopts converts CXL structures to Qemu PCIe devices
      and memory backend objects.
    
    - Add functions to e2e script API for listing, hotplugging
      and hotremoving CXL devices.
    
    - Move hardcoded Qemu accelerator from Vagrantfile template to
      topology2qemuopts that already generates Qemu machine parameters.
      There is no functional change: keep using kvm. However, this enables
      switching back and forth between "accel=kvm" with "accel=tcg" in
      generated Vagrantfile more cleanly than between "-enable-kvm" and
      "". While "accel=tcg" is needed to actually use CXL memory in vm, it
      is most likely faster to do test setup first with kvm, and finally
      reboot to tcg once the vm has been fully prepared for such a test.
    
    - Use explicitly specified PCIe root port for Qemu disk in
      Vagrantfile to avoid errors when CXL PCIe devices are added.
    
    - Fix a bug in SEPARATED_OUTPUT_VARS=1 output, where an empty
      deviceparams or objectparams list caused extra ", ," in output
      leading to errors in Vagrantfile.

Stacked on top of #587.

### DIFF
--- a/test/e2e/files/Vagrantfile.in
+++ b/test/e2e/files/Vagrantfile.in
@@ -89,12 +89,16 @@ Vagrant.configure("2") do |config|
     qemu.arch = "x86_64"
     qemu.machine = "QEMU_MACHINE"
     qemu.cpu = "QEMU_CPU"
+    qemu.drive_interface = "none,id=disk0" # qemu.extra_drive_args may not work, piggyback id to interface
     qemu.net_device = "virtio-net-pci"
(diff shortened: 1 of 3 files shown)


klihub:
@askervin This needs a rebase.

PR #612, opened 2025-12-18
e2e: enable custom kernel building, caching and installing (CXL-2)

askervin:
- kernel_getsource environment variable specifies either a
      shorthand ("vanilla", "mainline" or "cxl-next" to start with)
      or direct shell command(s) that result in "linux" directory with
      kernel sources in it.
    
    - kernel_config specifies kernel .config for a custom kernel filename.
      The file is located on host in playbook lookup('file', ...) path.
    
    - Ansible playbooks make sure that custom kernel packages, should they
      be needed, are located in vm when test script execution starts. The
      test script can install and uninstall the packages at will.  If
      kernel packages for kernel_getsource and kernel_config are already
      cached on host, playbooks copy them on vm. If not, playbooks fetch
      sources on vm, applies .config, installs kernel build dependencies,
      buildds kernel packages on vm and caches them on host.

### DIFF
--- a/test/e2e/playbook/custom-kernel-fedora.yaml
+++ b/test/e2e/playbook/custom-kernel-fedora.yaml
@@ -1,0 +1,28 @@
+---
+- name: Install kernel build dependencies
+  become: yes
+  ansible.builtin.package:
+    pkg:
+      - fedpkg
+      - fedora-packager
+      - rpmdevtools
+      - ncurses-devel
+      - pesign
+      - grubby
+      - make
+      - gcc
+      - flex
+      - bison
+      - elfutils-devel
+      - elfutils-libelf-devel
+      - dwarves
+      - openssl
+      - openssl-devel
+      - perl
+    state: present
+
+- name: Build kernel RPMs
+  shell: "cd linux && make -j$(nproc) binrpm-pkg"
+
+- name: Create vm:~/kernel-pkgs.tar containing kernel RPMs
+  shell: "cd linux/rpmbuild/RPMS/x86_64; tar cf ~/kernel-pkgs.tar kernel-*.rpm"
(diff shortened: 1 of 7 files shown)


test/e2e/playbook/custom-kernel-fedora.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - flex
+      - bison
+      - elfutils-devel
+      - elfutils-libelf-devel
+      - dwarves
+      - openssl
+      - openssl-devel
+      - perl
+    state: present
+    update_cache: true
```
klihub: nano-nit: This is maybe not even a nit, I just try to understand why you chose to update the cache here. Is this strictly necessary ? Don't we do this when we spin up the VM initially ? So unless the VM is really long-lived and we'd always want to build the kernel with the absolute latest and greatest available distro bits, the cache should be fresh (enough) ?

test/e2e/playbook/custom-kernel.yaml
```diff
@@ -0,0 +1,92 @@
...
+          stat:
+            path: "{{ cached_kernel_pkgs_tarball }}"
+          register: cached_kernel_stat
+
+        - name: If we have a cached kernel, push it to the vm
+          when: cached_kernel_stat.stat.exists
+          synchronize:
+            src: "{{ cached_kernel_pkgs_tarball }}"
+            dest: "/home/vagrant/kernel-pkgs.tar"
+            use_ssh_args: true
```
klihub, line 47: Again, nice. I think we don't do this in some other places where we could.

test/e2e/playbook/custom-kernel-fedora.yaml
askervin: Thanks! Fixed here and in the corresponding spot in `custom-kernel-ubuntu.yaml`, too.

test/e2e/playbook/custom-kernel.yaml
askervin: No action taken.

PR #614, opened 2025-12-18
e2e: add a test that hotplugs and hotremoves CXL memory (CXL-3)

askervin:


test/e2e/memory.test-suite/memory-policy/n4-cxl/test00-build-cxl-kernel/code.var.sh
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+vm-kernel-pkgs-install
+
+vm-command "command -v cxl || dnf install -y /usr/bin/cxl numactl golang"
+
+vm-command "mkdir udev-monitor 2>/dev/null" && {
+    udev_monitor_tool="${TEST_DIR%%/test/e2e/*}/scripts/udev-monitor/udev-monitor.go"
+    vm-put-file "$udev_monitor_tool" "./udev-monitor/udev-monitor.go"
+    vm-command "cd udev-monitor && go mod init udev-monitor && go mod tidy && go build . && cp -v udev-monitor /usr/local/bin"
```
klihub: nit: wouldn't it be better to `[GOARCH=amd64] go build ${TEST_DIR%%/test/e2e/*}/scripts/udev-monitor` locally then `vm-put-file udev-monitor /usr/local/bin/udev-monitor`. Then we wouldn't need to install the full golang toolchain just so we can get udev-monitor compiled on the VM...

klihub (approved):
I had a small optional nit, but already LGTM regardless.

test/e2e/memory.test-suite/memory-policy/n4-cxl/test00-build-cxl-kernel/code.var.sh
askervin: Good point. Switched to building on host and copying the binary as you suggested.

Also: noticed that starting vm-command-q udev-monitor in the background in the test prevents run.sh to exit after printing PASS. Therefore added vm-command "pkill udev-monitor" in the end of the script, which fixed the issue. (And of course it anyway better as it cleans up its own mess.)

PR #617, opened 2026-01-12
balloons: extend composite balloons with a kind of load balancing

askervin:
Composite balloons are balloons that get their CPUs by a union of the CPUs of their component balloons. Composite balloons are created from balloon types that have one or more `components` (balloon types).

Until now, when a new composite balloon is created, new balloon of each component type has been created, too. This enabled creating balloons that constist of an equal number of CPUs from every NUMA node. However if a system has two or more sockets, one may want to allocate an equal number of CPUs from every NUMA node from socket 0 for one container, and from socket 1 to another container, but without enforcing either of the sockets in the balloons annotation. Or more general, one may want to balance of similar containers across multiple balloon types. The purpose of this PR is to enable this kind of balancing.

When a new balloon of a composite balloon type is created, `componentCreation` specifies a strategy for choosing which components of the composite balloon that will be created and thus used for allocating CPUs for it. The default behavior is not changed: if componentCreation is unspecified or set to "all", one balloon from every composite type gets instantiated, and an equal number of CPUs is allocated through each component.

One new component creation strategy is introduced: "first-min-balloons". It will use only one of the components for CPU allocation: the first component whose balloon type that has the minimum number of balloon instances alive.

### DIFF
--- a/docs/resource-policy/policy/balloons.md
+++ b/docs/resource-policy/policy/balloons.md
@@ -275,13 +275,23 @@ Balloons policy parameters:
     should get their CPUs from separate cache blocks for best
     performance. Every listed class must be specified in
     `loadClasses`.
-  - `components` is a list of components of a balloon. If a balloon
-    consists of components, its CPUs are allocated by allocating CPUs
-    for each component balloon separately, and then adding them up.
-    See [combining balloons](#combining-balloons) for more details and
-    an example. Properties of components in the list are:
+  - `components` list includes component balloon types. If non-empty,
+    the balloon is a composite balloon whose CPUs are not allocated
+    directly to itself but its CPUs are the union of CPUs of its
+    component balloons. See [combining balloons](#combining-balloons)
+    for more details and an example. Properties of components in the
+    list are:
     - `balloonType` specifies the name of the balloon type according
-      to which CPUs are allocated to this component.
+      to which CPUs are allocated to the component.
+  - `componentCreation` is a string that specifies the strategy for
+    creating `components` for a new composite balloon. If unspecified,
+    one instance of each component is created. Valid strategies are:
+    - `all`: create one instance of every component (the default).
+    - `balance-balloons`: create only the first component whose
+      balloon types have the minimal number of balloons alive at the
+      time of creation. This strategy is suitable for balancing
+      containers over multiple balloon types. See [combining
+      balloons](#combining-balloons) for an example.
 - `loadClasses`: lists properties of loads that containers in balloons
   generate to some parts of the system. When the policy allocates CPUs
   for load generating balloon instances, it selects CPUs so that it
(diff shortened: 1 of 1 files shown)


cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -882,14 +903,49 @@ func (p *balloons) newCompositeBalloon(blnDef *BalloonDef, confCpus bool, freeIn
...
+		sort.Slice(instComp, func(i, j int) bool {
+			return instComp[i][0] < instComp[j][0] ||
+				(instComp[i][0] == instComp[j][0] && instComp[i][1] < instComp[j][1])
+		})
+		log.Debugf("instance-components of composite balloon %s: %v",
+			blnDef.Name, instComp)
+		// create only one component balloon for the component
+		compIdx := instComp[0][1]
+		comp := blnDef.Components[compIdx]
+		compDef := p.balloonDefByName(comp.DefName)
```
kad, line 969: Do we have chance here with race with line 929? or just additional check if compDef is not nil? or save it instead of calling again?

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -882,14 +903,49 @@ func (p *balloons) newCompositeBalloon(blnDef *BalloonDef, confCpus bool, freeIn
...
+		}
+		// sort, first by number-of-instances ascending, then by component-index ascending
+		sort.Slice(instComp, func(i, j int) bool {
+			return instComp[i][0] < instComp[j][0] ||
+				(instComp[i][0] == instComp[j][0] && instComp[i][1] < instComp[j][1])
+		})
+		log.Debugf("instance-components of composite balloon %s: %v",
+			blnDef.Name, instComp)
+		// create only one component balloon for the component
+		compIdx := instComp[0][1]
```
kad: for readability, maybe some const/enum for `[0]` and `[1]`?

docs/resource-policy/policy/balloons.md
```diff
@@ -275,13 +275,23 @@ Balloons policy parameters:
...
+    for more details and an example. Properties of components in the
+    list are:
     - `balloonType` specifies the name of the balloon type according
-      to which CPUs are allocated to this component.
+      to which CPUs are allocated to the component.
+  - `componentCreation` is a string that specifies the strategy for
+    creating `components` for a new composite balloon. If unspecified,
+    one instance of each component is created. Valid strategies are:
+    - `all`: create one instance of every component (the default).
+    - `first-min-balloons`: create only the first component whose
```
kad: I wonder if there is a better name for this? e.g. something with balancing or "lazy" creation strategies?

askervin: Thanks for the review, @kad! I already wanted to change the name, at minimum remove the "first-" prefix. And now that you brought this up, perhaps we could use "balance-balloons" instead of "min-balloons", what do you think?

Keeping "balloons" in the name would still leave room for future strategies, if we wish to balance by number of free preferred CPUs or what not in the future.

kad: I'm fine with `balance-balloons`

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Thanks, added compDef==nil check, similarly to other places in this function.

There is no race this function.

Even though it is a normal situation that this function gets called recursively via newBalloon(), it could be optimized by storing even `numInstances` instead of computing `balloonsByDefIncludingComponents()` many times. Yet new balloons get created, they and their newly created components won't still become accessible by balloon instance counter until returning from all the recursions and finally adding the "root balloon" of the created structure to the `p.balloons` slice. However, because the number of balloon types and balloons is not likely to grow too large, I decided to prefer simplicity and not add a cache that would be accessible through mutual recursive calls.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Good point. Replaced the int pair with `struct { count, index }` for readability.

docs/resource-policy/policy/balloons.md
askervin: Renaming to `balance-balloons` done in user interface, struct attributes and naming of related `const`s.

kad (approved):
LGTM

PR #618, opened 2026-01-23
balloons: support configuring Linux scheduling parameters 

askervin:


klihub:
> @klihub, it might be wise to add either to `applyProcessProperties()` or to its caller a check if these properties are configurable. That would avoid a lot of unnecessary calls and their failures in cache's container object. If you agree, would you have a suggestion how to know if applying these properties makes sense?

Sorry, I'm not sure how to interpret the question. I mean, the policy call context (which NRI event we're processing, for what container, and which container's what property we consider updating) should determine whether a container's parameter can be updated. Is this what you are referring to ?

klihub:
> > @klihub, it might be wise to add either to `applyProcessProperties()` or to its caller a check if these properties are configurable. That would avoid a lot of unnecessary calls and their failures in cache's container object. If you agree, would you have a suggestion how to know if applying these properties makes sense?
> 
> Sorry, I'm not sure how to interpret the question. I mean, the policy call context (which NRI event we're processing, for what container, and which container's what property we consider updating) should determine whether a container's parameter can be updated. Is this what you are referring to ?

One context where I see how policies end up doing the wrong thin is the initial sync for deleted and discovered containers. Both balloons and topology-aware basically implement it by looping through all deleted containers doing a `ReleaseResources()` then looping through all discovered/added containers doing a `AllocateResources()`. The latter, `AllocateResources()` is normally triggered by an NRI `CreateContainer` which is called prior to actually creating the container, but in this context it is called for a running container.

Ideally the policy should check what is the current state of the container, and if it is already already running (or created) it should not bother with anything else than resource updates which are still possible. Additionally, we should update the policy-level `Sync()` and allow it to return `(evict []*Container, error)` instead of just an error. Then the policy would have two alternatives during `Sync()` wrt. each discovered container: limit itself to updating/adjusting resources, or ask for the eviction of the container if the policy decision involves strictly required adjustments to the container which can't be done with an update.

fmuyassarov (commented):
LGTM

PR #619, opened 2026-01-24
resource-annotator: add resource annotator mutating webhook.

klihub:
This PR
- lifts over the origin `cri-resmgr-webhook` annotating mutating webhook (courtesy of @marquiz)
- adjusts it a bit for nri-plugins
- adds a helm chart for resource-annotator
- updates the common resmgr resource estimation code to prefer annotated resources if present

TODO (some probably with subsequent PRs):
- [x] deploy configuration by helm chart
- [ ] e2e test case
- [ ] cert-manager-injected certificates

Testing this in practice:
```shell
$ # For now you have to import and tag resource-annotator image from tarball (no ghcr.io hosted image yet)
$ ctr -n k8s.io image import $TARBALL
$ ctr -n k8s.io images tag docker.io/library/resource-annotator:$VERSION ghcr.io/containers/nri-plugins/resource-annotator:unstable
$ # Create TLS certificate for webhook HTTPS access
$ mkdir -p ~/resource-annotator
$ NS=kube-system SVC=resource-annotator
$ CRT=~/resource-annotator/server-crt.pem KEY=~/resource-annotator/server-key.pem
$ openssl req -x509 -newkey rsa:2048 -sha256 -days 365 -nodes \
       -keyout $KEY \
       -out $CRT \
       -subj "/CN=$SVC.$NS.svc" \
       -addext "subjectAltName=DNS:$SVC,DNS:$SVC.$NS,DNS:$SVC.$NS.svc"
$ # Install webhook using helm
$ helm -n $NS install webhook ./deployment/helm/resource-annotator \
        --set image.tag=unstable \
        --set image.pullPolicy=Never \
        --set service.base64Crt=$(base64 -w0 < $CRT) \
        --set service.base64Key=$(base64 -w0 < $KEY)
```

Now if you create a pod, you should see its containers' CPU and memory resource requirements get annotated on the pod:

```shell
$ cat busybox.yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox
spec:
  containers:
  - name: c0-shared
    image: busybox:latest
    command:
      - sh
      - -c
      - sleep 3600
    resources:
      requests:
        cpu: 123m
        memory: 75M
      limits:
        cpu: 456m
        memory: 95M
    imagePullPolicy: IfNotPresent
  - name: c1-shared
    image: busybox:latest
    command:
      - sh
      - -c
      - sleep 3600
    resources:
      requests:
        cpu: 258m
        memory: 99M
      limits:
        cpu: 369m
        memory: 101M
    imagePullPolicy: IfNotPresent
  terminationGracePeriodSeconds: 1
$ kubectl apply -f busybox.yaml
$ kubectl get pods busybox -ojson | jq '.metadata.annotations["noderesource.dev/resources"]' | jq fromjson
{
  "containers": {
    "c0-shared": {
      "limits": {
        "cpu": "456m",
        "memory": "95M"
      },
      "requests": {
        "cpu": "123m",
        "memory": "75M"
      }
    },
    "c1-shared": {
      "limits": {
        "cpu": "369m",
        "memory": "101M"
      },
      "requests": {
        "cpu": "258m",
        "memory": "99M"
      }
    }
  }
}
```

### DIFF
--- a/pkg/resmgr/cache/cache.go
+++ b/pkg/resmgr/cache/cache.go
@@ -68,6 +68,10 @@ const (
 	// UnlimitedBurstabileKey defines the preferred topology level for containers
 	// with unlimited burstability.
 	UnlimitedBurstableKey = "unlimited-burstable." + kubernetes.ResmgrKeyNamespace
+
+	// AnnotatedResourcesKey can be used to annotate resource requirements of
+	// containers and init containers on the pod.
+	AnnotatedResourcesKey = kubernetes.AnnotatedResourcesKey
 )
 
 // PodState is the pod state in the runtime.
--- a/pkg/resmgr/cache/container.go
+++ b/pkg/resmgr/cache/container.go
@@ -306,8 +306,36 @@ func isReadOnlyDevice(rules []*nri.LinuxDeviceCgroup, d *nri.LinuxDevice) bool {
 	return readOnly
 }
 
+func (c *container) GetAnnotatedResources() (v1.ResourceRequirements, bool) {
+	pod, ok := c.GetPod()
+	if !ok {
+		return v1.ResourceRequirements{}, false
+	}
+
+	data, ok := pod.GetAnnotation(AnnotatedResourcesKey)
+	if !ok {
+		return v1.ResourceRequirements{}, false
+	}
+
+	annotated := &kubernetes.AnnotatedResources{}
+	if err := annotated.Unmarshal([]byte(data)); err != nil {
+		log.Error("failed to unmarshal annotated resources for pod %s: %v",
+			pod.PrettyName(), err)
+		return v1.ResourceRequirements{}, false
+	}
+
+	resources, ok := annotated.Containers[c.GetName()]
+	return resources, ok
+}
+
 // Estimate resource requirements using the containers cgroup parameters and QoS class.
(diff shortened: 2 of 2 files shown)


cmd/resource-annotator/webhook.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+type Webhook struct {
+	logger.Logger
+	cfg *Config
+	srv *http.Server
+}
+
+func NewWebhook(cfg *Config) (*Webhook, error) {
+	cert, err := tls.LoadX509KeyPair(cfg.CertFile, cfg.KeyFile)
```
askervin, line 68: It seems we could drop 68-71 and give cert files in w.srv.ListenAndServeTLS(w.cfg.CertFile, w.cfg.KeyFile)... but this way we get an error a bit earlier and add our own prefix to the error message, so I suppose there are good reasons to keep these lines, too.

Consider this as a note, not a suggestion.

askervin (approved):
LGTM. Thanks @klihub, this is great!

cmd/resource-annotator/webhook.go
klihub: Yes, it was to catch that particular error rather early than a returned error from `ListenAndServer*()`.

PR #621, opened 2026-01-28
e2e/balloons: update expected metrics pattern.

klihub:
Update expected metrics pattern. Metrics names changed when we switched to OpenTelemetry for native metrics collection.


askervin (approved):
LGTM

PR #625, opened 2026-01-29
helm,resource-annotator: add chart support for cert-manager.

klihub:
Add support for HTTPS certificate generation and injection by cert-manager. Also, update default key and certificate names to match the ones injected by cert-manager. Finally, let the stock golang HTTP implementation know about the certificate and key file location and take them into use on its own accord, in the hope that it will also watch and reload certificates when cert-manager rotates/renews them.

### DIFF
--- a/deployment/helm/resource-annotator/README.md
+++ b/deployment/helm/resource-annotator/README.md
@@ -15,15 +15,22 @@ to discover container resource requirements instead of estimating them.
 
 ## Installing the Chart
 
+### Manually Generated HTTPS Certificate
+
 Path to the chart: `resource-annotator`
 
-At the moment the webhook does not you cert-manager. Instead you need
-to generate a certificate for the webhook before instantiating it and
-pass the certificate and its related key to helm. The below example
-demonstrates how this can be done.
+For setting up HTTPS access to the webhook, you can either provide a
+certificate and private key yourself, or you can reference a cert-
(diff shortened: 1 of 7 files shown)


deployment/helm/resource-annotator/README.md
```diff
@@ -15,28 +15,91 @@ to discover container resource requirements instead of estimating them.
...
+these commands:
+
+```shell
+# Install cert-manager, if you don't have it yet.
+$ helm install cert-manager oci://quay.io/jetstack/charts/cert-manager \
+       --version v1.19.2 --namespace cert-manager --create-namespace \
+       --set crds.enabled=true --set crds.keep=false
+
+# Bootstrap a local issuer for cert-manager if you don't have one yet.
+$ cat ./ca-bootstrap.yaml
```
askervin: A file has its benefits, but copy-pasteable commands that would not create any files would have other benefits... I'd anyway prefer copy-pasteable command sequence, with or without creating a file. Like,
```
cat > ca-bootstrap.yaml <<EOF
...
EOF
```
or without a file
```
kubectl apply -f - <<EOF
...
EOF
```

klihub: Updated accordingly to use a Here String.

fmuyassarov (approved):
LGTM

PR #626, opened 2026-02-03
topology-aware: initial scheduling class support.

klihub:
Notes: This PR is stacked on top of #618.

This patch series implements initial scheduling class support for the topology-aware policy. In particular, the PR 
- adds known scheduling classes to the topology aware configuration
- defines a new annotation key for annotating containers or pods to a scheduling class
- implements assigning Guaranteed QoS class containers with exclusive CPU allocation to annotated classes
- adds test cases for scheduling class assignment
- documents the current level of support for scheduling classes

### DIFF
--- a/docs/resource-policy/policy/topology-aware.md
+++ b/docs/resource-policy/policy/topology-aware.md
@@ -150,6 +150,42 @@ behavior. These options can be supplied as part of the effective
     burstability. The policy will try to allocate Burstable containers with
     no CPU limit to a pool at this topology level. The possible values are:
     `system`, `package`, `die`, `numa`.
+- `schedulingClasses`
+  - Define scheduling classes recognized by the policy. A scheduling class
+    has the following set of associated Linux scheduling policy and I/O
+    priority attributes
+    - `name` is the name of the scheduling class.
+    - `policy` is the Linux scheduling policy. Supported policies are:
+    `none`, `other`, `fifo`, `rr`, `batch`, `idle`, and `deadline`.
+    - `priority` is the scheduling priority. Refer to
+      sched_setscheduler(2) documentation for valid values depending on
+      the policy.
+    - `flags` is a list of scheduling flags. Supported flags are:
+      `reset-on-fork`, `reclaim`, `dl-overrun`, `keep-policy`,
+      `keep-params`, `util-clamp-min`, `util-clamp-max`.
+    - `nice`: nice value for the container process.
+    - `runtime`: runtime value for `deadline` scheduling policy (in
+      microseconds).
+    - `deadline`: deadline value for `deadline` scheduling policy (in
+      microseconds).
+    - `period`: period value for `deadline` scheduling policy (in
+      microseconds).
+    - `ioClass`: IO class for the container process. Supported classes
+      are: `none`, `rt` for realtime, `be` for best-effort, and `idle`.
+    - `ioPriority`: IO priority for the container process. Refer
+      to ionice(1) documentation for valid values.
+    These attributes are applied to containers which get assigned to the
+    class. Use the `scheduling-class.resource-policy.nri.io` annotation
+    key to annotate a pod or a container to a class.
+- `namespaceSchedulingClasses`
+  - can assign default scheduling classes to namespaces. If a container
+    is not annotated to use a specific scheduling class but its namespace
+    has a default scheduling class, this will apply to the container.
+- `podQoSSchedulingClasses`
+  - can assign default scheduling classes to Pod QoS classes. If container
+    is neither annotated to use a specific scheduling class nor its namespace
+    has a default scheduling class, but its Pod QoS class has a default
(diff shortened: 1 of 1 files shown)


docs/resource-policy/policy/topology-aware.md
```diff
@@ -150,6 +150,36 @@ behavior. These options can be supplied as part of the effective
...
+      microseconds).
+    - `period`: period value for `deadline` scheduling policy (in
+      microseconds).
+    - `ioClass`: IO class for the container process. Supported classes
+      are: `none`, `rt` for realtime, `be` for best-effort, and `idle`.
+    - `ioPriority`: IO priority for the container process. Refer
+      to ionice(1) documentation for valid values.
+    These attributes are applied to containers which get assigned to the
+    class. Use the `scheduling-class.resource-policy.nri.io` annotation
+    key to annotate a pod or a container to a class. Currently scheduling
```
askervin: Including Burstable and BestEffort containers could have additional benefits.

Guaranteed containers with exclusive CPUs have no kernel scheduling level competition on using CPU time, so they probably benefit most from I/O priority and realtime scheduling policy for faster wakeups.

On the other hand, separate scheduling classes would help prioritizing a responsive (online request-handling) Burstable container over an offline AI-agent Burstable, that would be fine using idle CPU cycles. Because these Burstables may run using the same CPU pool, scheduling policy and priority makes a difference.

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -552,6 +553,19 @@ func (p *policy) applyGrant(grant Grant) {
 		}
 		container.SetCpusetMems(mems.MemsetString())
 	}
+
+	if name := schedulingClassPreference(container); name != "" {
+		if sc := p.getSchedulingClass(name); sc == nil {
+			log.Errorf("%s: unknown scheduling class %q", container.PrettyName(), name)
+		} else {
+			if grant.SharedPortion() > 0 {
```
askervin: ...as in previous comment, I might have applied annotated scheduling class without any hesitation. But that's because I'm always eager giving a gun to the user who asks it, not because it would necessarily be wise. You probably have good reasons, like we don't want realtime BestEffort workloads to eat all available CPUs...

docs/resource-policy/policy/topology-aware.md
klihub: Yes, I agree that for exclusive Guaranteed probably I/O prioritization and realtime scheduling are the ones that probably make sense. For the rest, I wanted to hear some feedback/ideas.

One idea I was contemplating is a per-QoS class default scheduling class. IOW, you could assign by configuration a default scheduling class to (maybe to BestEffort, but definitely to) Burstable and Guaranteed QoS classes, and maybe a separate default one to 'GuaranteedExclusive', Guaranteed with exclusive CPUs, and those classes would get applied to all those containers in a matching QoS class which are not annotated otherwise. A  fourth possibility would be a 'Mixed' pseudoclass in case of the TopologyAware policy, representing containers with a mixture of exclusive and shared pool allocation.

But based on this discussion, I think I need to update this PR to allow at least scheduling classes for Burstable and non-exclusive Guaranteed QoS class containers.

cmd/plugins/topology-aware/policy/pools.go
klihub: No, I didn't have any real strong arguments that you would call 'good reasons', other than just wanting to start somewhere and then get some feedback...

test/e2e/policies.test-suite/topology-aware/n4c16/test17-scheduling-classes/code.var.sh
```diff
@@ -0,0 +1,169 @@
...
+
+# Annotated scheduling class for Guaranteed QoS containers with exclusive CPU allocation.
+CPU=1 MEM=100M \
+    ANN0="scheduling-class.resource-policy.nri.io/container.pod0c0: realtime" \
+    ANN1="scheduling-class.resource-policy.nri.io: idle" \
+        CONTCOUNT=2 create guaranteed
+report allowed
+
+verify 'len(cpus["pod0c0"]) == 1'
+expected_poliy=$SCHED_FIFO expected_prio=$((99 - 42)) verify-sched pod0c0
```
askervin: s/expected_poliy/expected_policy/g

askervin (commented):
One typo. I hope the tests will still pass after it. ;)

Otherwise, looks great.

test/e2e/policies.test-suite/topology-aware/n4c16/test17-scheduling-classes/code.var.sh
klihub: Argh. Thank you for spotting that. Fixed... and also added a check to spot omitted expected arguments, since in these tests we never omit them.

kad (approved):
lgtm

PR #628, opened 2026-02-06
docs: restructure topology-aware documentation.

klihub:
This PR is an attempt to restructure the topology-aware documentation. I threw copilot at it asking to reorganize the content taking the ToC from the corresponding balloons PR#627.

### DIFF
--- a/docs/resource-policy/policy/topology-aware.md
+++ b/docs/resource-policy/policy/topology-aware.md
@@ -1,6 +1,8 @@
 # Topology-Aware Policy
 
-## Background
+## Overview
+
+### What Problems Does the Topology-Aware Policy Solve?
 
 On server-grade hardware the CPU cores, I/O devices and other peripherals
 form a rather complex network together with the memory controllers, the
@@ -26,7 +28,19 @@ of resources for optimal workload performance requires identifying and
 understanding the multiple dimensions of access latency locality present
 in hardware or, in other words, hardware topology awareness.
 
-## Overview
+The topology-aware policy addresses these challenges by:
+
+- **Hardware topology awareness**: Automatically builds a tree of pools based on
+  detected CPU physical hardware topology (sockets and dies) and logical memory
+  hardware topology (NUMA nodes)
+- **Aligned resource allocation**: Assigns CPU, memory, and devices with optimal
+  topological alignment
+- **Multi-tier memory support**: Handles DRAM, PMEM, and HBM memory types
+- **Flexible CPU allocation**: Supports shared, exclusive, and mixed CPU core assignments
+- **Device locality**: Considers device connection topology to CPU and memory when
+  placing workloads
+
+### How the Topology-Aware Policy Works
 
 The `topology-aware` policy automatically builds a tree of pools based on the
 detected hardware topology. Each pool has a set of CPUs and memory zones
@@ -39,8 +53,7 @@ dies, sockets, and finally the whole of the system at the root node. Leaf NUMA
 nodes are assigned the memory behind their controllers / zones and CPU cores
 with the smallest distance / access penalty to this memory. If the machine
 has multiple types of memory separately visible to both the kernel and user
-space, for instance both DRAM and
-[PMEM](https://www.intel.com/content/www/us/en/products/memory-storage/optane-dc-persistent-memory.html),
+ space, for instance both DRAM and PMEM,
 each zone of special type of memory is assigned to the closest NUMA node pool.
 
 Each non-leaf pool node in the tree is assigned the union of the resources of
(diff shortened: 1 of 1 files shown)


docs/resource-policy/policy/topology-aware.md
```diff
@@ -83,40 +95,98 @@ implementation evolves, its basic principles are roughly
...
+  - Assign both exclusive and shared CPU cores to workload
+- **Kernel-isolated CPU support** (['isolcpus'](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html#cpu-lists))
+  - Use kernel-isolated CPU cores for exclusively assigned CPU cores
+- **Resource exposure**
+  - Expose assigned resources to workloads and notify about changes
+- **Dynamic memory relaxation**
+  - Dynamically widen workload memory set to avoid pool/workload OOM
+- **Multi-tier memory allocation**
+  - Assign workloads to memory zones of their preferred type
+  - Support for DRAM, PMEM (such as [Intel® Optane™ memory](https://www.intel.com/content/www/us/en/products/memory-storage/optane-dc-persistent-memory.html)), and [HBM](https://en.wikipedia.org/wiki/High_Bandwidth_Memory)
```
askervin: Link to Optane does not work. Couldn't (quickly) find any better option in wikipedia either. Perhaps "Intel Optane memory" is ok without a link...

docs/resource-policy/policy/topology-aware.md
```diff
@@ -1,6 +1,8 @@
 # Topology-Aware Policy
 
-## Background
+## 1. Overview
+
+### What Problems Does the Topology-Aware Policy Solve?
```
askervin, line 5: I really like this introduction/background on why resource policies matter in the first place. Let's take it in.

In a later phase, when possibly adding a separate "Why to use a resource policy and how to choose the right one" (sub)document, perhaps we can move this background information there and include links from the topology-aware and balloons policy documents into that.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -83,40 +95,98 @@ implementation evolves, its basic principles are roughly
...
+- **Configuration options** covers detailed configuration options
+- **Cookbook** provides recipes for common use cases
+- **Troubleshooting** offers troubleshooting guidance
+
+### Integration with Kubernetes
+
+The topology-aware policy integrates with Kubernetes through the
+[Node Resource Interface (NRI)](https://github.com/containerd/nri).
+It uses Kubernetes Pod and Container resource annotations to control
+resource allocation behavior.
```
askervin: Might be useful to mention running as a DaemonSet that is configured through TopologyAwarePolicies CRs, pod/container-specific configuration through pod annotations, possibly scheduler integration via NRTs. (The last one I hesitated in balloons because I haven't tried that out myself...)

docs/resource-policy/policy/topology-aware.md
```diff
@@ -83,40 +95,98 @@ implementation evolves, its basic principles are roughly
...
+
+Replace `kube-system` with the namespace where the plugin is deployed.
+
+### Configuration Scopes
+
+The policy supports different configuration scopes:
+
+- **Cluster-wide defaults**: Applied to all nodes unless overridden
+- **Node-specific overrides**: Applied to specific nodes based on labels
+- **Namespace-specific settings**: Applied to containers in specific namespaces
```
askervin: Configurations are indeed namespace-specific, but it requires that resource policy DaemonSets are deployed in multiple namespaces, doesn't it? And cluster nodes are grouped into namespaces, too, so that too policy plugins never run on the same node.... I haven't thought about that kind of deployment.

On the other hand, this might be AI-generated thingy. Would it be better to go with
- cluster
- node group (labels)
- node-specific (node name)
configuration levels without mentioning namespaces?

klihub: Yeah, it was just hallucinating... I haven't really had the time/energy to go through this yet. Thanks for proof-reading it !

docs/resource-policy/policy/topology-aware.md
klihub: I also could not find that original document, most accessible references are now to M.2 and PCI express SSDs. So I updated simply by removing the link.

docs/resource-policy/policy/topology-aware.md
klihub: If we had a sample scheduler plugin or some other component using NRT information published by our policies to guide scheduling, I'd be keen on mentioning it. Until then, not so sure. It's a bit different for balloons, because IIRC you push to NRT debugging information, so it might be useful for the user in some cases.

askervin: I wonder if we could have been able to piggybag this scheduler plugin:
https://scheduler-plugins.sigs.k8s.io/docs/plugins/noderesourcetopology/
...but our NRT's might not be too compatible with what that plugins expects.

And in any case, as the direction is towards DRA that'll solve scheduling issues properly (without races), I suppose the usefulness of NRT might be more in status/debugging ("show me created pools / CPU affinity of containers") rather than scheduling...

docs/resource-policy/policy/topology-aware.md
```diff
@@ -83,40 +94,107 @@ implementation evolves, its basic principles are roughly
...
+
+### Prerequisites
+
+- Kubernetes cluster with NRI-enabled container runtime
+- NRI plugins support enabled in the container runtime configuration
+
+### Installing with Helm
+
+The topology-aware policy can be installed using Helm charts. Refer to the
+[topology-aware Helm documentation](../../deployment/helm/topology-aware.md) for detailed instructions.
```
askervin, line 151: following this link in github.com/containers/nri-plugins repo docs takes to an empty page. Might work in generated documentation due to include directive in the linked target file. Writing this comment in the hopes that it reminds me to check  the docs site when the documentation is rebuilt with this PR merged.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -83,40 +94,107 @@ implementation evolves, its basic principles are roughly
...
+
+### Managing Configuration with kubectl
+
+The policy configuration can be managed using kubectl and the
```
askervin: It would feel natural to mention the name of the resource here. TopologyAwarePolicy.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -341,7 +354,8 @@ groups, are all pinned to run on the shared subset of CPU cores in the
 container's assigned pool. This shared subset can and usually does change
 dynamically as exclusive CPU cores are allocated and released in the pool.
 
-## Container CPU Allocation Preference Annotations
+
+### Container Annotations
```
askervin: The following paragraph used to be a short intro in the beginning of the first 3rd level section under new 2nd level section. Now the 2nd level section is changed into its own 3rd level section that sort of breaks the story of CPU allocation configuration around it.

I'd see two options: let's remove "### Container Annotations" and move the next paragraph "one line down" right under the next section header.

Alternatively, rename the "### Container Annotations" to "### Pod and Container Configuration by Annotations" or something and extend this a bit to give general principles how Kubernetes Pod Annotations override general configuration options for all and specific containers in a pod. That would also give a reason for using "container annotations" and "pod annotations" almost interchangeably in the document (possibly "container annotation" meaning a "kubernetes pod annotation targeted to a named container).

docs/resource-policy/policy/topology-aware.md
```diff
@@ -535,7 +615,16 @@ memory.preserve.resource-policy.nri.io/pod: "true"
...
 ```
 
-## Cold Start
+
+### Memory Configuration
+
+It is not possible for the policy to accurately determine memory requests for
+pods in the `Burstable` QoS class. If high accuracy is critical for such containers
+you can annotate the pod with exact per container resource requirements, or use the
+resource annotator web hook to do this for you. See the related Helm chart
```
askervin: s/web hook/webhook

docs/resource-policy/policy/topology-aware.md
```diff
@@ -615,7 +696,8 @@ metadata:
     prefer-reserved-cpus.resource-policy.nri.io/container.special: "false"
 ```
 
-## Controlling Topology Hints Via Annotations
+
+### Topology Hints
```
askervin, line 664: The next paragraph used to be again an introduction to topology hints in general, to cover the subsection. But now it became a sub-subsection itself. I still assume we wouldn't like to go to 4th level sections, so perhaps these two subsection ("topology hints" and "enabling or disabling...") could b merged.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -774,7 +856,8 @@ of the devices, the policy falls back to picking resource from the pool without
 considering device hints.
 
 
-## Container Affinity and Anti-Affinity
+
+### Container Affinity and Anti-Affinity
 
 ### Introduction
```
askervin: Remove "Introduction" title, otherwise we have an empty 3rd level section above.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -1059,7 +1142,169 @@ metadata:
...
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  annotations:
+    # Prefer exclusive CPUs; opt out from shared allocation
+    prefer-shared-cpus.resource-policy.nri.io/pod: "false"
+    # Prefer isolated CPUs when available
+    prefer-isolated-cpus.resource-policy.nri.io/pod: "true"
```
askervin: Perhaps we should not recommend isolcpus for generic HPC?

docs/resource-policy/policy/topology-aware.md
```diff
@@ -1059,7 +1142,169 @@ metadata:
...
+```
+
+```yaml
+apiVersion: config.nri/v1alpha1
+kind: TopologyAwarePolicy
+metadata:
+  name: default
+spec:
+  preferSharedCPUs: false
+  preferIsolatedCPUs: true
```
askervin: (same as above, isolcpus might be difficult...)

docs/resource-policy/policy/topology-aware.md
```diff
@@ -1059,7 +1142,169 @@ metadata:
...
```
askervin: Suspecting that isolcpus for any Guaranteed container as a default is not a safe choice.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -1059,7 +1142,169 @@ metadata:
...
+  - name: realtime-app
+    resources:
+      requests:
+        cpu: "4"
+        memory: "8Gi"
+      limits:
+        cpu: "4"
+        memory: "8Gi"
+
+# Best-effort background task
```
askervin, line 1185: Could we have one more workload in this cookbook? I'm thinking of a high-responsive, relatively low-latency Burstable container. We could treat it with slightly elevated scheduling priority (yet not realtime), and with memory locality set to NUMA level. That would be again clearly something that is not achievable with on-the-stock managers.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -26,7 +28,17 @@ of resources for optimal workload performance requires identifying and
 understanding the multiple dimensions of access latency locality present
 in hardware or, in other words, hardware topology awareness.
 
-## Overview
+The topology-aware policy addresses these challenges by:
+
+- **Hardware topology awareness**: Automatically builds a tree of pools based on
```
kad, line 33: I'd suggest to slightly adjust this phrase:
`... on detected CPU physical hardware topology (sockets and dies) and logical memory hardware topology (NUMA nodes).`

docs/resource-policy/policy/topology-aware.md
```diff
@@ -26,7 +28,17 @@ of resources for optimal workload performance requires identifying and
...
+  detected hardware topology (NUMA nodes, dies, sockets)
+- **Aligned resource allocation**: Assigns CPU, memory, and devices with optimal
+  topological alignment
+- **Multi-tier memory support**: Handles DRAM, PMEM, and HBM memory types
+- **Flexible CPU allocation**: Supports shared, exclusive, and mixed CPU core assignments
+- **Device locality**: Considers device topology when placing workloads
```
kad: `Considers device connection topology to CPU and Memory when placing workloads`

docs/resource-policy/policy/topology-aware.md
klihub: I think that's because we use inclusion which is not supported on github. It should work in the generated documenttion (I checked it with `make site-serve`) and points to a version of [this page](https://containers.github.io/nri-plugins/stable/docs/deployment/helm/topology-aware.html).

docs/resource-policy/policy/topology-aware.md
klihub: Updated to mention the resource name.

docs/resource-policy/policy/topology-aware.md
klihub: Updated using the first option, IOW folding paragraph into the beginning of the next section.

docs/resource-policy/policy/topology-aware.md
klihub: Yeah, this is a bad/hallucinated example. Usually shared preference should be set or left at the default true so that isolated CPUs would be preferred only for guaranteed containers with < 2 CPU request. Plus globally preferred isolated for all exclusive allocation requires explicitly annotated opt-out from shared allocation. We need to check and rethink the examples more carefully.

docs/resource-policy/policy/topology-aware.md
klihub: @askervin Shouldn't we also move the earlier 'Implicit Hardware Topology Hints' section here and combine it with the rest so we'd have everything related to hints described together/close to each other ?

askervin: Sounds good. It's easier to find the information under only one topic/section.

klihub: I ended up moving the first section about (Implicit Hardware) Topology Hints together with the rest later, turning them into a single chapter and using bold paragraph headers as topic separators.

docs/resource-policy/policy/topology-aware.md
klihub: Okay, I added a prioritized app like that, and changed the realtime app to take a single exclusive CPU and explicitly prefer isolated CPUs for it.

docs/resource-policy/policy/topology-aware.md
klihub: I updated this Guaranteed realtime app to ask for a single CPU, to let the isolated CPU preference make more sense.

docs/resource-policy/policy/topology-aware.md
klihub: I removed the HPC workload example altogether. It did not make sense with the isolated allocation attempt of multiple CPUs.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -1059,7 +1134,157 @@ metadata:
...
+        cpu: "2"
+        memory: "4Gi"
+  - name: backend
+    resources:
+      requests:
+        cpu: "2"
+        memory: "4Gi"
+```
+
+For more affinity options, see Section 3.8.
```
askervin: (Following has extra spaces around #, otherwise the suggestion becomes a link and thus unreadable in this comment.)

s/Section 3.8/
[Container Affinity and Anti-Affinity]( # container-affinity-and-anti-affinity)

askervin (commented):
One last thing, sorry about not catching this earlier. I think this LGTM then.

docs/resource-policy/policy/topology-aware.md
klihub: Updated.

PR #631, opened 2026-02-16
topology-aware: add 'excluded cpuset' config notation

klihub:
Add support for a semantically inverted configuration notion for `availableResources` and `reservedResources`, where instead of specifying CPUs to include one can specify the set of CPUs to exclude. For `availableResources` exclusion wrt. all the CPUs present on the system. For `reservedResources` exclusion is wrt. `availableResources/cpu`.


askervin (approved):
LGTM

askervin:
Thanks for super-quick reaction on this issue, @klihub. This already LGTM, even though it's as Draft.

PR #633, opened 2026-02-18
operator: bypass ansible-operator proxy for Helm deployments

fmuyassarov:
Fixes the broken operator and allow it to install template plugin as well.


fmuyassarov:
Let me try to merge, to check if I can reproduce the issue.

PR #638, opened 2026-03-05
topology-aware: implement strict hint and CPU isolation preferences.

klihub:
This patch series adds
- support for annotating containers for strict interpretation of topology hints
- support for strictly required isolated CPUs
- support for injecting test topology hints into containers (when extra test APIs are enabled)
- e2e test cases for these new strict preferences

### DIFF
--- a/test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
+++ b/test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
@@ -1,0 +1,121 @@
+
+cleanup() {
+    vm-command "kubectl delete pods --all --now"
+    vm-command "kubectl delete namespaces highprio lowprio --now --ignore-not-found"
+}
+
+cleanup
+helm-terminate
+helm_config=$(instantiate helm-config.yaml) helm-launch topology-aware
+
+# Create a container with a strict test hint for NUMA node 2.
+ANN0="strict.topologyhints.resource-policy.nri.io/container.pod0c0: 'true'"
+ANN1="test.topologyhints.resource-policy.nri.io/container.pod0c0: '{ test: { NUMAs: \"2\" } } }'"
+
+CONTCOUNT=1 CPU=1 MEM=100M create guaranteed
+report allowed
+verify \
+    'cpus["pod0c0"].issubset({"cpu08", "cpu09", "cpu10", "cpu11"})' \
+    'node_ids(nodes["pod0c0"]) == {2}'
+
+# Create a container with a strict test hint for NUMA node 3.
+ANN0="strict.topologyhints.resource-policy.nri.io/container.pod1c0: 'true'"
+ANN1="test.topologyhints.resource-policy.nri.io/container.pod1c0: '{ test: { NUMAs: \"3\" } } }'"
+
+CONTCOUNT=1 CPU=1 MEM=100M create guaranteed
+report allowed
+verify \
+    'cpus["pod1c0"].issubset({"cpu12", "cpu13", "cpu14", "cpu15"})' \
+    'node_ids(nodes["pod1c0"]) == {3}'
+
+# Create a container with a strict test hint for NUMA node 2.
+ANN0='strict.topologyhints.resource-policy.nri.io/container.pod2c0: "true"'
+ANN1='test.topologyhints.resource-policy.nri.io/container.pod2c0: "{ test: { NUMAs: \"2\" } } }"'
+CONTCOUNT=1 CPU=1 MEM=100M create guaranteed
+report allowed
+verify \
+    'cpus["pod2c0"].issubset({"cpu08", "cpu09", "cpu10", "cpu11"})' \
+    'node_ids(nodes["pod2c0"]) == {2}'
+
+# Try to create a container with a strict test hint NUMA node 2.
+# This one should not fit (not enough CPU left) and should fail with a
+# strict hint check error.
+ANN0='strict.topologyhints.resource-policy.nri.io/container.pod3c0: "true"'
+ANN1='test.topologyhints.resource-policy.nri.io/container.pod3c0: "{ test: { NUMAs: \"2\" } } }"'
+wait="" CONTCOUNT=1 CPU=3 MEM=100M create guaranteed
+
+PodReadyCond='condition=PodReadyToStartContainers'
+vm-command "kubectl wait --timeout=5s pod pod3 --for=$PodReadyCond" || {
+    error "failed to wait for pod3 to start containerd"
+}
+
+vm-command "kubectl get pod pod3 -o jsonpath='{.status.containerStatuses[0].state}' | \
+        grep -q \"fail strict hint\"" || {
+    error "pod3c0 unexpectedly passed strict topology hint check"
+}
+
+# Try to create a container with a strict test hints for NUMA node 2 and
+# required isolated CPUs. This one would fit but there are not isolated
+# CPUs so it should fail with a strict isolation preference error.
+ANN0='strict.topologyhints.resource-policy.nri.io/container.pod4c0: "true"'
+ANN1='test.topologyhints.resource-policy.nri.io/container.pod4c0: "{ test: { NUMAs: \"2\" } } }"'
+ANN2='require-isolated-cpus.resource-policy.nri.io/container.pod4c0: "true"'
+wait="" CONTCOUNT=1 CPU=1 MEM=100M create guaranteed
+
+PodReadyCond='condition=PodReadyToStartContainers'
+vm-command "kubectl wait --timeout=5s pod pod4 --for=$PodReadyCond" || {
+    error "failed to wait for pod4 to start containerd"
+}
+
+vm-command "kubectl get pod pod4 -o jsonpath='{.status.containerStatuses[0].state}' | \
+        grep -q \"isolated CPUs\"" || {
+    error "pod4c0 unexpectedly passed strict isolated CPU requirement check"
+}
+
+# Create a container with a strict test hint for NUMA node 2 and preference
+# for isolated CPUs. This one should fit and succeed because the unfulfilled
+# preference is not a strict requirement.
+ANN0='strict.topologyhints.resource-policy.nri.io/container.pod5c0: "true"'
+ANN1='test.topologyhints.resource-policy.nri.io/container.pod5c0: "{ test: { NUMAs: \"2\" } } }"'
+ANN2='prefer-isolated-cpus.resource-policy.nri.io/container.pod5c0: "true"'
+CONTCOUNT=1 CPU=1 MEM=100M create guaranteed
+report allowed
+verify \
+    'cpus["pod5c0"].issubset({"cpu08", "cpu09", "cpu10", "cpu11"})' \
+    'node_ids(nodes["pod5c0"]) == {2}'
+
+vm-command "kubectl delete pod pod5 --now"
+
+# Now recreate pod4 test but with more complex effective annotation.
+# Try to create a container with a strict test hints for NUMA node 2 an
+# required isolated CPUs. This one would fit but there are not isolated
+# CPUs so it should fail with a strict isolation preference error.
+ANN0='strict.topologyhints.resource-policy.nri.io/container.pod6c0: "true"'
+ANN1='test.topologyhints.resource-policy.nri.io/container.pod6c0: "{ test: { NUMAs: \"2\" } } }"'
+ANN2='prefer-isolated-cpus.resource-policy.nri.io/pod: "false"'
+ANN3='require-isolated-cpus.resource-policy.nri.io/container.pod6c0: "true"'
+wait="" CONTCOUNT=1 CPU=1 MEM=100M create guaranteed
+
+PodReadyCond='condition=PodReadyToStartContainers'
+vm-command "kubectl wait --timeout=5s pod pod6 --for=$PodReadyCond" || {
+    error "failed to wait for pod6 to start containerd"
+}
+
+vm-command "kubectl get pod pod6 -o jsonpath='{.status.containerStatuses[0].state}' | \
+        grep -q \"isolated CPUs\"" || {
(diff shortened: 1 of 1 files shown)


pkg/resmgr/cache/container.go
```diff
@@ -613,6 +632,22 @@ func (c *container) GetTopologyHints() topology.Hints {
 	return c.TopologyHints
 }
 
+func (c *container) StrictTopologyHints() bool {
```
askervin, line 643: Do you think it would be possible (without causing complexity explosion) to handle strictness as a property of a topology hint, instead of an on-off-property of all hints generated for a container?

The reason I'm asking is that there might be topology hints that a user may not be aware of, or that might be nice-to-have but definitely not strictly required. In that case, annotating all hints strict might cause unexpected/unnecessary container creation failures, even though it would look like that the hint that the user considers the (only) important one should be fine.

For example: `strict.topologyhints.resource-policy.nri.io/pod: pod-resource,isolcpus` for two types of hints, or `strict.topologyhints.resource-policy.nri.io/container.c0: "*"` for the same effect as "true" now, but giving a hint to the user, that you could be more specific, too.

klihub: > Do you think it would be possible (without causing complexity explosion) to handle strictness as a property of a topology hint, instead of an on-off-property of all hints generated for a container?
> 
> The reason I'm asking is that there might be topology hints that a user may not be aware of, or that might be nice-to-have but definitely not strictly required. In that case, annotating all hints strict might cause unexpected/unnecessary container creation failures, even though it would look like that the hint that the user considers the (only) important one should be fine.
> 
> For example: `strict.topologyhints.resource-policy.nri.io/pod: pod-resource,isolcpus` for two types of hints, or `strict.topologyhints.resource-policy.nri.io/container.c0: "*"` for the same effect as "true" now, but giving a hint to the user, that you could be more specific, too.

At least for now I would prefer not to try adding it in the first version.
We already have a mechanism to disable topology hints, selectively based on types/sources (mounts, devices, pod resource API), or more specifically using paths/globbing. I think for now it would be a better idea to just disable any unwanted hints if you go for strict alignment. For instance, just disable all mount and device hints, if you want strict alignment based on hints gathered from the pod resource API, etc.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -129,8 +131,12 @@ func boolConfigPreference(ptr *bool) (bool, prefKind) {
 // for the given container. If an effective annotation is not found, it uses
 // the global configuration for isolated CPU preference.
 func isolatedCPUsPreference(pod cache.Pod, container cache.Container) (bool, prefKind) {
-	key := preferIsolatedCPUsKey
+	key := strictPreferIsolatedCPUsKey
 	value, ok := pod.GetEffectiveAnnotation(key, container.GetName())
+	if !ok {
+		key = strictPreferIsolatedCPUsKey
```
askervin: I'd expect that `if !ok` branch should re-check effective annotation with
```
key =  preferIsolatedCPUsKey
```
(and wish that we would have had an e2e test that would have catched this...)

docs/resource-policy/policy/topology-aware.md
```diff
@@ -462,6 +462,11 @@ metadata:
 These Pod annotations have no effect on containers which are not eligible for
 exclusive allocation.
 
+The `require-isolated-cpus.resource-policy.nri.io` annotation key can be
```
askervin, line 465: In the implementation (given that my guess for the key order is correct), require-isolated-cpus annotation overrides whatever prefer-isolated-cpus might have had. Even so that if there is a pod-level require-isolated-cpus, then it overrides container-specific prefer-isolated-cpus.

Currently the documentation could be miss-interpreted so that "require-isolated-cpus" makes "prefer-isolated-cpus" a requirement. I'd suggest rephrasing this so that nobody would use both, if possible. Maybe something like:

> If creating a container must fail with an error when there are not enough isolated CPUs for it, use `require-isolated-cpus.resource-policy.nri.io` annotation key instead of `prefer-isolated-cpus`.

test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
```diff
@@ -0,0 +1,85 @@
...
+# for isolated CPUs. This one should fit and succeed because the unfulfilled
+# preference is not a strict requirement.
+ANN0='strict.topologyhints.resource-policy.nri.io/container.pod5c0: "true"'
+ANN1='test.topologyhints.resource-policy.nri.io/container.pod5c0: "{ test: { NUMAs: \"2\" } } }"'
+ANN2='prefer-isolated-cpus.resource-policy.nri.io/container.pod5c0: "true"'
+CONTCOUNT=1 CPU=1 MEM=100M create guaranteed
+report allowed
+verify \
+    'cpus["pod5c0"].issubset({"cpu08", "cpu09", "cpu10", "cpu11"})' \
+    'node_ids(nodes["pod5c0"]) == {2}'
```
askervin, line 85: Do you think we could have a `prefer-isolated-cpus` test step, too, without `require-isolated-cpus` present in the same pod?

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Argh, copypaste'o crept in. Thanks for spotting it.

test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
klihub: Yes, based on the existing bugs in the code, it is more than warranted to add test cases to verify correct semantics for the coexistence of both keys in the annotations.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Fixed isolation preference semantics. But more properly also considering the co-existence of soft and hard preferences with different scopes.

test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
klihub: Actually I think what we really need to test is the semantics of co-existence of those  two annotations, but at different (pod vs. container) scopes.

And I think we should start considering for future preference annotations how the current syntax for a base key could be extended, in addition to the current scoping, with preference vs. requirement semantics or other semantics, probably using as a prefix to the key.

So something like

key=$key
preference: prefer.$key/$scope
requirement: require.$key/$scope

Then this would become:

prefer.isolated-cpus/pod: "true"
require.isolated-cpus/container.ctr0: "true"
prefer.shared-cpus/container.ctr1: "true"

And a querying function, which return all of the nuances related to the semantics of a key which supports semantic prefixing.

Such a scheme would also force us to be more consistent with the annotations.

klihub: Anyway, upated adding more test checking correctness of the interpretation of multiple coexisting keys.

docs/resource-policy/policy/topology-aware.md
klihub: Reworded docs.

test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
klihub: Let me know if you think more tests/combinations are warranted.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -129,12 +131,24 @@ func boolConfigPreference(ptr *bool) (bool, prefKind) {
...
-	key := preferIsolatedCPUsKey
-	value, ok := pod.GetEffectiveAnnotation(key, container.GetName())
-	if !ok {
+	sKey := strictPreferIsolatedCPUsKey
+	sVal, sScope, sOk := pod.QueryEffectiveAnnotation(sKey, container.GetName())
+
+	pKey := preferIsolatedCPUsKey
+	pVal, pScope, pOk := pod.QueryEffectiveAnnotation(pKey, container.GetName())
+
+	if !sOk && !pOk {
```
askervin: super-nit: we have used quite eagerly switch-case structures in the code, even in case of two options only. I think we could continue that habit to handle these three different cases (lines 140-150), too, without if-branching. But this is of course a matter of taste, and I'm happy to leave it to you to decide.

test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
```diff
@@ -0,0 +1,124 @@
...
+vm-command "kubectl wait --timeout=5s pod pod6 --for=$PodReadyCond" || {
+    error "failed to wait for pod6 to start containerd"
+}
+
+vm-command "kubectl get pod pod6 -o jsonpath='{.status.containerStatuses[0].state}' | \
+        grep -q \"isolated CPUs\"" || {
+    error "pod6c0 unexpectedly passed strict isolated CPU requirement check"
+}
+
+verify 'True'
```
askervin: verify 'True' looks like a leftover from debugging...

klihub: Well... I was trying to get us an 'assertion holds' for the test with this... which we do get, but now I'm not sure whether there is any real benefit to it.  So... removed.

test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
klihub: > And I think we should start considering for future preference annotations how the current syntax for a base key could be extended, in addition to the current scoping, with preference vs. requirement semantics or other semantics, probably using as a prefix to the key.
> 
> So something like
> 
> key=$key preference: prefer.$key/$scope requirement: require.$key/$scope
> 
> Then this would become:
> 
> prefer.isolated-cpus/pod: "true" require.isolated-cpus/container.ctr0: "true" prefer.shared-cpus/container.ctr1: "true"
> 
> And a querying function, which return all of the nuances related to the semantics of a key which supports semantic prefixing.
> 
> Such a scheme would also force us to be more consistent with the annotations.

Actually I'm still contemplating implementing this scheme already now, together with some internal translation of existing keys that could use this scheme. @askervin WDYT ?

askervin: I was (once again) delighted seeing how cleanly you were able to write prefer/require for isolated CPUs.

Because I don't see that merging this solution would prevent later on generalizing the approach for other keys, and that generalization would not change the "annotation API" introduced in this PR, I'd prefer leaving the generalization in another PR, if you're ok with that.

klihub: Yes, I am more than okay with that, especially because I took a stab at the implementation, and it is a bit more involved than what I was hoping for such a non-functional change (in the context of this PR).

Although the way how I bolted that PoC in does change the very same extended annotation lookup API introduced as part of this PR in an incompatible way... but maybe we'll cross that bridge one way or another when we get there. 


@askervin Anyway, here is my branch for reference with that PoC implementation as the top-most commit: https://github.com/klihub/nri-plugins/tree/devel/annotation-modifiers

askervin: Without a strong reason why -- perhaps mostly to avoid returning four values from a function -- this PoC made me wonder if both scope and modifiers could be grouped together as "annotation attributes". Then QueryEffectiveAnnotation() could return them in one variable, and possibly use scope(s) as query inputs... if that ever makes sense.

Anyway, that's not really important. The more important thing is: I didn't catch the annotation incompatibility between current implementation and the PoC. Is that something we should still think a bit before merging this PR?

klihub: > Without a strong reason why -- perhaps mostly to avoid returning four values from a function -- this PoC made me wonder if both scope and modifiers could be grouped together as "annotation attributes". Then QueryEffectiveAnnotation() could return them in one variable, and possibly use scope(s) as query inputs... if that ever makes sense.

Yes, that's true. I haven't think of that at all, and it can be that it would make more sense. 

> Anyway, that's not really important. The more important thing is: I didn't catch the annotation incompatibility between current implementation and the PoC. Is that something we should still think a bit before merging this PR?

I think there is no annotation incompatibility. Actually I think it's a bit of the opposite. For annotation notation backward compatibility, the PoC allows two syntaxes: both dot and dash-prefixes, IOW allowing all of `prefer-`, `prefer.`, `require-`, and `require.` as semantic-altering prefixes. In reality I'd rather only allow dash-prefixes for existing keys only and otherwise only allow the `$prefix.$key/$scope` syntax, but I think that could be fairly easily implemented.

Or is this what you were referring to by annotation incompatibility ? Or did you refer to the altered function signature of the annotation querying function which would then be backward incompatible ?

askervin: Yeah, I was thinking exactly the annotation incompatibility that you mentioned. Thanks for explaining. Sticking with <prefix><dash><key> syntax would sound safer to me than allowing more flexibility in separators, or allowing different set of separators with different keys. (Errors in almost correctly written annotations have been already hard enough to spot...)

askervin: Let's take this in now, and generalize require/prefer prefixes (possibly with new keys) in future PRs. Glad to make this improvement available to users already as is.

PR #649, opened 2026-03-30
topology-aware: Skip size arbitration if both offers are nil

ozhuraki:
@klihub: Not sure if this is a real issue outside of the unit tests and a right fix for it. Just ran into panic here when was looking into different thing under tests.

### DIFF
--- a/cmd/plugins/topology-aware/policy/pools.go
+++ b/cmd/plugins/topology-aware/policy/pools.go
@@ -1059,20 +1059,22 @@ func (p *policy) compareScores(request Request, pools []Node, scores map[int]Sco
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


askervin (approved):

klihub (approved):

TonyxSun:
Could we cut a v0.12.3 patch release including the fix from this PR #649 ? 

We're hitting this nil-pointer panic on memory-packed nodes running v0.12.2: the topology-aware plugin panics in compareScores() (cmd/plugins/topology-aware/policy/pools.go) when both memory offers are nil, manifesting as NRI plugin CrashLoopBackOff.

Happy to open the backport PR against release-0.12 if that is easier.

PR #662, opened 2026-04-15
operator: replace deprecated gcr.io/kubebuilder/kube-rbac-proxy

fmuyassarov:
The `gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1` image is deprecated. 
Switch to the actively maintained https://github.com/kube-rbac-proxy/kube-rbac-proxy (which was also mirrored by the deprecated image). No functional changes expected, this just updates to a supported source.

Ref: https://github.com/kubernetes-sigs/cluster-api-addon-provider-helm/issues/318

### DIFF
--- a/deployment/operator/config/default/manager_auth_proxy_patch.yaml
+++ b/deployment/operator/config/default/manager_auth_proxy_patch.yaml
@@ -31,7 +31,7 @@ spec:
           capabilities:
             drop:
               - "ALL"
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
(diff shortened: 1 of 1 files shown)


fmuyassarov:
@klihub this is the reason why the operator tests didn't pass in the operator hub submission. I guess, we would need to cut a new patch release once we have this fix so that we can rollout the working version of operator based on v0.12.x. WDYT?

fmuyassarov:
now that I think more, knowing that there is very less probability of someone consuming the operator if not none, I guess it doesn't qualify to force us to cut a new release, even if it is a patch...I don't know tbh.

klihub (approved):

PR #663, opened 2026-04-16
instrumentation, log: add support for otel log collection.

klihub:
Note: this PR is stacked on #664.

This patch series:
- reworks the internal logging backend using slog, adding support for structured logging
- cleans up logging calls to consistently use the f-variants for formatted/non-structured logging  
- adds support for exporting logs to an opentelemetry compatible collector
- adds a test case of otel log collection
- updates documentation


Makefile
```diff
@@ -396,7 +396,7 @@ ginkgo-tests:
 	    --covermode atomic \
 	    --output-dir $(COVERAGE_PATH) \
 	    --junit-report junit.xml \
-	    --coverprofile $(COVERAGE_PATH)/coverprofile \
```
askervin, line 399: I don't know how to add a comment on commit message directly, but there's a tiny typo. s/converprofile/coverprofile.

pkg/version/version.go
```diff
@@ -31,7 +31,7 @@
 //
 //   1) add the corresponding variables to this modules
 //   2) arrange the default values to be correctly overridden during linking
-//   3) add printing of the new metadata to PrintVersionInfo()
```
askervin, line 34: Perhaps Info()->Infof() in PrintVersionInfo() is unnecessary.

pkg/instrumentation/logging/logging.go
```diff
@@ -0,0 +1,154 @@
...
+	default:
+		log.Infof("exporter effectively disabled, no exporter set")
+		return nil
+	}
+
+	log.Infof("starting logging exporter...")
+
+	provider = sdklog.NewLoggerProvider(
+		sdklog.WithProcessor(
+			sdklog.NewBatchProcessor(exp, sdklog.WithExportInterval(interval)),
```
askervin, line 102: ...checked that if interval==0, this should be fallback to default 1 second interval. Sounds reasonable as is.

Makefile
klihub: Yeah, I suspect it might not be possible. Thanks for spotting it. Fixed.

pkg/version/version.go
klihub: Yeah, that was the result of my sed-butchering the source tree. I tried to remove the unnecessary ones, but I suspect there might be more of the same. Thanks for spotting it. Fixed.

fmuyassarov:
Sorry for delay, I will go through it tomorrow.

kad (commented):
Let's rebase on top of current main, update deps and then it is lgtm from me.

askervin (approved):
LGTM

askervin:
Let's swallow this, close #688 (included here), and sometimes later see if we could have an e2e test for this... Thanks @klihub!

kad (approved):
LGTM

PR #664, opened 2026-04-16
golang,golang-ci-lint: bump versions, fix linting errors.

klihub:
Bump golang and golang-ci-lint versions. Fix linting errors found by updated linter.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-      - uses: golangci/golangci-lint-action@v8
+      - uses: golangci/golangci-lint-action@v9
         with:
-          version: v2.4.0
+          version: v2.11.4
           args: --timeout=10m
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/containers/nri-plugins
 
-go 1.25.0
+go 1.26.0
 
 require (
 	github.com/askervin/gofmbt v0.0.0-20250119175120-506d925f666f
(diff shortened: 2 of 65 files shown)


fmuyassarov:
@klihub would you mind to rebase your branch with HEAD?

.github/workflows/lint.yaml
```diff
@@ -28,5 +28,5 @@ jobs:
           go-version-file: go.mod
       - uses: golangci/golangci-lint-action@v8
         with:
-          version: v2.4.0
+          version: v2.11.4
           args: --timeout=10m
```
fmuyassarov, line 32: I would suggest that we switch to the `golangci/golangci-lint-action@v9` if we don't have any hard dependencies on v8

fmuyassarov (approved):
LGTM

.github/workflows/lint.yaml
klihub: Updated to v9.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containers/nri-plugins
 
-go 1.25.0
+go 1.26.0
```
fmuyassarov, line 4: can we please replicate the changes to the Dockerfile: https://github.com/containers/nri-plugins/blob/90b7a10eaa95027d012bea557f9ecb296b3d6b2b/docs/Dockerfile#L3 and https://github.com/containers/nri-plugins/blob/90b7a10eaa95027d012bea557f9ecb296b3d6b2b/Makefile#L40 ?

fmuyassarov: and all other dockerfiles from each plugin

klihub: Fixed.

PR #665, opened 2026-04-20
[WIP] Implement exact reserved cpuset for some cases like using Intel priority core turbo

louie-tsai:
**balloons: add reservedCPUMode=hard-exact for exact reserved cpuset**
This patch adds a reservedCPUMode config knob and keeps the built-in reserved balloon pinned to the exact reserved cpuset instead of allowing resize/inflation outside it.
This PR also addresses new numa binding feature for NV GPUs
https://docs.vllm.ai/en/latest/configuration/optimization/#numa-binding-for-multi-socket-gpu-nodes

e.g. 
for below 2 NUMA nodes system
balloon1

We assigned reserved cpu list as below list with hard-exact.
balloon2

Therefore, we could let vllm-gpu pod using those reserved-cpu which could PCT cores or cpus binding with vLLM GPU workers.  In the meantime, vllm-cpu pods can use reminding default CPU lists in those 2 numa nodes with vLLM TP=2.
balloon3

### DIFF
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -1104,19 +1104,23 @@ func (p *balloons) newBalloon(blnDef *BalloonDef, confCpus bool) (*Balloon, erro
 		}
 	}
 	if p.reservedExact && blnDef.Name == reservedBalloonDefName {
-		exact := p.reserved.Clone()
-		if exact.Size() == 0 {
+                requested := p.reserved.Clone()
+                if requested.Size() == 0 {			
 			return nil, balloonsError("reservedCPUMode=hard-exact requires non-empty reserved cpuset")
 		}
-		if exact.Difference(p.freeCpus).Size() > 0 {
-			return nil, balloonsError("reserved exact cpuset %q is not fully free, missing %q", exact, exact.Difference(p.freeCpus))
+                if requested.Difference(p.freeCpus).Size() > 0 {
+                        return nil, balloonsError("reserved exact cpuset %q is not fully free, missing %q", requested, requested.Difference(p.freeCpus))			
 		}
-		newCpus, err := p.cpuAllocator.AllocateCpus(&exact, exact.Size(), blnDef.AllocatorPriority.Value().Option())
+                // AllocateCpus() may mutate the preferred/requested cpuset argument.
+                // Keep an immutable copy for post-allocation validation and logging.
+                exact := requested.Clone()
+                newCpus, err := p.cpuAllocator.AllocateCpus(&exact, requested.Size(), blnDef.AllocatorPriority.Value().Option())
 		if err != nil {
-			return nil, balloonsError("failed to allocate exact reserved cpuset %q: %w", exact, err)
+			return nil, balloonsError("failed to allocate exact reserved cpuset %q: %w", requested, err)
 		}
-		if !newCpus.Equals(exact) {
-			return nil, balloonsError("failed to allocate exact reserved cpuset: requested %q, got %q", exact, newCpus)
+                if !newCpus.Equals(requested) {
+                        return nil, balloonsError("failed to allocate exact reserved cpuset: requested %q, got %q", requested, newCpus)
+			
 		}
 		p.freeCpus = p.freeCpus.Difference(newCpus)
 		bln.Cpus = newCpus
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -76,9 +76,10 @@ type balloons struct {
 	options   *policy.BackendOptions // configuration common to all policies
 	bpoptions *BalloonsOptions       // balloons-specific configuration
 	cch       cache.Cache            // nri-resource-policy cache
-	allowed   cpuset.CPUSet          // bounding set of CPUs we're allowed to use
-	reserved  cpuset.CPUSet          // system-/kube-reserved CPUs
-	freeCpus  cpuset.CPUSet          // CPUs to be included in growing or new ballons
+	allowed       cpuset.CPUSet // bounding set of CPUs we're allowed to use
+	reserved      cpuset.CPUSet // system-/kube-reserved CPUs
+	reservedExact bool          // keep built-in reserved balloon on the exact reserved cpuset
+	freeCpus      cpuset.CPUSet // CPUs to be included in growing or new ballons
 	ifreeCpus cpuset.CPUSet          // initially free CPUs before assigning any containers
 	cpuTree   *cpuTreeNode           // system CPU topology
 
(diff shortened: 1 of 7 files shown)


deployment/helm/balloons/templates/daemonset.yaml
```diff
@@ -99,7 +100,9 @@ spec:
           image: {{ .Values.image.name }}:{{ .Values.image.tag | default .Chart.AppVersion }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           securityContext:
-            allowPrivilegeEscalation: false
+            privileged: true
+            allowPrivilegeEscalation: true
+            runAsUser: 0
```
klihub, line 105: @louie-tsai I don't see anything in this PR which would require such a change.

If we need to alter the capability set of the policy (for reasons beyond this PR), we should
- do it in a separate PR
- try to do it with additional capabilities instead of making the container privileged, if that is possible
- make it an opt-in using template conditionals and some suitable chosen Helm value

cmd/plugins/balloons/policy/flags.go
```diff
@@ -35,6 +36,9 @@ var (
 )
 
 const (
+	ReservedCPUModePreferred = cfgapi.ReservedCPUModePreferred
+	ReservedCPUModeHardExact = cfgapi.ReservedCPUModeHardExact
```
klihub, line 40: @louie-tsai Can't we achieve this without introducing explicit dedicated modes for the existing and the new behavior ? Maybe with something like "if we have the reserved CPUs defined as an explicit CPU set, and we have an explicit definition of the reserved balloon in the configuration, and that explicit definition has MaxBalloons == 1 && MinCpus == ReservedCpus.Size() && MaxCpus == MinCpus, then we treat that as an immutable/unresizable reserved balloon" ?

klihub: I mean, something that would make this behavior less of an exception, and instead more of just applying existing/expected behavior to the reserved balloon.

deployment/helm/balloons/values.yaml
```diff
@@ -3,9 +3,11 @@
 # Declare variables to be passed into your templates.
 ---
 image:
-  name: ghcr.io/containers/nri-plugins/nri-resource-policy-balloons
+  #name: ghcr.io/containers/nri-plugins/nri-resource-policy-balloons
   # tag, if defined will use the given image tag, otherwise Chart.AppVersion will be used
   #tag: unstable
+  name: image-registry.openshift-image-registry.svc:5000/nri-plugins/nri-resource-policy-balloons
```
klihub, line 9: @louie-tsai We publish our images in the `ghcr.io/containers/nri-plugins` registry, so the default should point there. You can alter/customize it using a Helm value.

deployment/helm/balloons/templates/daemonset.yaml
louie-tsai: yes. agree. that is for Openshift, so let me revert this change.

deployment/helm/balloons/values.yaml
louie-tsai: sorry. removed it accordingly.

deployment/helm/balloons/templates/daemonset.yaml
```diff
@@ -22,6 +22,7 @@ spec:
         {{ $name }}: "{{ $value }}"
       {{- end }}
     spec:
+      hostPID: true
```
fmuyassarov, line 25: Based on the all other changes, this seems to me as unjustified security escalation, basically same as @klihub had below.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1996,13 +2033,29 @@ func (p *balloons) resizeBalloon(bln *Balloon, newMilliCpus int) error {
...
 	cpuCountDelta := newCpuCount - oldCpuCount
+	if p.reservedExact && bln.Def.Name == reservedBalloonDefName {
+		if cpuCountDelta == 0 {
+			if !bln.Cpus.Equals(p.reserved) {
+				return balloonsError("reserved balloon exact cpuset drifted: expected %q, got %q", p.reserved, bln.Cpus)
+			}
+			return nil
+		}
+		return balloonsError("reserved balloon exact cpuset cannot be resized: expected %q, current %q", p.reserved, bln.Cpus)
+	}
```
fmuyassarov, line 2058: When the reserved balloon's CPU count matches `exactCount` but the cpuset has drifted, it falls
through and hits the generic `if oldCpuCount == newCpuCount { return nil }` before drift detection is ever reached. As such, currently the drift error in is dead code,  it can never be triggered. We could merge both blocks into one that always returns, so the generic early-return can't short-circuit it. Example:

```diff
diff --git a/cmd/plugins/balloons/policy/balloons-policy.go b/cmd/plugins/balloons/policy/balloons-policy.go
index 3211893f..39a73cfe 100644
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -2035,10 +2035,13 @@ func (p *balloons) resizeBalloon(bln *Balloon, newMilliCpus int) error {
        }
        if p.reservedExact && bln.Def.Name == reservedBalloonDefName {
                exactCount := p.reserved.Size()
-               if oldCpuCount == exactCount && bln.Cpus.Equals(p.reserved) {
-                       return nil
+               if oldCpuCount == exactCount {
+                       if bln.Cpus.Equals(p.reserved) {
+                               return nil
+                       }
+                       return balloonsError("reserved balloon exact cpuset drifted: expected %q, got %q", p.reserved, bln.Cpus)
                }
-               newCpuCount = exactCount
+               return balloonsError("reserved balloon exact cpuset cannot be resized: expected %q, current %q", p.reserved, bln.Cpus)
        }
        log.Debugf("resize %s to fit %d mCPU", bln, newMilliCpus)
        log.Debugf("- change size from %d to %d full cpus", oldCpuCount, newCpuCount)
@@ -2047,15 +2050,6 @@ func (p *balloons) resizeBalloon(bln *Balloon, newMilliCpus int) error {
                return nil
        }
        cpuCountDelta := newCpuCount - oldCpuCount
-       if p.reservedExact && bln.Def.Name == reservedBalloonDefName {
-               if cpuCountDelta == 0 {
-                       if !bln.Cpus.Equals(p.reserved) {
-                               return balloonsError("reserved balloon exact cpuset drifted: expected %q, got %q", p.reserved, bln.Cpus)
-                       }
-                       return nil
-               }
-               return balloonsError("reserved balloon exact cpuset cannot be resized: expected %q, current %q", p.reserved, bln.Cpus)
-       }
        p.forgetCpuClass(bln)
        defer func() {
                if err := p.useCpuClass(bln); err != nil {

```

fmuyassarov (commented):
Left some comments inline but also it would be nice if @louie-tsai you could run`gofmt -w cmd/plugins/balloons/policy/balloons-policy.go` to fix the indentation issues. Thanks.

klihub:
> Therefore, we could let vllm-gpu pod using those reserved-cpu which could PCT cores or cpus binding with vLLM GPU workers. In the meantime, vllm-cpu pods can use reminding default CPU lists in those 2 numa nodes with vLLM TP=2.

Hmm..., I think we shouldn't run a vLLM workload pod in the reserved pool which, at least normally, is reserved for the kube-system K8s namespace, so IOW for K8s control plane pods.

@louie-tsai Why do you try to use reserved balloons for vLLM ? If the only/primary reason is that the reserved balloon allows you to hand-pick a CPU set, and you are trying to (mis)use this to effectively assign PCT-enabled cores to vLLM workloads (by first hand-assigning PCT cores to the reserved pool, and then assigning vLLM pods to the reserved pool), then this is not the right way to achieve that goal. We are working on adding PCT-support to both the balloons and topology-aware policies. Once we have that support, you should be able to use that to request PCT-enabled cores for those workloads which benefit from it, in this case for vLLM pods.

klihub:
Closing in favor of #667.

PR #667, opened 2026-05-13
balloons: add cpuClasses with cpufreq and PCT turbo allocation

askervin:
This PR introduces `pkg/resmgr/cpuclass` as a replacement for `pkg/resmgr/control/cpu`. The latter is not removed in this patch series, but it is disconnected from the balloons policy and thus left unused.

Unlike the CPU controller, that is decoupled from resource policies, `cpuclass` is designed for policy-controlled CPU tuning. This is needed with more complex CPU tuning scenarios, for instance:
- configure frequencies of CPUs that are unassigned or shared by multiple containers
- provide hints to policy's CPU allocation about CPUs where wanted tunings are possible.

This PR introduces new configuration section `cpuClasses` for specifying CPU tunings. This is aligned with `schedulingClasses` and `loadClasses` already available on the top level in balloons policy configuration. It uses the same class notation as other classes and balloonTypes, that is, a list of objects with "name" attribute specifying a class name, replacing `control.cpu.classes` object where key was the class name.

This PR enables using `cpuclass` from the balloons policy. However, it is expected to be usable from other policies, too. Of course the topology-aware policy in particular. Therefore `cpuclass` API associates only cpusets with CPU class names, and it knows nothing about "balloons", "shared CPUs" or "idle CPUs", that would have no direct equivalents in the topology-aware policy.

CPU frequency configuration is extended with support for units (3100MHz or 3.1GHz), and symbolic frequencies: "min" (minimum frequency), "base" (base frequency) and "turbo" (max turbo frequency).

The cpufreq allocator (internal to `cpuclass`) works by changing symbolic frequency interpretation. CPUs belonging to highest `turboPriority` CPU classes are eligible to the "turbo" frequency. But if "turbo" is specified as minFreq or maxFreq for CPUs that belong to CPU classes with a lower `turboPriority`, their "turbo" is interpreted as "base". However, when all containers that belonged to the highest `turboPriority` classes are removed, CPU classes with the next highest `turboPriority` become highest ones, and they get their "turbo" interpreted as "turbo" again. In other words, at any point of time, highest priority CPU classes get the turbo while other get at most "base".

The pct allocator (internal to `cpuclass`) works in two modes. The assoc-only mode associates CPUs to user-specified CLOSes. In this mode, user specifies `pctClosID` in cpuClasses to associate CPUs in that class in correct predefined CLOS. The managed mode replaces system SST configuration with its own, removing need for using intel-speed-select or BIOS PCT configuration. In this mode, user specifies `pctPriority: high` in cpuClasses to tune CPUs into a CLOS with maximum performance.


askervin:
@kad, @marquiz, do you think we could approach turbo budget sharing with this kind of architecture in the balloons policy?

I'm adding `cpuClasses` under `resmgr` similarly to `schedulingClasses` to pave the way taking them into the topology-aware policy's guaranteed containers later on, too.

askervin:
There is some technical and architectural debt that I wish to pay still in this PR. That is, the CPU controller should not directly modify frequencies, but this should be via cache and aligned with the spirit of applying "pending updates".

Unfortunately controller hooks are container-specific and possibly called multiple times while handling single NRI event, whereas all CPU properties should be written once per NRI event. I'll add yet another hook to the Controller interface to commit whatever changes a controller has stored since the previous Commit().

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1399,14 +1382,28 @@ func changesBalloons(opts0, opts1 *BalloonsOptions) bool {
...
 		o0.BalloonDefs[i].CpuClass = ""
 		o1.BalloonDefs[i].CpuClass = ""
 	}
+	// 2. Schedulingpolicy parameters
+	o0.SchedulingClasses = nil
+	o1.SchedulingClasses = nil
+	for i := range o0.BalloonDefs {
+		o0.BalloonDefs[i].SchedulingClass = ""
+		o1.BalloonDefs[i].SchedulingClass = ""
+	}
```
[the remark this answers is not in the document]
askervin: schedulingclasses cannot be changed in running containers, and therefore a configuration change in schedulingclasses only should not cause re-assigning cpusets of all running containers, that is very disruptive and still leaves them running with old schedulingclass configuration.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -124,35 +141,176 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
...
+				}
+			}
+			// Update the cache even on failure: the desired value
+			// is unchanged so retrying on every Assign would just
+			// spam logs without ever succeeding. A subsequent
+			// configure() resets lastFreq so a real configuration
+			// change still triggers a fresh attempt.
+			state.min = min
+			state.hasMin = true
 		}
```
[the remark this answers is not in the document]
askervin: we expect all similar sysfs writes to fail, and we do not want to flood the logs with error messages by errors from hopeless write retrials of already-tried data on every commit.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -124,35 +141,176 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
...
+			log.Debugf("enforcing cpu frequency max %d from class %q on cpu %d", max, class, cpu)
+			if err := utils.SetCPUScalingMaxFreq(utils.ID(cpu), int(max)); err != nil {
+				log.Errorf("cannot set max freq %d on cpu %d: %v", max, cpu, err)
+				if firstErr == nil {
+					firstErr = err
+				}
+			}
+			state.max = max
+			state.hasMax = true
 		}
```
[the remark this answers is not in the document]
askervin: if this failed, retries will fail, too. therefore we update the cache anyway.

pkg/resmgr/control/cpu/cpu.go
```diff
@@ -124,35 +141,176 @@ func (ctl *cpuctl) PostStopHook(c cache.Container) error {
...
+			log.Debugf("enforcing cpu frequency governor %q from class %q on cpu %d", governor, class, cpu)
+			if err := utils.SetCPUScalingGovernor(utils.ID(cpu), governor); err != nil {
+				log.Errorf("cannot set cpufreq governor %q on cpu %d: %v", governor, cpu, err)
+				if firstErr == nil {
+					firstErr = err
+				}
+			}
+			state.governor = governor
+			state.hasGov = true
 		}
```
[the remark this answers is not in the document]
askervin: if this failed, retrying it would fail, too. therefore, just update the cache.

cmd/plugins/balloons/policy/cpuclass.go
```diff
@@ -0,0 +1,419 @@
...
+		return nil, fmt.Errorf("no CPUs found in system topology")
+	}
+	for _, id := range cpuIDs {
+		cpu := sys.CPU(id)
+		if cpu == nil || !cpu.Online() {
+			continue
+		}
+		freq := cpu.FrequencyRange()
+		baseFreq := cpu.BaseFrequency()
+		if baseFreq == 0 || freq.Max == 0 {
```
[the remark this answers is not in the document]
askervin: using max as basefreq if basefreq is not available.

cmd/plugins/balloons/policy/cpuclass.go
```diff
@@ -0,0 +1,644 @@
...
+
+// WithCPUClasses provides the user-facing CPUClass definitions.
+func WithCPUClasses(classes []*CPUClass) TurboOption {
+	return func(a *CPUClassTurboAllocator) error {
+		a.classes = classes
+		a.classByName = make(map[string]*CPUClass, len(classes))
+		for _, cc := range classes {
+			a.classByName[cc.Name] = cc
+		}
+		return nil
```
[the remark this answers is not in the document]
askervin: null items do not pass balloonspolicy validation

cmd/plugins/balloons/policy/cpuclass.go
```diff
@@ -0,0 +1,644 @@
...
+// Reconfigure replaces the CPU class set, idle class name and turbo
+// domain mode. Resets per-domain turbo winners so the next
+// UseClass/ForgetClass call recomputes the effective frequencies, and
+// re-pushes class definitions to the CPU controller.
+func (a *CPUClassTurboAllocator) Reconfigure(classes []*CPUClass, idleClass, turboDomain string) error {
+	a.classes = classes
+	a.classByName = make(map[string]*CPUClass, len(classes))
+	for _, cc := range classes {
+		a.classByName[cc.Name] = cc
+	}
```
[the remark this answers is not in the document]
askervin: empty names are caught in validateConfig. nil is not considered an issue and would be checked in validateConfig if somewhere. now nil checks are not done to other similar config structures either.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1600,6 +1615,32 @@ func (p *balloons) validateConfig(bpoptions *BalloonsOptions) error {
 	if len(undefinedSchedulingClasses) > 0 {
 		return balloonsError("schedulingClass(es) defined in balloonTypes but missing from schedulingClasses: %v", undefinedSchedulingClasses)
 	}
+	// Validate CPUClasses.
+	cpuClassNames := map[string]struct{}{}
+	for _, cc := range bpoptions.CPUClasses {
+		if cc.Name == "" {
+			return balloonsError("missing or empty name in a cpuClasses entry")
+		}
```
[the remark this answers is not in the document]
askervin: nil/null is not checked in other similar configs either in validateConfig, and it is prevented by kubernetes balloonspolicy validation. not an issue here.

pkg/apis/config/v1alpha1/balloons-policy.go
```diff
@@ -32,13 +37,45 @@ func (c *BalloonsPolicy) CommonConfig() *CommonConfig {
...
+			ctrl.CPU.Classes[cc.Name] = cpucfg.Class{
+				MinFreq:                     freqKHzOrZero(cc.MinFreq),
+				MaxFreq:                     freqKHzOrZero(cc.MaxFreq),
+				EnergyPerformancePreference: cc.EnergyPerformancePreference,
+				UncoreMinFreq:               freqKHzOrZero(cc.UncoreMinFreq),
+				UncoreMaxFreq:               freqKHzOrZero(cc.UncoreMaxFreq),
+				FreqGovernor:                cc.FreqGovernor,
+				DisabledCstates:             cc.DisabledCstates,
+			}
+		}
```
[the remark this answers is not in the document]
askervin: same as above.

askervin:
Back to draft.

Adding PCT core adjustments cleanly changes cpuClasses implementation drasticly.
- Having cpufreq  and goresctrl/sst backends requires cleaner separation of concerns compared to what is in this patch.
- Using existing cpu controller for backwards compatibility and as an internal backend to cpuClasses is questionable. We do not need to tweak controller interface if cpuClasses uses its own backends to cpufreq/cpuidle/sst.
- Current cpuClasses interface is not policy-agnostic. For instance, the concept of idle CPUs is balloons specific, and it can be avoided.
- Current cpuClasses interface does not provide hints on to-be-allocated CPUs where intended CPU tuning would be possible to do. For instance, a punit inside a package might have run out of PCT cores, and therefore highest turbo frequencies would be only available on cores of another punit in the same package. Hint interface should be such that it could be used as is in the topology-aware policy, too.

askervin:
Fixing lint issues requires updating sysfs and cpuallocator to use new SST API from goresctrl. That turned out to be non-trivial in allocator's BF-based CPU prioritization. Perhaps it'll be best to take goresctrl v0.13.0 update as a separate PR with new API use in it (to keep the lint happy)...

pkg/resmgr/cpuclass/internal/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,338 @@
...
+			continue
+		}
+		var d domainID
+		switch mode {
+		case turboDomainSystem:
+			d = systemDomainID
+		default:
+			d = domainID(c.PackageID())
+		}
+		a.cpuDomain[int(cpuID)] = d
```
klihub, line 225: Does not `SystemDomainID` being `0` result in all CPUs of package `0` having `d` set to `0` here both for `turboDomainSystem` and `turboDomainPackage` modes ? And if it does, isn't it a problem ?

pkg/resmgr/cpuclass/internal/cpufreq/sysfs.go
```diff
@@ -0,0 +1,153 @@
...
+type Writer struct {
+	hooks       Hooks
+	lastWritten map[int]cpufreqWritten
+}
+
+// NewWriter returns a Writer wired to the given hooks. Pass a
+// zero-valued Hooks to use real sysfs via goresctrl.
+func NewWriter(hooks Hooks) *Writer {
+	return &Writer{
+		hooks:       hooks,
```
klihub, line 57: This is not something to be addressed, it is merely a subjective notice. AFAICT, with things being hooked up as they are, if you did something like this here before returning the writer

```golang
...
    if w.hooks.SetMin == nil {
		w.hooks.SetMin = func(cpu, freq int) error { 
	        return utils.SetCPUScalingMinFreq(utils.ID(cpu), freq)
        }
    }
    if w.hooks.SetMax == nil {
		w.hooks.SetMax = func(cpu, freq int) error { 
	        return utils.SetCPUScalingMaxFreq(utils.ID(cpu), freq)
        }
    }
    if w.hooks.SetGov == nil {
		w.hooks.SetGov = func(cpu int, governor string) error { 
            return utils.SetCPUScalingGovernor(utils.ID(cpu), governor)
        }
    }
...
```

you could eliminate the indirect calls via `Writer.callSet{Min,Max,Gov}` and always simply call `w.hooks.Set{Min,Max,Gov}` directly instead.

pkg/agent/node-extended-resources.go
```diff
@@ -0,0 +1,309 @@
...
+// resources left over by a prior plugin process (helm reinstall,
+// pod crash, switch to a different policy, etc.) get pruned by
+// the regular diff logic on the next publish.
+var extendedResourcesSynced bool
+
+// extendedResourceDomain is the per-domain prefix the agent owns.
+// Only resources whose name starts with this prefix are touched
+// by the agent (other extended resources advertised by other
+// controllers are left alone).
+const extendedResourceDomain = "cpuclass.balloons.nri.io/"
```
klihub, line 53: Not necessarily for this PR, but I think eventually we should
- avoid hardcoding this,
- allow more than a single value, and
- make extended resource names of prefixes registerable from the outside

pkg/resmgr/cpuclass/internal/cpufreq/cpufreq.go
askervin: The only problem coming into my mind is that of debugging: logs show domain being 0 on every package. Choosing systemDomainID=-1 could be easier to distinguish in the logs: this definitely does not refer to any package.

That said, I'm not sure if supporting "system" as a turbo domain makes much sense. It's here mostly as a place holder and for testability. Let me explain... Sub-package turbo domains will make more sense, because on some systems CPU sets of "SST power units" happen to match "numa" zones. But unlike SST power units, numa zones we can read without SST support, permissions and CLOS configuration complexity/side effects. Towards this end, I wanted to make sure that current design will be able to support multiple turbo domains instead of hardcoding it to be always "package". But on the other hand, I did not want to implement full support yet due to already huge complexity of this PR. And as I scoped it out but, nevertheless, wanted to have something extremely simple to test, I ended up with simple split: "package" (good default) and "system" (questionable). "numa" will never be a good default: it can be too small or too large, depending on BIOS configuration.

systemDomainID matching package IDs cannot cause functional issues, as far as I can see. The allocator understands only single "turbo domain" split for the whole system, and therefore IDs of domains in different splits do not collide.

pkg/resmgr/cpuclass/internal/cpuidle/cpuidle.go
```diff
@@ -0,0 +1,122 @@
...
+	enCpuCstates.SetAttrs(cstates.AttrDisable, "0")
+	disCpuCstates.SetAttrs(cstates.AttrDisable, "1")
+	log.Debugf("cstates: class %q on cpus %v: enable=%v disable=%v", class, cpus, enabledCstates, disabledCstates)
+	if err := enCpuCstates.Apply(); err != nil {
+		return fmt.Errorf("cannot enable cstates %v on cpus %v: %w", enabledCstates, cpus, err)
+	}
+	if err := disCpuCstates.Apply(); err != nil {
+		return fmt.Errorf("cannot disable cstates %v on cpus %v: %w", disabledCstates, cpus, err)
+	}
+	return nil
```
klihub, line 98: Similar subjective notice here as earlier. AFAICT, if you split the default handling out to something like `goresctrlApply()`, initialize unset Hooks.Apply in `NewWriter()` to `goresctrlApply` (and move the early return check to `goresctrlApply()`) then you could simply do a `return w.hooks.Apply(cpus, nil, disabledCstates)` here. Maybe not as clear a win as in the earlier case.

pkg/resmgr/cpuclass/internal/cpuidle/cpuidle.go
```diff
@@ -0,0 +1,122 @@
...
+func (w *Writer) Enforce(class string, disabledCstates []string, cpus []int) error {
+	if len(cpus) == 0 {
+		return nil
+	}
+	if len(disabledCstates) == 0 && w.cs == nil && w.hooks.Apply == nil && cstatesEnvOverridesJson == "" {
+		return nil
+	}
+	if w.hooks.Apply != nil {
+		return w.hooks.Apply(cpus, nil, disabledCstates)
+	}
```
klihub, line 69: nit: if this is not reorganized as suggested in the next comment, it would look clearer to me to first try `w.hooks.Apply()`, and only then the default handler, now with the `w.hooks.Apply` nil-check removed.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1671,13 +1859,37 @@ func (p *balloons) setConfig(bpoptions *BalloonsOptions) error {
...
+	p.bpoptions = bpoptions
+
+	// Construct the CPU class handler that fronts both cpufreq and
+	// PCT internals.
+	if p.cpuClasses == nil {
+		h, err := cpuclass.New(p.options.System)
+		if err != nil {
+			return balloonsError("failed to create CPU class handler: %w", err)
+		}
+		p.cpuClasses = h
```
kad, line 1873: is this handler shutdown somewhere?

pkg/agent/node-extended-resources.go
```diff
@@ -0,0 +1,309 @@
...
+	// remove keys we owned before but are now gone.
+	type jsonPatchOp struct {
+		Op    string      `json:"op"`
+		Path  string      `json:"path"`
+		Value interface{} `json:"value,omitempty"`
+	}
+
+	ops := []jsonPatchOp{}
+	for name, qty := range resources {
+		if !strings.HasPrefix(name, extendedResourceDomain) {
```
kad, line 107: for this one, would it make sense to make more policy neutral/reusable across policies?

klihub: @kad I had the exact same comment (I'd like to have support for more than one resource name and/or prefixes, and I'd also like to make them registrable from outside the agent outside by callers), but we agreed with @askervin that we'll generalize extended resource support with subsequent PRs.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Good and difficult question...

It might make sense to call cpuclass Shutdown() when stopping a policy. Currently resource manager Run() only calls agent.Stop() on SIGINT/SIGTERM but the policy interface has no Stop(). That is, there is no really a clean place to call Shutdown().

I'd leave adding the Stop() into the framework and restoring/resetting system state to follow-up PRs. Currently balloons documentation gives an [example config](https://containers.github.io/nri-plugins/stable/docs/resource-policy/policy/balloons.html#reset-cpu-and-memory-pinning) to reset CPU and memory affinities to "every container is allowed to use all CPUs and memories". But at the moment I'm not even sure if policy.Stop() -- that might be only done only in order to restart the policy container -- should do anything this dramatic things fully automatically. And if CPU affinities are not reset, then resetting CPU attributes is questionable, too.

kad (approved):
lgtm

klihub (approved):

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -91,6 +91,7 @@ type balloons struct {
 
 	cpuAllocator cpuallocator.CPUAllocator    // CPU allocator used by the policy
 	memAllocator *libmem.Allocator            // memory allocator used by the policy
+	cpuClasses   *cpuclass.Handler            // CPU class handler (cpufreq + PCT internals)
```
marquiz: nit: would "SST internals" or "SST/PCT internals" be more correct (and future proof)?

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -788,83 +827,104 @@ func largest(sliceLen int, valueOf func(i int) int) ([]int, int) {
...
+func (p *balloons) resolveCpuClassName(name string) string {
+	if name != "" {
+		return name
+	}
+	for _, cc := range p.bpoptions.CPUClasses {
+		if cc.Name == defaultCpuClassName {
+			return defaultCpuClassName
+		}
+	}
+	return name
```
marquiz: thought:just caught my eye (not sure if this makes it any more readable, though)

```suggestion
if slices.ContainsFunc(p.bpoptions.CPUClasses, func(cc CPUClass) bool {
    return cc.Name == defaultCpuClassName
}) {
    return defaultCpuClassName
}
return ""
```

There are other places where  the `slices.ContainsFunc` would apply, too

pkg/apis/config/v1alpha1/resmgr/policy/cpuclass.go
```diff
@@ -0,0 +1,98 @@
...
+	PctPriority string `json:"pctPriority,omitempty"`
+	// PctClosID pins this class to a specific SST-CP CLOS ID
+	// (0..ClosCount-1, typically 0..3) and signals "assoc-only"
+	// mode: nri-plugin will only associate this class's CPUs to
+	// the given CLOS, without touching the SoC-wide SST state
+	// (no CPReset, no TFEnable, no CLOS reconfiguration). Use
+	// this when an operator or the BIOS has pre-configured the
+	// CLOSes. Mutually exclusive with PctPriority.
+	// +kubebuilder:validation:Minimum=0
+	PctClosID *int `json:"pctClosID,omitempty"`
```
marquiz: Sorry for coming up for the comment this late, but I started to think the correctness and "future-proofness" of the `Pct` prefix/qualifier here. The parameters are more generic, targeting SST(-CP) so should we just use more generic sstClosID (and ditto others, although for the PctPriority field I'm not sure how it should be named)? AFAIU, the PCT is just a name for the bios feature, when we configure the stuff ourselves it's kinda generic SST-TF (overriding the possible PCT pre-configured state)...

My major concern is that when we merge (and release) this it becomes part of the user-facing API and hard and inconvenient to change later. We might add later more generic SST-CP support – not enabling/requiring SST-TF, and, exposing also the now hidden config parameters. In this case the Pct* parameters are very misleading.

Thoughts?

docs/resource-policy/policy/balloons.md
```diff
@@ -100,10 +100,10 @@ helm repo add nri-plugins https://containers.github.io/nri-plugins
 helm repo update
 ```
 
-Install the balloons policy with default configuration:
+Install the balloons policy with default configuration [with PCT enabled]:
 
 ```sh
-helm install nri-resource-policy-balloons nri-plugins/nri-resource-policy-balloons --namespace kube-system
+helm install nri-resource-policy-balloons nri-plugins/nri-resource-policy-balloons --namespace kube-system [--set pct=true]
```
marquiz: The `--set pct=true` here does not actually enable anything. Other than deploy container in privileged mode and mount hostdev(?) If so, I think this is somewhat misleading and could be more generic, like `hostdev=true` or `privileged=true`. Thoughts?

pkg/agent/node-extended-resources.go
```diff
@@ -0,0 +1,309 @@
...
+		return
+	}
+
+	extendedResourcesLock.Lock()
+	defer extendedResourcesLock.Unlock()
+
+	toRemove := map[string]struct{}{}
+	for k := range lastPublishedExtendedResources {
+		toRemove[k] = struct{}{}
+	}
```
marquiz, line 237: observation: we refuse to publish anything that outside "our domain" so this looks like unnecessary step (they'll get removed by the code below). But maybe a useful safety measure for future(?)

pkg/agent/node-extended-resources.go
```diff
@@ -0,0 +1,309 @@
...
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	// stable order without pulling in sort
+	for i := 1; i < len(keys); i++ {
+		for j := i; j > 0 && keys[j-1] > keys[j]; j-- {
+			keys[j-1], keys[j] = keys[j], keys[j-1]
+		}
+	}
```
marquiz: What's wrong with sort (or `slices.Sort()`)??

pkg/sysfs/system.go
```diff
@@ -340,6 +342,16 @@ type cacheOverride struct {
 
 var cacheEnvOverrides map[int][]*Cache
 
+// cpufreqOverride specifies frequency values to use instead of reading sysfs.
```
marquiz, line 345: note: For the override stuff a note that these are targeted for testing (only) would help the reader

pkg/resmgr/policy/policy.go
```diff
@@ -143,6 +151,9 @@ type Policy interface {
 	ExportResourceData(cache.Container)
 	// GetTopologyZones returns the policy/pool data for 'topology zone' CRDs.
 	GetTopologyZones() []*TopologyZone
+	// GetExtendedResources returns node-level extended resources
+	// the active policy wishes to publish on the local Node.
+	GetExtendedResources() map[string]int64
```
marquiz: We could also use `resource.Quantity` here(?) Would align with what we have in `ZoneResource` for example

marquiz (commented):
Some notes below, what I spotted with a somewhat cursory review of the latest iteration of this Giga-PR 💦 For example the new allocator stuff I needed to skip completely with the contstrained time I had 😬 

I think we could merge this although the "pct" naming of the config api is not ideal.

pkg/apis/config/v1alpha1/resmgr/policy/cpuclass.go
klihub: That would sound like a reasonable change.

pkg/agent/node-extended-resources.go
klihub: I second this. We already pull in sort to many other packages, so avoiding it here is hardly a win, considering it is harder to read.

docs/resource-policy/policy/balloons.md
klihub: If we want to keep this target-feature-oriented instead of achieving-means-oriented (which I think is a good UX idea and IIRC we do it for other features as well), maybe we could simply have this as `--set allowPCT=true`. As it allows PCT support but does not in itself enable it yet...

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Yes. Let's use SST/PCT because we ended up with "pct" in config API...

pkg/apis/config/v1alpha1/resmgr/policy/cpuclass.go
askervin: When trying to read this through the eyes of a user who reads [PCT documentation](https://cdrdv2-public.intel.com/846906/846906_PCT_Technology_Technical_Article_Rev_1_5_1.pdf), it says:

> When PCT Technology is enabled, the BIOS programs the Intel SST-TF CLOS
registers for priority cores and non-priority cores... 

...that is, sstClosID should be good and even better (for the future) than pctClosID.

I'll do this change.

docs/resource-policy/policy/balloons.md
askervin: Switched to --allowPCT=true.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Explicit `return ""` is definitely more readable than returning name while it's known and expected to be empty. I did not switch to ContainsFunc() for now. Perhaps those changes can be done in a bunch later.

pkg/agent/node-extended-resources.go
askervin: uh-oh, sorry, switched to iterating `slices.Sorted(maps.Keys(m))`.

pkg/sysfs/system.go
askervin: Added a note on this.

pkg/resmgr/policy/policy.go
askervin: Changed. Added new commit for this to keep the change separate (just in case I failed to do it properly and we need to revert).

PR #671, opened 2026-06-07
resctrl-mon: revert CRI-O pidfile fallback due to security concerns

cmcantalupo:
### Summary

Revert the CRI-O pidfile fallback added in f7883b9e, and add a runtime
version check that rejects CRI-O < 1.36 at startup.

### Context

CRI-O <= 1.35 does not populate `Container.Pid` in NRI events. The
reverted commit attempted to work around this by reading PID from
CRI-O's internal pidfile. However, enabling this requires a host mount
that violates the principle of least privilege.

The upstream fix (CRI-O commit 19d319695) populates the PID field
directly in NRI and is available in CRI-O >= 1.36.

### Impact

| Runtime | Effect |
|---------|--------|
| containerd | No impact — PID is always provided via NRI |
| CRI-O >= 1.36 | No impact — PID is provided via NRI |
| CRI-O < 1.36 | **Plugin refuses to start** with a clear error message; DaemonSet enters CrashLoopBackOff |

The hard fail is intentional: without PIDs, mon_groups are created but
remain empty, silently reporting zeros for all metrics. This is worse
than a crash because it produces data indistinguishable from "this pod
uses no LLC/bandwidth," leading to incorrect capacity decisions. A loud
crash with a clear log message ("CRI-O 1.35.x does not provide
container PIDs via NRI, requires >= 1.36") gives the admin immediate
actionable feedback.

### Security analysis

The current deployment in `origin/main` is **not actively insecure**.
The default Helm daemonset template mounts only:

- `/var/run/nri` (NRI socket)
- `/sys/fs/resctrl` (resctrl filesystem)

There is no volume or volumeMount for `/run/containers/storage` in the
shipped template. Without that mount, `readContainerPID()` calls
`os.ReadFile` on a path that does not exist inside the container, gets
`ENOENT`, and returns 0. The fallback is a complete no-op in the
default deployment — no information is leaked and no error is raised.

However, the code **cannot function** without the insecure mount, and
the only way to make it function is to add a hostPath volume for
`/run/containers/storage`. That directory contains:

- `/run/containers/storage/overlay-containers/<id>/userdata/pidfile`
  (the target file)
- `/run/containers/storage/overlay-containers/<id>/userdata/config.json`
  (full OCI runtime spec including process environment variables)

Mounting this path grants the monitoring plugin read access to secrets
and environment variables of **every container on the node**. This is
an unacceptable privilege escalation for a read-only monitoring plugin.

**In summary:** the upstream code is dead code that can only be
activated via an insecure deployment modification. Rather than ship
non-functional code whose only activation path is insecure, we remove
it entirely.

### DIFF
--- a/cmd/plugins/resctrl-mon/plugin.go
+++ b/cmd/plugins/resctrl-mon/plugin.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -72,6 +74,9 @@ func newPlugin() *plugin {
 // Configure handles connecting to container runtime's NRI server.
 func (p *plugin) Configure(ctx context.Context, config, runtime, version string) (stub.EventMask, error) {
 	log.Infof("Connected to %s %s...", runtime, version)
+	if err := checkRuntimeVersion(runtime, version); err != nil {
+		return 0, err
+	}
 	if config != "" {
 		log.Debugf("loading configuration from NRI server")
 		if err := p.setConfig([]byte(config)); err != nil {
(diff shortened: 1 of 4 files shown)


kad (approved):
lgtm

klihub (approved):
LGTM

PR #675, opened 2026-06-09
mount NRI socket directory as read-only

fmuyassarov:
Plugins don't write to `/var/run/nri` directory but only dial to the socket that exists there.


fmuyassarov:
and once we know that k8s 1.30 is not supported anymore, we should add `recursiveReadOnly: Enabled` too, 
xref: https://kubernetes.io/blog/2024/04/23/recursive-read-only-mounts/#new-mount-option-recursivereadonly

klihub (approved):

PR #677, opened 2026-06-11
Move tar.go to testutils

uniemimu:
There are various CodeQL findings from inside tar.go related to extracting potentially malicious tar files. However tar.go was only ever used inside test code. Hardening test code is futile as any malicious user can modify the code base of their fork in any way they want and thus attack the CI system.

Moving tar.go to testutils and putting in a comment stating that the code must not be used in production makes it clear that this code is only to be used for testing. The CodeQL findings can then be marked as "wontfix".

### DIFF
--- a/cmd/plugins/topology-aware/policy/libmem_test.go
+++ b/cmd/plugins/topology-aware/policy/libmem_test.go
@@ -23,7 +23,7 @@ import (
 	cfgapi "github.com/containers/nri-plugins/pkg/apis/config/v1alpha1/resmgr/policy/topologyaware"
 	policyapi "github.com/containers/nri-plugins/pkg/resmgr/policy"
 	system "github.com/containers/nri-plugins/pkg/sysfs"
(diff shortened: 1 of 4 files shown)


kad (approved):
LGTM

fmuyassarov:
Hi @uniemimu. Could you check the failure on the licensing workflow?

uniemimu:
> Hi @uniemimu. Could you check the failure on the licensing workflow?

It looks like a timeout. The verify-action is given 5 minutes only, and running go-licenses is slow. I tested in my own laptop with the PR, and running the make target takes over 4 minutes. Adding the installation of go-licenses and other parts of the verify-action, the 5-minute threshold is precariously low even during normal runs.

I would suggest retrying the running of the check a couple of times. If it sporadically fails to timeouts, I would consider either optimizing the running or increasing the timeout limit from 5 minutes. Please see #678

I don't have the possibility to retrigger the check run myself.

PR #678, opened 2026-06-12
Parallelize makefile running

uniemimu:
This greatly increases especially the speed of license creation. Also benefits the build of the plugins.


kad (approved):
Thanks for improvement.

PR #676, opened 2026-06-10
Fixes for various CodeQL findings

uniemimu:
CodeQL findings. Mostly fixes for incorrect conversion between integer types, but includes a regexp cleanup. From copilot autofix.

PR #682, opened 2026-06-12
pkg/sysfs: replace interface{}-based sysfs I/O with typed generics

kad:
CodeQL flagged integer conversion bugs here (PR #676 was the autofix attempt). The real problem: readSysfsEntry/writeSysfsEntry accepted interface{} and every integer parser called strconv.ParseInt with bitSize=0, meaning 64-bit, then cast the result down -- so parsing "300" into int8 silently gave you 44. Patching the casts doesn't help; the issue is the parse site itself.

Replaced the whole interface{}-based layer with typed generic functions: readSysfsRaw, readSysfsString, readSysfsInt[T], readSysfsUint[T], readSysfsIDSet, readSysfsIntList[T], readSysfsEPP on the read side, and writeSysfsRaw, writeSysfsInt[T], writeSysfsUint[T] for writes.

parseIntTo[T] and parseUintTo[T] use a type switch to get T's actual bit width and pass it straight to strconv.ParseInt/ParseUint. Out-of-range values now error. The unsafe import is gone.

A few other things were broken in the old code:
- parseValueList `break`-ed on empty tokens, stopping the whole parse instead of just skipping the empty slot
- parseIDSet split on "-" without a limit, so "0-7" got three pieces when the separator also happened to be "-"; fixed with SplitN(..., 2)
- formatValueList returned "" instead of the string it built -- removed along with formatValue and other dead helpers

All 27 call sites in system.go updated. Added utils_test.go covering all new utility functions.

Replaces: #676

### DIFF
--- a/pkg/sysfs/system.go
+++ b/pkg/sysfs/system.go
@@ -531,7 +531,7 @@ func (sys *system) SetCpusOnline(online bool, cpus idset.IDSet) (idset.IDSet, er
 			continue
 		}
 
-		if _, err := writeSysfsEntry(entry, "online", desired, &current); err != nil {
+		if err := writeSysfsInt(entry, "online", desired, &current); err != nil {
 			return nil, sysfsError(entry, "failed to set online to %d: %v", desired, err)
 		}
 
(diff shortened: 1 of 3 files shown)


klihub:
@kad Should we use the copilot suggested conversion overflow check for named types and remove the unused parseUintSlice[T] ? Otherwise looks good to me.

kad:
> @kad Should we use the copilot suggested conversion overflow check for named types and remove the unused parseUintSlice[T] ? Otherwise looks good to me.

I've got better idea, to use reflect to get actual size of any type, so it should be simplier now. Please check newer revision.

askervin (approved):
LGTM.

Haven't seen this causing any issues in e2e tests either, only unrelated instability. I think this is ok to merge.

PR #684, opened 2026-06-15
e2e: install cri-tools/crictl on Fedora 43 and later.

klihub:
The ansible playbooks on Fedora assume crictl to be available without installing cri-tools. Fix this.


askervin (approved):
LGTM, Thanks! Got bitten by this, too, but hadn't figured out where to fix it yet...

kad (approved):

PR #687, opened 2026-06-17
e2e: fix CIDRs added to vm's no_proxy to be effective

askervin:
Using Dotenv.load() does not override environment's no_proxy if it is specified, and therefore CIDRs that are added to no_proxy in <e2e-vm-name>/env, "192.168.76.0/24" in particular, are not effective. This leads to kubelet connection to in-vm API server escape from vm to proxy servers, and never succeed. Using overload() forces env file definition override environment variables.


klihub (approved):

fmuyassarov (approved):
LGTM

PR #688, opened 2026-06-17
Otel library updates

uniemimu:
This fixes the vulnerability findings reported by Trivy.


pkg/instrumentation/resource.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"go.opentelemetry.io/otel/attribute"
 	otelresource "go.opentelemetry.io/otel/sdk/resource"
-	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.41.0"
```
klihub, line 24: I think if we make this change, then we also need to bump semconv to the same v1.41.0 in pkg/instrumentation/trace/trace.go. To avoid the 'incompatible schema' error which we have in current HEAD (only fixed in #663).

klihub (commented):
If we expect #663 to be difficult or time consuming to digest, review and merge, then I think it would make sense to take this in and rebase #663 on this. If we do that , I think we need to make a small additional otel semconv/schema version synchronization change in this PR to fix the 'incompatible schema' otel error we have in HEAD, too.

@askervin @kad @uniemimu What do you guys think ?

fmuyassarov:
> If we expect #663 to be difficult or time consuming to digest, review and merge, then I think it would make sense to take this in and rebase #663 on this. If we do that , I think we need to make a small additional otel semconv/schema version synchronization change in this PR to fix the 'incompatible schema' otel error we have in HEAD, too.
> 
> @askervin @kad @uniemimu What do you guys think ?

In my personal view, I would prioritize https://github.com/containers/nri-plugins/pull/663 first and then move with this patch others.

uniemimu:
> > If we expect #663 to be difficult or time consuming to digest, review and merge, then I think it would make sense to take this in and rebase #663 on this. If we do that , I think we need to make a small additional otel semconv/schema version synchronization change in this PR to fix the 'incompatible schema' otel error we have in HEAD, too.
> > @askervin @kad @uniemimu What do you guys think ?
> 
> In my personal view, I would prioritize #663 first and then move with this patch others.

This PR is not really needed if #663 takes care of the Trivy findings. I will fix the thing pointed out by Krisztian anyway.

What I would like is to be able to run the Trivy against a clean main branch preferably today. Whether 663 or 688 gets merged, I don't care at all.

kad:
I think it is better to update #663 to latest versions and get it merged.

klihub (approved):

PR #690, opened 2026-06-18
memtierd: harden temporary file handling

askervin:
Avoid predictable /tmp paths, restrict permissions on per-container runtime directories and files (0700/0600/0400), reject unsafe path segments, refuse symlinks under the run directory, and write the generated memtierd config atomically with O_EXCL|O_NOFOLLOW.

In the Helm chart, always pass --run-dir explicitly and mount an in-memory emptyDir when outputDir is not set, so state never lands in /tmp or the writable container layer.

### DIFF
--- a/cmd/plugins/memtierd/main.go
+++ b/cmd/plugins/memtierd/main.go
@@ -36,6 +36,99 @@ import (
 	_ "github.com/containers/nri-plugins/cmd/plugins/memtierd/licensechack"
 )
 
+// Permissions for memtierd's per-container runtime directory tree.
+const (
+	runDirPerm  os.FileMode = 0700
+	cfgFilePerm os.FileMode = 0600
+	outFilePerm os.FileMode = 0600
+	pidFilePerm os.FileMode = 0400
+)
+
+// defaultRunDirPath is used when --run-dir is not given.
+const defaultRunDirPath = "/var/run/nri-memtierd"
+
+// sanitizePathSegment rejects strings that are not safe to use as a
+// single path component.
+func sanitizePathSegment(s string) error {
+	if s == "" {
+		return fmt.Errorf("empty path segment")
+	}
+	if s == "." || s == ".." {
+		return fmt.Errorf("path segment %q is not allowed", s)
+	}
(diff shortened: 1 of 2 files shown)


cmd/plugins/memtierd/main.go
```diff
@@ -36,6 +36,101 @@ import (
...
+}
+
+// ensureSecureDir creates runDir/seg1/seg2/... with runDirPerm. Each
+// existing segment must be a non-symlink directory and is chmod'd to
+// runDirPerm.
+func ensureSecureDir(runDir string, segments ...string) error {
+	if err := os.MkdirAll(runDir, runDirPerm); err != nil {
+		return err
+	}
+	cur := runDir
```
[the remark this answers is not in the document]
askervin: We manage permissions of only the directories that we own, and do not change permissions in directories managed (created) by others. In the latter case we trust that managers know what they do, and we do not cause unexpected side effects in the system.

deployment/helm/memtierd/templates/daemonset.yaml
```diff
@@ -58,9 +58,11 @@ spec:
             - "{{ .Values.nri.plugin.index | int | printf "%02d"  }}"
             - --config
             - /etc/nri/memtierd/config.yaml
-            {{- if .Values.outputDir }}
             - --run-dir
+            {{- if .Values.outputDir }}
```
klihub, line 62: @askervin This PR LGTM, but I have one question/comment related to the `outputDir` Helm value. I found a bit hard to understand that parameter. Based on its name and type, which looks to be a string as it is set by default to `""`, I got the impression that it would be the name of the/some 'output directory' to be used. But looking at this PR and then the Helm chart template, looks like it is semantically used as a boolean. Couldn't it be a boolean, set by default to `false` in values.yaml ?

klihub: Ah, okay. I see now that it is indeed used as string value for an 'output directory' if set, with an emptyDir fallback if unset. Sorry for my initial confusion.

klihub (approved):

PR #692, opened 2026-06-18
fixes: fix hint allow-/deny-list filtering tests. 

klihub:
Fix broken hint filtering tests. Instead of trying to set up a fake sysfs that would result in the desired topology hint generation for the test hosts /boot filesystem, enable filtering for annotated test hints and use annotated hints for testing the filtering logic.


askervin (approved):
LGTM.

golangci-lint is a bit grumpy and wants os.Setenv() return value to be checked even if there is no way it could fail with this constant key and value.

PR #693, opened 2026-06-18
fix resctrl-mon: retain mon_group until pod sandbox removal

cmcantalupo:
The plugin removed a pod's mon_group in StopContainer as soon as the pod's last container stopped. For a pod with restartPolicy Always or OnFailure, the container restarts under the same pod UID after the process exits (e.g. a workload that runs for a fixed duration). Removing the mon_group on the last container stop released the RMID, and the next PostCreateContainer recreated the group with a freshly assigned RMID.

The new RMID carries residual hardware counter values from whatever task previously held it, producing a counter discontinuity that surfaces as a false energy/bandwidth spike in downstream rate() consumers.

Tie the mon_group lifetime to the pod sandbox instead of the container:

- StopContainer now only drops the container from in-memory state and retains the mon_group when the last container stops.
- Add a RemovePodSandbox handler that releases the mon_group and its RMID at actual pod teardown. The NRI stub auto-subscribes to the RemovePodSandbox event via interface assertion, so no event-mask wiring is needed.

This keeps the RMID stable across container restarts. Pods are still keyed by UID (no per-container groups), so the missing-UID path remains a safe no-op.

Tests updated/added accordingly:
- TestMultiContainerPod asserts retention after last container stop and release on RemovePodSandbox.
- TestStopContainer_RemovesStateOnRmdirFailure -> TestRemovePodSandbox_RemovesStateOnRmdirFailure.
- TestCompactUID_StopContainerCleansUp -> TestCompactUID_RemovePodSandboxCleansUp.
- Add TestContainerRestart_RetainsMonGroup, TestRemovePodSandbox_UnknownPod and TestMissingPodUID_NoMonGroup.

### DIFF
--- a/cmd/plugins/resctrl-mon/plugin.go
+++ b/cmd/plugins/resctrl-mon/plugin.go
@@ -313,17 +313,50 @@ func (p *plugin) StopContainer(ctx context.Context, pod *api.PodSandbox, ctr *ap
 		return nil, nil
 	}
 
+	// Drop only the container from tracking. The mon_group is intentionally
+	// retained until the pod sandbox is removed (see RemovePodSandbox).
+	//
+	// A pod with restartPolicy Always/OnFailure restarts its container under
+	// the same pod UID after the process exits (e.g. a workload that runs for
+	// a fixed duration). If we removed the mon_group here, the kernel would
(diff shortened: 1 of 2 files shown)


askervin (approved):
LGTM.

klihub (approved):

PR #694, opened 2026-06-22
e2e: strict topology hint test fixes

klihub:
In strict topology hint tests, wait for the 'waiting.reason' of a container's status to become 'CreateContainerError' before checking for a particular error message from its status. Otherwise we might not get the status we look for.

### DIFF
--- a/test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
+++ b/test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
@@ -4,6 +4,21 @@ cleanup() {
     vm-command "kubectl delete namespaces highprio lowprio --now --ignore-not-found"
 }
 
+wait_for_waiting_status_reason() {
+    local pod=$1 ctridx=$2 status=$3
+    local maxtry=5 retry=0
+
+    while [ $retry -lt $maxtry ]; do
+        echo "Waiting for $pod/ctr$ctridx to enter status $status..."
+        vm-command "kubectl get pod $pod -o jsonpath='{.status.containerStatuses[$ctridx].state}' | \
+                        jq '.waiting.reason'"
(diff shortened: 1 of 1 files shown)


test/e2e/policies.test-suite/topology-aware/n4c16/test18-strict-alignment/code.var.sh
```diff
@@ -4,6 +4,23 @@ cleanup() {
...
 }
 
+wait_for_waiting_status_reason() {
+    local pod=$1 ctridx=$2 status=$3
+    local maxtry=5 retry=0
+
+    while [ $retry -lt $maxtry ]; do
+        echo "Waiting for $pod/ctr$ctridx to enter status $status..."
+        vm-command "kubectl get pod $pod -o jsonpath='{.status.containerStatuses[$ctridx].state}' | \
+                        jq '.waiting.reason'" | grep -q $status && break
```
askervin: Because `vm-command` prints extra stuff (like prompt and commands to be executed), grepping directly its output is a bit risky and hides some benefits of `vm-command`.

I'd propose first running `vm-command` and then grepping from `COMMAND_OUTPUT`. This would help debugging (let user see what `jq` finds):
```
vm-command "kubectl get pod $pod -o jsonpath='{.status.containerStatuses[$ctridx].state}' | \
                        jq '.waiting.reason'"
grep -q <<< "$COMMAND_OUTPUT" && return 0
```

Or if you think it's better to hide this whole thing from user, then it would be better to use `vm-command-q` instead of `vm-command`.

Once expected status is seen, could the function return immediately from the while loop? That is, without extra 1 second sleep and returning if the status is still expected?

klihub: Updated to use `$COMMAND_OUTPUT` and return immediately on grep success as suggested.

askervin (approved):
LGTM

PR #696, opened 2026-06-22
cache: avoid a test cache population data race

klihub:
Omit creating synchronization channel if we don't fetch pod resources, to avoid a data read/write race during test cache population in tests.


askervin (approved):
LGTM

PR #697, opened 2026-06-23
Update to goresctrl v0.13.0 and use new SST API

askervin:


klihub (commented):
This looks ok to me, at least nothing stands out as obviously wrong based on a readthrough.

klihub (approved):

pkg/cpuallocator/allocator.go
```diff
@@ -1476,59 +1483,72 @@ func (c *topologyCache) discoverSstCPUPriority(sys sysfs.System, pkgID idset.ID)
...
+				if !ok {
+					log.Debugf("package #%d punit #%d: unable to get package for SST-PP info, skipping", pkgID, punitID)
+					continue
+				}
+				plInfos, err = sstPkg.GetPerfLevelInfo(currentPerfLevel)
+				if err != nil || plInfos == nil {
+					log.Debugf("package #%d punit #%d: unable to get SST-PP perf level inf, skipping: %v", pkgID, punitID, err)
+					continue
+				}
+				plInfo, ok := plInfos[idset.ID(punitID)]
```
marquiz: This is now inside the ` if plInfos == nil` block. Looks incorrect(?)

pkg/cpuallocator/allocator.go
```diff
@@ -1476,59 +1483,72 @@ func (c *topologyCache) discoverSstCPUPriority(sys sysfs.System, pkgID idset.ID)
...
-	}
 
-	if !active && sst != nil && sst.BFEnabled {
-		log.Debugf("package #%d: using SST-BF based CPU prioritization", pkgID)
-		for _, i := range cpuIDs {
-			id := idset.ID(i)
-			p := PriorityLow
-			if sst.BFCores.Has(id) {
-				p = PriorityHigh
+		if !active && punit.PP.Supported && punit.BF.Enabled {
```
kad, line 1517: If BF.Enabled, I don't see path where it would mark as active = true.

pkg/sysfs/system.go
```diff
@@ -1839,27 +1846,39 @@ func (sys *system) discoverSst() error {
...
+	// For instance, onlined CPUs may have changed in old sys.sst.
+	sstPlatform, err := sst.Init()
+	if err != nil || sstPlatform == nil {
+		return fmt.Errorf("failed to initialize Speed Select Technology (SST) platform: %v", err)
+	}
+	sys.sst = sstPlatform
 
 	for _, pkg := range sys.packages {
-		sstInfo, err := sst.GetPackageInfo(pkg.id)
+		// sstInfo, err := sst.GetPackageInfo(pkg.id)
```
marquiz:
```suggestion
```

pkg/sysfs/system.go
```diff
@@ -1839,27 +1846,39 @@ func (sys *system) discoverSst() error {
...
+		sstPkg, ok := sys.sst.Package(pkg.id)
+		if !ok {
+			return fmt.Errorf("failed to get SST handle for package %d: %v", pkg.id, err)
+		}
+		//sys.DebugBlock("", "Speed Select Technology info detected for package %d:\n%s", pkg.id, utils.DumpJSON(sstInfo))
```
marquiz: Uncomment or drop

pkg/cpuallocator/allocator.go
```diff
@@ -1476,59 +1483,72 @@ func (c *topologyCache) discoverSstCPUPriority(sys sysfs.System, pkgID idset.ID)
...
+			var err error
+			log.Debugf("package #%d punit #%d: using SST-BF based CPU prioritization", pkgID, punitID)
+			currentPerfLevel := punit.PP.CurrentLevel
+			if plInfos == nil {
+				sstPkg, ok := sys.Sst().Package(pkgID)
+				if !ok {
+					log.Debugf("package #%d punit #%d: unable to get package for SST-PP info, skipping", pkgID, punitID)
+					continue
+				}
+				plInfos, err = sstPkg.GetPerfLevelInfo(currentPerfLevel)
```
kad, line 1527: in case of multiple punints, will it work correctly?

marquiz (commented):
Looks good, generally. Two nits and one real review comment below.

pkg/sysfs/system.go
```diff
@@ -1839,27 +1846,39 @@ func (sys *system) discoverSst() error {
...
+		return fmt.Errorf("failed to initialize Speed Select Technology (SST) platform: %v", err)
+	}
+	sys.sst = sstPlatform
 
 	for _, pkg := range sys.packages {
-		sstInfo, err := sst.GetPackageInfo(pkg.id)
+		// sstInfo, err := sst.GetPackageInfo(pkg.id)
+		sstPkg, ok := sys.sst.Package(pkg.id)
+		if !ok {
+			return fmt.Errorf("failed to get SST handle for package %d: %v", pkg.id, err)
```
kad: this `err` comes from where? drop it?

pkg/sysfs/system.go
```diff
@@ -1839,27 +1846,39 @@ func (sys *system) discoverSst() error {
...
+					clos, err := sys.sst.GetCPUClosID(id)
+					if err != nil {
+						return fmt.Errorf("failed to get SST-CP clos id for cpu %d: %v", id, err)
+					}
+					sys.cpus[id].sstClos = clos
 				}
-
-				sys.cpus[id].sstClos = clos
 			}
+			pkg.sstInfo = pkgStatus
```
kad: in case of multiple punits, it would be the last one?

pkg/sysfs/system.go
kad: cleanup before merge?

pkg/cpuallocator/allocator.go
```diff
@@ -1476,59 +1483,72 @@ func (c *topologyCache) discoverSstCPUPriority(sys sysfs.System, pkgID idset.ID)
...
+				if err != nil || plInfos == nil {
+					log.Debugf("package #%d punit #%d: unable to get SST-PP perf level inf, skipping: %v", pkgID, punitID, err)
```
kad: typo, `info`

kad (changes requested):

pkg/sysfs/system.go
askervin: Thanks, fixed.

pkg/sysfs/system.go
askervin: Thanks, dropped.

pkg/cpuallocator/allocator.go
askervin: Great catch! Only the first punit got handled properly. Now it's out of that if-branch and both punits will get their HighPriorityCPUs iterated and prioritized accordingly.

pkg/cpuallocator/allocator.go
askervin: Thanks! Definitely, that was missing. Fixed. I'll keep the original behavior: mark active=true even if there were no HighPriorityCPUs --- it's enough if BF was enabled.

pkg/cpuallocator/allocator.go
askervin: I would expect so... pkg.GetPerfLevelInfo returns a map associating package's every punit id to punit's PerfLevelInfo that contains BF.HighPriorityCPUs in that punit. We are iterating through punits and verified that punit's BF is enabled.

pkg/sysfs/system.go
askervin: Thanks, fixed!

pkg/sysfs/system.go
askervin: pkgStatus contains info on all punits. This implementation set pkg.sstInfo, but in a suspicious location. Thanks for pointing this out!

I moved setting pkg.sstInfo out of the punit iteration. Now pkg.sstInfo gets set if all sstClos's could be set without errors.

pkg/sysfs/system.go
askervin: cleaned up.

pkg/cpuallocator/allocator.go
askervin: Thanks, fixed.

klihub (approved):

kad (approved):
LGTM

PR #700, opened 2026-06-23
e2e: detect and try to fix potential post reboot image blob corruption.

klihub:
Under certain preconditions, one known such being running with BTRFS and a (recent) self-compiled kernel from Torvalds git tree, seems to trigger a BTRFS bug where some image blob files become unreadable, with read failing with `EOPNOTSUPP`.

This is regularly triggered by `balloons/n4c16/test30-numa-disabled` test case. That failure in turn causes all the remaining test cases running on the same VM (IOW with the same emulated HW topology) to unconditionally fail as, among other things, all container creations fail afterwards. Since this is the last balloons test case for that topology, it causes all topology-aware tests on the same topology to fail.

To work around this (add a function, a rather big hammer to) try to detect and fix up when the runtime gets into such a condition. Currently this is only implemented for containerd. CRI-O is a TODO.

### DIFF
--- a/test/e2e/lib/vm.bash
+++ b/test/e2e/lib/vm.bash
@@ -988,3 +988,129 @@ vm-kernel-pkgs-uninstall() { # script API
 
     vm-command "uname -a"
 }
+
+vm-post-reboot-runtime-check() { # script API
+    # Usage: vm-post-reboot-runtime-check [image_type]
+    #
+    # Check whether a reboot rendered some runtime image blobs unreadable
+    # and try to fix it. Optionally also reimport and tag a local image
+    # from a tarball for the given (balloons or topology-aware) image type.
+    #
+    # The blob problem is known to be triggerable when the underlying fs
+    # is BTRFS and the old or new running kernel was from Torvalds tree
+    # ('vanilla' in provisioning). The exact reasons are unknown.
+
+    local image_type=$1
+
+    if ! $k8scri-check-unreadable-blobs; then
+        echo "no post-boot runtime blob read failures detected..."
+        return 0
+    fi
+
+    echo "post-boot runtime blob read failures detected... trying to fix them"
+    vm-command "systemctl stop kubelet"
+    vm-command "systemctl stop $k8scri"
+    vm-command "killall -9 etcd"
+    vm-command "killall -9 kube-apiserver"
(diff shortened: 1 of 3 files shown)


test/e2e/lib/vm.bash
```diff
@@ -988,3 +988,129 @@ vm-kernel-pkgs-uninstall() { # script API
...
+    if [ -z "$ref" ]; then
+        command-error "failed to resolve image ref for sha256:$sha256"
+    fi
+
+    vm-command "ctr -n k8s.io images tag --force $ref $tag" || \
+        command-error "failed to tag reimported image"
+}
+
+crio-check-unreadable-blobs() {
+    command-error "TODO: implement unreadable blob check for CRI-O..."
```
askervin: `error` works better than `command-error` here. The latter prints extra information from previously executed `vm-command` or `host-command` to help user debug the command that caused the error. In crio-* functions we don't want to associate the error to previously executed commands, so `error` is enough.

test/e2e/policies.test-suite/balloons/n4c16/test30-numa-disabled/code.var.sh
```diff
@@ -1,4 +1,7 @@
-vm-command '[ -d /sys/devices/system/node ]' && vm-kernel-pkgs-install
+vm-command '[ -d /sys/devices/system/node ]' && {
```
askervin, line 1: We probably have exactly the same issue in

policies.test-suite/topology-aware/n4c16/test30-numa-disabled/code.var.sh

So it would be great update it in this PR, too. And it offers possibility to test that
`vm-post-reboot-runtime-check topology-aware`
works as expected.

askervin (commented):
Looks nice. Two comments...

test/e2e/lib/vm.bash
klihub: Updated.

test/e2e/policies.test-suite/balloons/n4c16/test30-numa-disabled/code.var.sh
klihub: Added exact same fixups to the corresponding topology-aware test case and tested that it works identically.

askervin (approved):
LGTM

PR #701, opened 2026-06-24
Makefile: make sure test cleanup is not run in parallel with tests.

klihub:
Make sure any package/test-specific cleanup scripts are only run after the tests themselves have completed.


askervin (approved):
LGTM

PR #702, opened 2026-06-24
pkg/topology,Makefile: fix flaky tests, ensure subpackage tests run.

klihub:
Fix flaky pkg/topology tests(etup). Make sure subpackage tests are run.

### DIFF
--- a/pkg/topology/topology_test.go
+++ b/pkg/topology/topology_test.go
@@ -24,10 +24,23 @@ import (
 	"testing"
 )
 
+type logger int
+
+func (logger) Debugf(format string, v ...interface{}) {
+	fmt.Printf("D: [topology] "+format+"\n", v...)
+}
+
+var pkgDir string
+
 func setupTestEnv(t *testing.T) func() {
+	SetLogger(logger(0))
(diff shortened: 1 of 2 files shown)


pkg/topology/topology_test.go
```diff
@@ -24,18 +24,35 @@ import (
...
+	if pkgDir == "" {
+		pkgDir = pwd
+	}
 	if path, err := filepath.EvalSymlinks(pwd); err == nil {
 		pwd = path
 	}
 	SetSysRoot(pwd + "/testdata")
 	teardown := func() {
 		SetSysRoot("")
+		if pkgDir != "" {
```
marquiz: pkgDir should never be empty string, right(?) so this could be removed

marquiz (approved):
Lookgs good, one small nit comment/question

pkg/topology/topology_test.go
klihub: True, it should never be empty at that point. Removed emptiness test.

askervin (approved):
LGTM

PR #703, opened 2026-06-24
e2e: show active test case in 'VM prompt'.

klihub:
Instead of using a static `root@vm>` "VM prompt", use `root@vm $policy/$testcase>`.


askervin (approved):
LGTM

PR #704, opened 2026-06-24
e2e: more robust otel log collection test.

klihub:
Don't expect otel logs to be collected by the time the test pod is ready. Check and wait for the last container to be running before trying to verify collected logs.


askervin (approved):
LGTM

PR #706, opened 2026-06-25
e2e: give otel log dumping more slack.

klihub:
Don't expect collected logs to be dumped to the filesystem by the time the last container is running. Retry log verification with a 5 second timeout (plenty wrt. the 1 second collection batch buffering).


askervin (approved):
LGTM

PR #707, opened 2026-06-25
Dockerfiles: work around parallel make license extraction breakage.

klihub:
Work around parallel make (commit b3b8d54ae6) breaking license extraction during dockerized image building, but do it without adding a target dependency on `install-go-licenses` to `licenses`.


askervin (approved):
LGTM

PR #708, opened 2026-06-26
e2e: use nop BTRFS/blob corruption handlers for cri-o.

klihub:
It looks like the post-reboot BTRFS/blob corruption bug is not triggered with cri-o as a runtime, only with containerd as surprising as this may sound considering that it is an fs-level corruption (unreadable blob file).

Anyway, switch the cri-o-specfic handlers to mere logging nops for now. We'll revert this and look into it again if we ever see it triggered.


askervin (approved):
LGTM

PR #711, opened 2026-06-29
balloons: fix publishing PCT HP CPU cores as extended resources

askervin:


askervin:
@marquiz, I broke this feature in a last-minute "fix" before release (f206c26447d400865816e9a234298058fcfdf49b), didn't have regression test for it, and trusted that it would work without giving it a shot in a real system. As a last commit and with all these red flags it was easy to spot. What a stupid error in both the code and the process...

marquiz (approved):
Ach, sorry about my last-minute churn 😬 

LTM

askervin:
It was a good point. Only the implementation and the process sucked. 😖

fmuyassarov (approved):
LGTM

PR #720, opened 2026-07-22
balloons: support affinity to extended resources from device plugins

askervin:
preferCloseToDevices supports "podresourceapi:example.com/device"

Enable CPU selection based on locality of individual container's physical devices. podresourceapi:RESOURCE syntax allows specifying which physical device(s) (matching logical resources to given glob-expression) is preferred to be local to container's CPUs.

In order to allocate all container's CPUs near container's physical device, the container must be placed into a new balloon. Technically, this changes balloon creation internals: so far CPU selection to a new balloon has been affected by balloon types and instances only, but after this change the first container put into a new balloon may affect it, too.

### DIFF
--- a/cmd/plugins/balloons/policy/balloons-policy.go
+++ b/cmd/plugins/balloons/policy/balloons-policy.go
@@ -16,13 +16,17 @@ package balloons
 
 import (
 	"fmt"
+	"maps"
 	"math"
+	"path"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
 	"sync"
 
+	"github.com/containers/nri-plugins/pkg/agent/podresapi"
 	cfgapi "github.com/containers/nri-plugins/pkg/apis/config/v1alpha1/resmgr/policy/balloons"
 	"github.com/containers/nri-plugins/pkg/cpuallocator"
 	"github.com/containers/nri-plugins/pkg/irq"
@@ -1080,7 +1084,7 @@ func (p *balloons) virtDevsChangeDuringCpuAllocation(loadClassNames []string) bo
 	return false
 }
 
-func (p *balloons) newCompositeBalloon(blnDef *BalloonDef, confCpus bool, freeInstance int) (*Balloon, error) {
+func (p *balloons) newCompositeBalloon(blnDef *BalloonDef, confCpus bool, freeInstance int, c cache.Container) (*Balloon, error) {
 	componentBlns := make([]*Balloon, 0, len(blnDef.Components))
 	deleteComponentBlns := func() {
 		for _, compBln := range componentBlns {
@@ -1099,7 +1103,7 @@ func (p *balloons) newCompositeBalloon(blnDef *BalloonDef, confCpus bool, freeIn
 				return nil, balloonsError("unknown balloon definition %q in composite balloon %q",
 					comp.DefName, blnDef.Name)
 			}
-			compBln, err := p.newBalloon(compDef, confCpus)
+			compBln, err := p.newBalloon(compDef, confCpus, c)
 			if err != nil || compBln == nil {
 				deleteComponentBlns()
 				return nil, balloonsError("failed to create component balloon %q for composite balloon %q: %v",
@@ -1137,7 +1141,7 @@ func (p *balloons) newCompositeBalloon(blnDef *BalloonDef, confCpus bool, freeIn
 			return nil, balloonsError("unknown balloon definition %q in balanced composite balloon %q",
 				comp.DefName, blnDef.Name)
 		}
-		compBln, err := p.newBalloon(compDef, confCpus)
+		compBln, err := p.newBalloon(compDef, confCpus, c)
 		if err != nil || compBln == nil {
 			deleteComponentBlns()
 			return nil, balloonsError("failed to create component balloon %q for composite balloon %q: %v",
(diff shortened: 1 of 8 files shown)


cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -2163,19 +2187,177 @@ func mergeCpuClassHints(opts *cpuTreeAllocatorOptions, provider cpuClassHints, i
...
+			dev.GetResourceName(), c.PrettyName(), resourceName, devNumas, dev.GetDeviceIds())
+	}
+	if len(numas) == 0 {
+		return emptyCpuSet, false
+	}
+	numaIDs := make([]string, 0, len(numas))
+	for id := range numas {
+		numaIDs = append(numaIDs, strconv.FormatInt(id, 10))
+	}
+	cpuStr := p.options.System.NodeHintToCPUs(strings.Join(numaIDs, ","))
```
klihub: nit: if we split out the cpuset construction bits from sysfs.System.NodeHintToCPUs() as sysfs.System.NodeHintToCPUSets() we woudn't need to construct a cpuset, convert it to a string, then parse it to a cpuset again here.  Or, since before this PR it is only used in one place, we can also change the existing function to return the resulting CPUSet instead of its string representation and update the existing call site.

klihub (commented):
Looks good to me. My minor nit is related to an existing interface.

klihub (approved):

test/e2e/policies.test-suite/balloons/n4c16/test24-podresources/code.var.sh
```diff
@@ -0,0 +1,175 @@
...
+declare -a EXTLIM=( "tech.com/tpu: \"1\"" "telco.com/nic: \"1\"" "cpuclass.balloons.nri.io/pct-hp: \"2\"" )
+CPUREQ=2 CPULIM=2 MEMREQ=10M MEMLIM=10M \
+       POD_ANNOTATION="balloon.balloons.resource-policy.nri.io: hp-near-tpu" \
+       CONTCOUNT=1 \
+       create balloons-busybox
+report allowed
+verify-podres-locality "tech.com/tpu" pod2c0
+
+declare -a EXTREQ=( "tech.com/tpu: \"1\"" "telco.com/nic: \"1\"" )
+declare -a EXTREQ=( "tech.com/tpu: \"1\"" "telco.com/nic: \"1\"" )
```
kad: EXTLIM?

kad (commented):
one issue, seems like copy&paste typo. 
new unit tests for `podResourceHintDevName` and `podResourceDeviceName` might be good (not critical).

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -2163,19 +2187,177 @@ func mergeCpuClassHints(opts *cpuTreeAllocatorOptions, provider cpuClassHints, i
...
+	if cpuStr == "" {
+		return emptyCpuSet, false
+	}
+	cpus, err := cpuset.Parse(cpuStr)
+	if err != nil {
+		log.Errorf("failed to parse CPUs %q of device %q for container %s: %v",
+			cpuStr, resourceName, c.PrettyName(), err)
+		return emptyCpuSet, false
+	}
+	return cpus.Intersection(p.allowed), true
```
fmuyassarov, line 2306: Probably not a big deal, but I want to confirm that this is intentional. If I'm not wrong, we are saying here locality is resolved and use this preference but the preference can be an empty cpuset? Should we actually check if we got an empty list by time time we arrive here and if yes return `false` and not `true`?
```GO
cpus = cpus.Intersection(p.allowed)
if cpus.IsEmpty() {
    return emptyCpuSet, false
}
return cpus, true
```

cmd/plugins/balloons/policy/balloons-policy.go
askervin: I ended up building node-to-CPU mapping using system.Node.CPUs. The reason is that the source of this node list is not our system (sysfs) library, and therefore I didn't want to trust that every nodeID being valid. The cost of trusting an invalid node ID is a crash.

Validating nodeID turned out to be slightly hairier than I thought, as `system.Node(untrustedNodeID) == nil` doesn't work: it returns non-nil interface, yet underlying node ==nil. However, `system.FilterNode(untrustedNodeID)` is useful and fast enough for validation, so I didn't extend system interface either.

Would this solution be good enough? The other option would indeed be extending system interface with "NodeToCPUs", but it would be overlapping and would still leave validation/error logging to be done somewhere or silently ignored (that would not sound good).

test/e2e/policies.test-suite/balloons/n4c16/test24-podresources/code.var.sh
askervin: Thanks for spotting this, fixed!

askervin: Unit tests added, too.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Yes, this is intentional. `false` as the second return value means there is not locality information available for any reason. `true` means that the function was able to get the information, yet the resulting cpuset may still be empty. This may happen when the device is located on a NUMA node whose CPUs are not in balloons policy's AvailableResources. If that happens, CPU allocation will just have one hint that it cannot use, but at least there is a chance handling such a thing (logging, for instance).

kad (approved):
LGTM

fmuyassarov (approved):
Thank you @askervin 
LGTM

fmuyassarov:
I'm merging it assuming comments are addressed.

PR #721, opened 2026-07-22
balloons: clean up unnecessary (non-free) and possibly misleading (insufficient) PCT CPU hints

askervin:
Developed on top of PR #719 to avoid conflict.

### DIFF
--- a/pkg/resmgr/cpuclass/internal/pct/pct.go
+++ b/pkg/resmgr/cpuclass/internal/pct/pct.go
@@ -919,14 +919,20 @@ func (a *Allocator) Hints(intent types.AllocationIntent) types.AllocationHints {
 	if a == nil || !a.Active() {
 		return types.AllocationHints{}
 	}
+	if intent.RequestedCount < 1 {
+		// Not implemented for shrinking CPUsets (negative RequestedCount).
+		// No hints for allocations of unknown size (RequestedCount == 0).
+		return types.AllocationHints{}
+	}
 	out := types.AllocationHints{}
 
 	if closID, ok := a.classClosID(intent.ClassName); ok {
 		closCpus := a.closCpus(closID)
-		if !closCpus.IsEmpty() {
+		freeClosCpus := closCpus.Intersection(intent.FreeCpus)
+		if freeClosCpus.Size() >= intent.RequestedCount {
 			out.Prefer = append(out.Prefer, types.CpuPreference{
 				Name: virtDevSstClosHint(closID),
-				Cpus: []cpuset.CPUSet{closCpus},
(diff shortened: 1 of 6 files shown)


pkg/resmgr/cpuclass/internal/pct/pct.go
```diff
@@ -923,10 +923,11 @@ func (a *Allocator) Hints(intent types.AllocationIntent) types.AllocationHints {
 
 	if closID, ok := a.classClosID(intent.ClassName); ok {
 		closCpus := a.closCpus(closID)
-		if !closCpus.IsEmpty() {
+		freeClosCpus := closCpus.Intersection(intent.FreeCpus)
+		if !freeClosCpus.IsEmpty() && freeClosCpus.Size() >= intent.RequestedCount {
```
klihub: nit: is the `freeClosCpus.IsEmpty()` test necessary here ? Can intent.RequestedCount be 0 ?

klihub (approved):
LGTM. With a minor nit-question.

pkg/resmgr/cpuclass/internal/pct/pct.go
```diff
@@ -923,10 +923,11 @@ func (a *Allocator) Hints(intent types.AllocationIntent) types.AllocationHints {
 
 	if closID, ok := a.classClosID(intent.ClassName); ok {
 		closCpus := a.closCpus(closID)
-		if !closCpus.IsEmpty() {
+		freeClosCpus := closCpus.Intersection(intent.FreeCpus)
+		if freeClosCpus.Size() >= intent.RequestedCount {
```
kad, line 932: `!freeClosCpus.IsEmpty() && freeClosCpus.Size() >= intent.RequestedCount`

?

fmuyassarov (commented):
LGTM, apart from @kad 's comment

pkg/resmgr/cpuclass/internal/pct/pct.go
askervin: Because the IsEmpty() check makes a difference only if RequestedCount equals 0, and because that particular case was far from clear in the existing implementation, I ended up writing a bit bigger bunch of changes, mostly non-functional (read: safe) stuff.

This bunch started from clearly documenting meaning of negative, zero and positive RequestedCount in the allocation intent, declaring 0 as reserved/unspecified at the moment, and handling RequestedCount <= 0 separately in the beginning of this function. This required some changes in unit tests, too. But this change does not affect topology-aware or balloons policy allocation hints because there RequestedCount > 0 whenever hints matter.

kad (approved):
LGTM

fmuyassarov (approved):
lgtm

PR #722, opened 2026-07-23
balloons: IRQ CPU affinity controls

askervin:
Control IRQ CPU affinities using new `irqClaim` and `irqMode` balloon type options.

For each IRQ, the logic is:
1. Find balloons that want exactly this IRQ (irqClaim matches) and set IRQ's CPU affinity to union of their CPUs.
2. If no balloon claimed the IRQ, then find balloons that want all IRQs (irqMode: sink), and set CPU affinity to union of sink balloons' CPUs.
3. With no claims and sinks around for the IRQ, find balloons that do not want any IRQs (irqMode: isolate), and set CPU affinity to all the remaining CPUs.
4. If no irqClaim and irqMode options are present in existing balloon instances, but they are used in some balloon types, let IRQs run on all CPUs.
5. If no irqClaim and irqMode options are present in any balloon types, do not touch IRQ CPU affinities in the system.

Combining irqClaim with "irqMode: isolate" in the same balloon type creates balloons that handle only claimed IRQs and no others.


cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -931,6 +936,98 @@ func (p *balloons) forgetCpuClass(bln *Balloon) {
 	}
 }
 
+const (
+	// irqModeSink adds CPUs of a balloon to the affinity of
+	// unclaimed IRQs.
+	irqModeSink = "sink"
+	// irqModeIsolate removes CPUs of a balloon from the affinity
+	// of IRQs that are neither claimed nor sinked.
+	irqModeIsolate = "isolate"
```
klihub: nit: IMO it would be better to define these constants in the previous commit under `pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go` (since that is that where it is effectively defined that these plus "" are the valid modes) and then alias them here with a const-assignment.

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -1526,6 +1623,7 @@ func (p *balloons) Reconfigure(newCfg interface{}) error {
 	p.BlockMeters()
 	defer p.UnblockMeters()
 	defer p.commitCpuClasses()
+	defer p.applyIrqAffinities()
```
klihub, line 1617: Related to configuration and reconfiguration: is my understanding correct that an `irqMode: sink, irqClaim: [ $IRQ_CLAIMS ]` is not sensible/meaningful combination for any single balloon type. If it is, should we verify against it in `pkg/apis/config/*/resmgr/policy/balloons.Config.Validate()` ?

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	cfgapi "github.com/containers/nri-plugins/pkg/apis/config/v1alpha1/resmgr/policy/balloons"
 	"github.com/containers/nri-plugins/pkg/cpuallocator"
+	"github.com/containers/nri-plugins/pkg/irq"
```
klihub, line 28: Related to the commit message of commit "balloons: implement IRQ affinity control", which AFAICT github does not allow to comment on. Is the following description of sink mode

"""
...
For each IRQ, the logic is: find balloons that want exactly this
IRQ (irqClaim). If nobody, then find balloons that want all
IRQs (irqMode: sink).
...
"""

accurate ? Or should it be "... If nobody, then find balloons that want all
otherwise unclaimed IRQs (irqMode: sink). ..." ?

klihub (commented):
I have a few minor questions and comments.

klihub (commented):
One more minor nit.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Good point, thanks! Fixed.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Yes, that's the most obscure combination.

On the other hand, if almost all balloon types would require isolation from IRQs, declaring the few non-sensitive balloon types as IRQ sinks might be more convenient than declaring all latency-sensitive-tailored balloon types isolated from IRQs. Under these circumstances it could be that a non-sensitive balloon type would also want to claim a certain IRQ, yet still being ok to participate in balancing all other IRQs, too.

In other words, a sink+claim balloon among other sink balloons would make equally sense than (plain) claim balloon among dont-care balloons when many isolated balloons are present.

Therefore, I wouldn't add a verification against this configuration. I hope that sink/isolate/claim semantics would be clear enough (and documented clearly enough) for users to understand what happens in this case.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Thanks, added "unclaimed".

fmuyassarov:
Thanks @askervin . This is a very useful addition and actually something that we have been thinking about, but more on the topology-aware plugin.  I will take my time tomorrow and go through the changes.

klihub (approved):
LGTM.

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -307,6 +307,19 @@ type BalloonDef struct {
...
+	// pattern matching a device or chip name in /proc/interrupts
+	IrqClaim []string `json:"irqClaim,omitempty"`
+	// IrqMode controls how CPU affinities of unclaimed (see
+	// irqClaim) IRQs relate to the CPUs of balloons of this
+	// type. "sink" adds the CPUs to the affinity of IRQs. If sink
+	// balloons are present, they take unclaimed IRQs from other
+	// balloons. "isolate" removes the CPUs of this balloon type
+	// from the affinity of IRQs that are neither claimed nor
+	// sinked.  +kubebuilder:validation:Enum="";sink;isolate
+	IrqMode string `json:"irqMode,omitempty"`
```
klihub, line 322: @askervin A late question regarding validation. Do we intentionally accept balloon definitions with a non-empty `IrqClaim` but empty `IrqMode` ?

IIUC, that causes `irqOptionsInUse()` to return true (checks if at least one of those is nonempty, not if both of those is nonempty). But then if I interpret `applyIrqAffinities()` correctly, if we only have one such balloon and only one such balloon, then effectively we end up all IRQs getting sinked to the p.allowed, the allowed CPU set of balloons. And that does not immediately seem like an intuitive end result or even an intentional one to me.

So shouldn't we protect against it by either rejecting a configuration with such a balloon, or by changing the condition in `irqOptionsInUse()` to test for nonemptiness of both `IrqClaim` and `IrqMode` ?

cmd/plugins/balloons/policy/balloons-policy.go
```diff
@@ -931,6 +936,89 @@ func (p *balloons) forgetCpuClass(bln *Balloon) {
 	}
 }
 
+// irqOptionsInUse returns whether any balloon type configures IRQ
+// affinities.
+func (p *balloons) irqOptionsInUse() bool {
+	for _, blnDef := range p.bpoptions.BalloonDefs {
+		if len(blnDef.IrqClaim) > 0 || blnDef.IrqMode != "" {
```
klihub, line 943: @askervin Shouldn't this be `len(blnDef.IrqClaim) > 0 && blnDef.IrqMode != ""` ? See my other question/comment about balloon definition with a nonempty `IrqClaim` but empty `IrqMode`..

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
klihub: No sorry... I think I misread the code. The effective end result of such a (quite normal setup) IIUC is setting the affinity of the claimed IRQs to the union of CPUs of balloon instances, which a very intuitive end result.

cmd/plugins/balloons/policy/balloons-policy.go
klihub: Nope, sorry for my confusion. It shouldn't be...

cmd/plugins/balloons/policy/balloons-policy.go
klihub: > Yes, that's the most obscure combination.
> 
> On the other hand, if almost all balloon types would require isolation from IRQs, declaring the few non-sensitive balloon types as IRQ sinks might be more convenient than declaring all latency-sensitive-tailored balloon types isolated from IRQs. Under these circumstances it could be that a non-sensitive balloon type would also want to claim a certain IRQ, yet still being ok to participate in balancing all other IRQs, too.
> 
> In other words, a sink+claim balloon among other sink balloons would make equally sense than (plain) claim balloon among dont-care balloons when many isolated balloons are present.

Okay, so what I distill from this is that semantically a sink+claim balloon means: 'I care about these interrupts and want the kernel to handle them on my CPUs, but I am not latency sensitive wrt. to the kernel potentially handling also other unrelated IRQs on my CPUs, too'. Right ?

fmuyassarov (approved):
LGTM

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
askervin: That's correct.

cmd/plugins/balloons/policy/balloons-policy.go
askervin: No problem!

cmd/plugins/balloons/policy/balloons-policy.go
askervin: Exactly.

PR #724, opened 2026-07-28
e2e: update fedora image locations.

klihub:
Update fedora image locations to add already released 44 and match recent changes/archival of 42.


askervin (approved):
LGTM

PR #725, opened 2026-07-28
e2e: accept/fix v-prefixed k8s, containerd and crio releases.

klihub:
Fix up explicitly given v-prefixed releases similarly to how we do for resolved "latest" release. This allows giving releases both without and with v-prefix (which was already allowed for some) but fixes provisioning errors with v-prefixes which result from incorrectly generated repo/tarball locations (with double v-prefix).


askervin (approved):
LGTM

PR #727, opened 2026-07-29
resmgr: name to container ID mapping fixes

klihub:
This patch series improves container name to ID mapping, used for transient duplicate detection.

In particular, we now mark containers as exited during initial state synchronization with the runtime. This should prevent subsequent configuration updates from trying to allocate resources for both the old and the new instance, in case the old has not been removed yet. Also during initial sync, remove unmapped (disambiguated older duplicate) instances from the list of containers to be (re)allocated. This should prevent double accounting of resources for transient duplicates.
    
Additionally, when trying to detect transient stale duplicate containers for overlapping container create/crash/stop then re-create events, try to be smarter about which instance is the stale one. For a CreateContainer event always choose the old one. Otherwise for an initial sync, use the CreatedAt timestamps if available. If not, then assume that a container in running state is stale if the other one is in created state.

Finally, add a new e2e test case to trigger and test the handling of transient dupicates.

### DIFF
--- a/pkg/resmgr/nri.go
+++ b/pkg/resmgr/nri.go
@@ -123,13 +123,13 @@ func (p *nriPlugin) onClose() {
 	os.Exit(1)
 }
 
-func (p *nriPlugin) syncNamesToContainers(containers []cache.Container) []cache.Container {
-	unmapped := make([]cache.Container, 0, len(p.byname))
+func (p *nriPlugin) syncNamesToContainers(containers []cache.Container) map[string]cache.Container {
+	unmapped := map[string]cache.Container{}
 
 	for _, ctr := range containers {
 		old := p.mapNameToContainer(ctr)
-		if old != nil && old.GetID() != ctr.GetID() {
-			unmapped = append(unmapped, old)
+		if old != nil {
+			unmapped[old.GetID()] = old
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/resmgr/nri.go
```diff
@@ -140,13 +140,43 @@ func (p *nriPlugin) mapNameToContainer(ctr cache.Container) cache.Container {
...
+			p.byname[name] = ctr
+			nri.Infof("%s: remapped container from %s to %s (by state)",
+				name, old.GetID(), ctr.GetID())
+			return old
+		case ctr.GetState() == cache.ContainerStateRunning &&
+			old.GetState() == cache.ContainerStateCreated:
+			nri.Infof("%s: keeping container mapped to %s (by state)", name, old.GetID())
+			return ctr
+		default:
+			nri.Infof("%s: remapped container from %s to %s (as fallback, might be wrong)",
```
askervin: Should we add a Warnf (or even Errorf) when remapping a container with insufficient information to make the educated guess on which one is stale?

test/e2e/policies.test-suite/topology-aware/n4c16/test31-duplicate-disambiguation/code.var.sh
```diff
@@ -0,0 +1,97 @@
...
+
+wait-container-restart() {
+    local statuses="" pid=""
+
+    while ! [[ "$statuses" == "Running" ]]; do
+        vm-command "kubectl get pods -A --no-headers=true | tr -s '\t' ' '| cut -d ' ' -f4 | sort -u"
+        statuses=$COMMAND_OUTPUT
+        [[ "$statuses" == *"Running"*"Unknown"* ]] && (
+            vm-command "kubectl get pods -A -o name | grep topology-aware"
+            pid=$COMMAND_OUTPUT
```
askervin: nit: "pod" would be more familiar in this context than "pid".

askervin (commented):
Very nice... just one question and one nit.

test/e2e/policies.test-suite/topology-aware/n4c16/test31-duplicate-disambiguation/code.var.sh
klihub: Fixed.

pkg/resmgr/nri.go
klihub: Yes, should at least be a warning. Switched to Errorf.

klihub:
@askervin Addressed review comments. Also the actual fix was missing to prevent re-allocation of resource for old instances of transient duplicates. Added it an updated the test case to verify that the internal post-allocated duplicate detection runs clean, never detecting any duplicate resource allocations.

askervin (approved):
LGTM.

test/e2e/policies.test-suite/topology-aware/n4c16/test31-duplicate-disambiguation/code.var.sh
```diff
@@ -0,0 +1,155 @@
...
+
+    echo "Retrying..."
+    let retries=$retries+1
+
+    if [ $retries -ge 5 ]; then
+        echo "Max retries ($retries) reached, could not reproduce the issue, giving up"
+        break
+    fi
+
+    # Re-create all pods, otherwise CrashLoopBackup may prevents finding remaps.
```
askervin: My bad. `s/may//`

askervin (dismissed):
@klihub,

Yet I already put my approval on this, may I ask you to fix my mistake in the commit string?

I'd also propose removing "fixup: e2e:..." part from the commit message of:
"e2e: add test for duplicate disambiguation".
(Removes another spelling error: "ry" instead of "try", and we won't need that part in the commit message anyway.)

I think with these fixes this PR is good for merging.

klihub:
> @klihub,
> 
> Yet I already put my approval on this, may I ask you to fix my mistake in the commit string?
> 
> I'd also propose removing "fixup: e2e:..." part from the commit message of: "e2e: add test for duplicate disambiguation". (Removes another spelling error: "ry" instead of "try", and we won't need that part in the commit message anyway.)

Argh... that was just me fat-fingering the squashed commit message, unintentional. Thanks for spotting it.

test/e2e/policies.test-suite/topology-aware/n4c16/test31-duplicate-disambiguation/code.var.sh
klihub: Fixed.

askervin (approved):
LGTM, Thanks @klihub, this is a great fix!

PR #728, opened 2026-08-01
agent: log watch error reason.

klihub:
If we receive an error event from a k8s watch, log the watch error message to help figuring out if something is wrong with the watched object itself. Without the error message it can be really difficult to track down the root cause of the failure.

### DIFF
--- a/pkg/agent/watch/object.go
+++ b/pkg/agent/watch/object.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/watch"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/agent/watch/object.go
```diff
@@ -112,6 +113,12 @@ func (w *ObjectWatch) run() error {
 				}
 
 				if e.Type == Error {
+					log.Errorf("watch failure event: %+v", e)
+
+					if status, ok := e.Object.(*metav1.Status); ok {
+						log.Errorf("watch error: %s: %s", status.Reason, status.Message)
+					}
+
```
fmuyassarov: nit, maybe something like to reduce double logging?
```go
if e.Type == Error {
    if status, ok := e.Object.(*metav1.Status); ok {
        log.Errorf("watch %s failed: %s: %s", w.watchname(), status.Reason, status.Message)
    } else {
        log.Errorf("watch %s failed with unexpected event: %+v", w.watchname(), e)
    }
}
```

klihub: Updated.

askervin (approved):
LGTM

PR #729, opened 2026-08-03
topology-aware: add CPU class support.

klihub:
Please, review/merge #736 first.

This patch series adds basic CPU class support to the topology-aware policy.

With these patches in place, CPU classes can now be assigned to
- CPUs exclusively assigned to (Guaranteed QoS class) containers,
- CPUs of shared pools, and
- CPUs of the reserved pool

CPU class support is turned on by defining some CPU classes in the configuration and specifying at least the CPU class for shared pool by setting `sharedCPUClass` in the configuration. This is the minimal valid configuration.

Additionally, an optional reserved pool CPU class can be specified for CPUs of the reserved pool by setting `reservedCPUClass` in the configuration. An unset reserved pool class defaults to the shared pool class. Furthermore, also optionally a default CPU class for exclusive CPUs can be specified by setting `defaultExclusiveCPUClass` in the configuration. If set exclusive CPUs of otherwise unnannotated containers are assigned to the default exclusive class.

Containers can also be explicitly annotated with a CPU class using the `cpu-class[.resource-policy.nri.io]` effective annotation base key. Exclusive CPUs of annotated containers are assigned to their annotated class. Once CPUs are released from exclusive use and returned to a shared pool, they are (re-)assigned to the shared pool CPU class.

Note: This PR is stacked on top of #736. It is not github-stacked because (I think) PRs can't be github-stacked if the PR branches are not originating from the same source repository.


pkg/apis/config/v1alpha1/resmgr/policy/topologyaware/config.go
```diff
@@ -210,6 +232,36 @@ func (c *Config) Validate() error {
 		}
 	}
 
+	if len(c.CPUClasses) > 0 {
+		cc, _, _, err := ValidateCPUClasses(c.CPUClasses)
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		if c.SharedCpuClass == "" {
```
askervin, line 241: For convenience, balloons policy handles CPU class name "default" differently than others: if a balloon type does not specify any CPU class, the "default" CPU class will be applied on CPUs of its instances.

https://containers.github.io/nri-plugins/stable/docs/resource-policy/policy/balloons.html#cpu-tuning

I was thinking if it would make sense to use `c.{Shared,Reserved,DefaultExclusive}CpuClass = "default"` when a user has specified "default" CPU class. Pros: one could use the same CPU class configuration if evaluating and switching between policies, and it might be more intuitive than exclusive allocations using `sharedCPUClass` by default. Cons: if the existence of the "default" CPU class is not required and the current mechanism for figuring out a default class would be used in that case, it would make make the whole thing unnecessarily complex.

Having special semantics for a class named "default" fits quite naturally in balloons, as there is the "default" balloon type, too. In T-A, there is no similar context, so this might not be a good idea.

askervin (commented):
Adding one note... I'll continue review soon...

cmd/plugins/topology-aware/policy/resources.go
```diff
@@ -982,24 +1053,43 @@ func (cs *supply) GetScore(req Request) Score {
...
+				score.cpu = cpus
+				hints := p.cpuClasses.Hints(cpuclass.AllocationIntent{
+					ClassName:      cr.cpuClass,
+					CurrentCpus:    cpuset.New(),
+					FreeCpus:       cpus,
+					RequestedCount: cr.full,
+				})
+				if len(hints.Prefer) > 0 {
+					ccCpus := cpus.Intersection(hints.Prefer[0].Cpus)
+					if ccCpus.Equals(cpus) {
```
askervin: This will be more useful with #719 because then we can go through alternative (equally acceptable) cpuset hints. Then it'll be more likely that there is a cpuset that is aligned with the set from GetCPUOffer().

cmd/plugins/topology-aware/policy/pools.go
```diff
@@ -417,6 +417,15 @@ func (p *policy) allocatePool(container cache.Container, poolHint string) (Grant
 		offer *libmem.Offer
 	)
 
+	if p.cpuClasses != nil {
+		if cc := cpuClassPreference(container); cc != "" {
+			if !p.cpuClasses.IsKnownClass(cc) {
+				return nil, policyError("%s: nonexistent CPU class %q requested",
+					container.PrettyName(), cc)
+			}
```
askervin: Would it be good to log a warning, an error, or even return policyError if a non-Guaranteed container is annotated with cpu-class annotation?

klihub: Logging a warning definitely makes sense. Though we need to do it for containers not eligible for exclusive CPU allocation, which is not quite the same as non-Guaranteed QoS class.

Why it's not that obvious/I'm a bit hesitant on outright erroring out is that with effective annotations a common pattern is to assign a pod-wide default with `.../pod: default-class` and then punch exceptions into it with container-specific further annotations like `.../container.$CTR0: some-class` and `.../container.$CTR1: other-class`. And in that case if you have containers with partial/shared pool CPU allocations, it is normal/an expected side-effect that their CPU class resolves to `default-class`. Though, it would be technically possible to possible to check for this and allow it, but error out in other cases...

pkg/apis/config/v1alpha1/resmgr/policy/topologyaware/config.go
klihub: > and it might be more intuitive than exclusive allocations using sharedCPUClass by default.

Well, we don't do that ( that explicitly) by default, although that is the effective end result (for non-isolated CPUs) if there is no `DefaultExclusiveCpuClass` set and the CPUs were ever part of a shared pool after which they get assigned exclusively to a container which does not have an annotated class.

Maybe we could use a defined `default` class as `DefaultExclusiveCpuClass` if one is not given, and log this fact. Then again, if we did that then there would be 2 ways of setting a default class, and one of them would be quite implicit, not obvious without reading the documentation: does `default` apply by default to all CPUs, shared CPUs only, exclusively allocated CPUs only, or something else ?

cmd/plugins/topology-aware/policy/resources.go
klihub: Yes. This is already rebased on #719.

cmd/plugins/topology-aware/policy/pools.go
klihub: I implemented checking and rejecting non-eligible containers with an explicitly annotated per container cpuclass, which is clearly something that would not take effect.

pkg/apis/config/v1alpha1/resmgr/policy/topologyaware/config.go
klihub: @askervin I don't have too strong opinions about this, but so far I have not implemented this additional implicit semantic interpretation if a `default` CPU class is defined. Mostly because I did not want multiple ways for defining a default exclusive CPU class.

Let's think more about this, and let me know if you still feel it would make sense. Related to this, I thing I'm still considering is to change the pool-related CPU class configuration variables from `{shared,reserved}CPUClass` to `{shared,reserved}PoolCPUClass` to be more explicit in the naming about their semantics...

klihub:
Closing in favor of stacked #742

PR #730, opened 2026-08-05
resmgr: force a cache save after CreateContainer.

klihub:
Force a cache save after CreateContainer, ensuring its state gets updated to Created on disk, too. Failing to do so could leave the last created container in the synthetic Creating state which is ignored during startup. This in turn could cause the last container to not get properly reallocated if the policy gets restarted or crashes.


askervin (approved):
LGTM

PR #734, opened 2026-08-06
cpuclass, balloons: split out common cpuclass validation.

klihub:
Please, review/merge #733 first.

Split out policy agnostic cpuclass validation code from balloons, moving it to config/resmgr/policy.

Note: This PR is stacked on top of #733. It is not github-stacked because (I think) PRs can't be github-stacked if the PR branches are not originating from the same source repository.

PR #735, opened 2026-08-06
cpuclass: add IsKnownClass

klihub:
Please, review/merge #734 first.

Add `IsKnownClass()` to allow policies and other cpuclass infra users to use common code for verifying if all policy-specific pre-configured classes exist.

Note: This PR is stacked on top of #734. It is not github-stacked because (I think) PRs can't be github-stacked if the PR branches are not originating from the same source repository.


klihub:
Closing in favor of stacked #740

PR #736, opened 2026-08-06
cpufreq: always log class enforcement

klihub:
Please, review/merge #735 first.

Always debug log cpufreq class enforcement intent. This helps with log-based verification in e2e tests with repeated class assignments. Other otherwise some assignment of a cpufreq-only class could be omitted from logs if it does not change cpufreq state for a CPU.

Note: This PR is stacked on top of #735. It is not github-stacked because (I think) PRs can't be github-stacked if the PR branches are not originating from the same source repository.


klihub:
Closing in favor of stacked #741

PR #737, opened 2026-08-07
e2e: force RPM installation in vm-kernel-pkgs-install.

klihub:
We might try to install an old kernel compared to the underlying VM distro. Therefore use `--force`, which does a `--replacepkgs --replacefiles --oldpackage`.


askervin (approved):
LGTM

PR #741, opened 2026-08-07
cpufreq: always (debug-)log class enforcement.

klihub:
Always debug log cpufreq class enforcement intent. This helps with log-based verification in e2e tests with repeated class assignments. Otherwise some assignment of a cpufreq-only class could be omitted from logs if it does not change cpufreq state for a CPU.

---


askervin (approved):
LGTM.

fmuyassarov (approved):
LGTM

kad (approved):

PR #742, opened 2026-08-07
topology-aware: add CPU class support.

klihub:
This patch series adds basic CPU class support to the topology-aware policy.

With these patches in place, CPU classes can now be assigned to
- CPUs exclusively assigned to (Guaranteed QoS class) containers,
- CPUs of shared pools, and
- CPUs of the reserved pool

CPU class support is turned on by defining some CPU classes in the configuration and specifying at least the CPU class for shared pool by setting `sharedCPUClass` in the configuration. This is the minimal valid configuration.

Additionally, an optional reserved pool CPU class can be specified for CPUs of the reserved pool by setting `reservedCPUClass` in the configuration. An unset reserved pool class defaults to the shared pool class. Furthermore, also optionally a default CPU class for exclusive CPUs can be specified by setting `defaultExclusiveCPUClass` in the configuration. If set exclusive CPUs of otherwise unnannotated containers are assigned to the default exclusive class.

Containers can also be explicitly annotated with a CPU class using the `cpu-class[.resource-policy.nri.io]` effective annotation base key. Exclusive CPUs of annotated containers are assigned to their annotated class. Once CPUs are released from exclusive use and returned to a shared pool, they are (re-)assigned to the shared pool CPU class.


klihub:
This PR is a stacked recreation of #729. So please see #729 for the previous review comments and related discussion.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -275,6 +277,25 @@ func schedulingClassPreference(ctr cache.Container) (*cfgapi.SchedulingClass, er
 	return sc, err
 }
 
+// cpuClassPreference returns the explicitly annotated or default CPU class
+// for Guaranteed QoS class containers.
+func cpuClassPreference(ctr cache.Container) (string, error) {
+	class, scope, ok := ctr.QueryEffectiveAnnotation(keyCpuClass)
+
+	if ctr.GetQOSClass() == corev1.PodQOSGuaranteed {
```
askervin: I was wondering if reserved (kube-system or annotated to prefer reserved CPUs) Guaranteed containers would need here a separate CPU class preference to ReservedCpuClass. Yet this class is applied to reserved CPU pool at initialization and reserved pods will end up there, as long as there is space, there is still the case where reserved pool is not sufficiently large.

But IIUC, real and wanna-be-reserved containers, Guaranteed or not, will get their CPUs always from a shared pool in that case, so probably there is no need to consider ReservedCpuClass here or in resolveCpuClass after all.

test/e2e/policies.test-suite/topology-aware/n4c16/test19-cpuclass/code.var.sh
```diff
@@ -0,0 +1,307 @@
...
+pod=pod1
+ANN0="cpu-class.resource-policy.nri.io/container.${pod}c0: class1" \
+ANN1="cpu-class.resource-policy.nri.io/container.${pod}c1: class2" \
+    CONTCOUNT=4 create guaranteed
+
+cpu0=$(cpuids-of $pod ${pod}c0)
+assert-cpu-freq ${pod}c0 $cpu0 class1
+cpu1=$(cpuids-of $pod ${pod}c1)
+assert-cpu-freq ${pod}c1 $cpu1 class2
+cpus2=$(cpuids-of $pod ${pod}c2)
```
askervin: cpu2=.... was specified on line 171. After this, both $cpu2 and $cpus2 are defined, and wrong one gets checked on line 211.

I was also thinking if we could make a different size of CPU requests in containers of pod0 and pod1, so that cpusets in the log entries could not accidentally match. In principle, it might reveal slightly larger range of potential cpuclass bugs, too.

test/e2e/policies.test-suite/topology-aware/n4c16/test19-cpuclass/code.var.sh
```diff
@@ -0,0 +1,307 @@
...
+
+vm-command "kubectl -n kube-system get topologyawarepolicies.config.nri/default -ojson | jq '.status.nodes[].errors'"
+
+grep -q 'unknown reserved CPU class \\"nonexistent\\"' <<<$COMMAND_OUTPUT ||
+    error "Missing reserved pool CPU class validation error in configuration CR"
+
+RESERVED_CPUCLASS=reserved
+helm-terminate
+
+# Non-existent reserved pool CPU class
```
askervin: s/reserved/shared

docs/resource-policy/policy/balloons.md
```diff
@@ -867,135 +867,8 @@ IRQs CPU affinity.
...
-  names. Resolves `"turbo"` directly to the hardware maximum
-  turbo frequency, regardless of soft `turboPriority`
-  arbitration.
-- `pctMaxFreq` (string, optional): CLOS maximum frequency,
-  defaults to `maxFreq`. Same caveats as `pctMinFreq`.
-
-On hosts without SST support the PCT fields are ignored with a
-warning, so a single cpuClass YAML can be portable across PCT and
-non-PCT systems.
+See the [CPU class description](common-functionality.md#class-based-cpu-tuning) for CPU class
```
askervin: Yet this is very clean cut and link to a nicely and mostly policy-agnostically worded documentation, I still wish there was more reader-friendly solution that would fulfill two contradicting goals here:
1. A user who configures a policy, a task that is complex enough alone, would not be bothered with the fact that there are other plugins/policies that share common configuration sections.
2. A developer who changes a common feature cannot forget to update all relevant documentations accordingly.

I wanted to have a single, fully self-contained documentation balloons. Therefore I wrote balloons-specific installation and configuration CR(D) sections into this same document, even though they work exactly similarly for topology-aware, and in many parts are common with other nri-plugins, too. I wanted to avoid reader's context switches with the cost of more inconvenient maintenance.

Therefore, I'm hesitating to take a step towards more scattered documentation and reader experience via the link to common parts, if the referenced text could kept here, too... possibly with some notes that happen to be relevant for a balloons user.

askervin (commented):
I'm unsure about this documentation split. Let's discuss offline.

test/e2e/policies.test-suite/topology-aware/n4c16/test19-cpuclass/code.var.sh
klihub: Thanks. Fixed. I also fixed all the wrong configuration keys (sharedPoolCPUClass, reservedPoolCPUClass after a late change vs. the original sharedCPUClass, reservedCPUClass).

test/e2e/policies.test-suite/topology-aware/n4c16/test19-cpuclass/code.var.sh
klihub: Fixed, and updated tests as suggested to use differently sized exclusive CPU sets for different pods' containers.

docs/resource-policy/policy/balloons.md
klihub: I was thinking/hoping that the balloons cookbooks/howtos would/could be the self-contained ones (albeit a bit lighter on the 'why' details and focusing only on the 'how' parts), while in the main documentation we'd try to minimize duplication/multiplication of the same information which then tend to get out of sync.

But if the consensus is to not distill this to a shared section, then I'll update the docs by undoing the split and adding corresponding extra bits to the T-A documentation update. Let's check/chat about this during our sync to see what is the consensus and then I act upon it accordingly.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Yes, we always do shared allocation in the reserved pool... so we could add a bit of extra check (here or later when we get closer to applying classes), to log at least a warning if a reserved pool guaranteed QoS class container has a CPU class annotation, which we then will not obey, since it will run in the reserved pool.

askervin: Good point, CPU class annotation that can't be obeyed is worth a warning. On the other hand, not respecting ReservedCpuClass due to container overflow from reserved to shared pool doesn't sound like that. CPU class is probably not the biggest issue in that situation.

docs/resource-policy/policy/balloons.md
klihub: I reverted this split-out and added it to the topology-aware documentation instead.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: @askervin I added more comprehensive and consistent error checking, with errors now on most code paths causing container creation to fail.

askervin (approved):
LGTM.

Maybe we should take this in well ahead of cutting next release to play around with it and all other changes in the main branch (including soon IRQs).

cmd/plugins/topology-aware/policy/cpu-class.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (p *policy) resolveCpuClass(ctr cache.Container) (string, bool, error) {
+	if p.cpuClasses == nil {
+		return "", false, nil
+	}
+
+	class, isCtrScoped, err := cpuClassPreference(ctr)
+	if err != nil {
+		return "", isCtrScoped, err
```
marquiz: super nit: why return isCtrScoped in the error case?

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -275,6 +277,25 @@ func schedulingClassPreference(ctr cache.Container) (*cfgapi.SchedulingClass, er
...
+
+	switch {
+	case !ok && qos == corev1.PodQOSGuaranteed:
+		return opt.DefaultExclusiveCpuClass, false, nil
+	case ok && qos == corev1.PodQOSGuaranteed:
+		return class, scope == cache.ContainerScopedAnnotation, nil
+	case ok && scope == cache.ContainerScopedAnnotation:
+		return "", true, fmt.Errorf("CPU class (%q) invalid for non-Guaranteed QoS class (%v)",
+			class, qos)
+	}
```
marquiz: Just thinking aloud while reading this. Wonder if some other formatting would make it more obvious which cases/permutations are handled and which fall through to the default


```suggestion
	switch {
	if ok {
		if qos == corev1.PodQOSGuaranteed {
			return class, scope == cache.ContainerScopedAnnotation, nil
		}
		if scope == cache.ContainerScopedAnnotation {
			return "", true, fmt.Errorf("CPU class (%q) invalid for non-Guaranteed QoS class (%v)", class, qos)
		}
	} else {
		if qos == corev1.PodQOSGuaranteed {
			return opt.DefaultExclusiveCpuClass, false, nil
		}
	}
```

marquiz (approved):
Monumental effort @klihub. Two small comments but looks good to be merged

cmd/plugins/topology-aware/policy/cpu-class.go
klihub: No real particular reason at the moment TBH, as it is not used in the caller on the `err != nil` branch. Was just not sure if I'll need it so I made it as transparent as I could. Updated it now to not always return false if `err != nil`, also in the later in the `!IsKnownClass()` branch for consistency.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Updated/rewrote it to a nested if-else construct, and added a clarifying comment.

kad (approved):

PR #744, opened 2026-08-07
config-manager: don't panic if containerd/config.toml is empty

askervin:
Fixes helm install --set nri.runtime.patchConfig=true init container crash loop on systems with no real content in containerd config.toml.


klihub (approved):
LGTM

PR #747, opened 2026-08-10
topology-aware: implement IRQ affinity control.

klihub:
Add basic IRQ affinity control support. Guaranteed QoS class containers with excusive CPU allocation can now be annotated to claim and/or mask interrupts, which can be defined either by an interrupt number or by a shell glob pattern matching 0 or more device or chip names in `/proc/interrupts`.

When a container with exclusive CPU is allocated or released affinities for all interrupts are recalculated. This happens by calculating the union of all exclusive CPUs of containers that claim or mask the interrupt. These are the claiming and resulting CPU sets for the interrupt.

If an interrupt is only claimed, its affinity is updated to the claiming CPU set. If it is only masked, its affinity is updated to the allowed CPU set minus the masked ones. If an interrupt is both claimed and masked by some CPUs, claiming ones win by default. This behavior can be altered by setting the optional IRQ affinity `Mode` to `"claim-first"`, in which case masking ones win if both are present.

---


cmd/plugins/topology-aware/policy/irq-affinity.go
```diff
@@ -0,0 +1,107 @@
...
+func (p *policy) irqCpus(hwIrq *irq.Irq) (preMask, claim, mask cpuset.CPUSet) {
+	preMask, claim, mask = cpuset.New(), cpuset.New(), cpuset.New()
+	for _, g := range p.allocations.grants {
+		if g.GetContainer().GetQOSClass() != corev1.PodQOSGuaranteed {
+			continue
+		}
+		if irqs := g.IrqAffinity(); irqs != nil {
+			for _, c := range irqs.Claim {
+				if hwIrq.Match(c) {
+					claim = claim.Union(g.IsolatedCPUs()).Union(g.ExclusiveCPUs())
```
askervin: Could this be simplified to
```
claim = claim.Union(g.ExclusiveCPUs())
```
because g.IsolatedCPUs() seems to return a subset of ExclusiveCPUs() (that is, those exclusive CPUs that are isolated, too)?

The same applies to preMask.Union and mask.Union below.

cmd/plugins/topology-aware/policy/irq-affinity.go
```diff
@@ -0,0 +1,107 @@
...
+	// IrqAffinityModeDefault is the default IRQ affinity mode.
+	IrqClaimFirst  IrqMode = "claim-first"
+	IrqMaskFirst   IrqMode = "mask-first"
+	IrqModeDefault IrqMode = IrqMaskFirst
+)
+
+func (p *policy) irqCpus(hwIrq *irq.Irq) (preMask, claim, mask cpuset.CPUSet) {
+	preMask, claim, mask = cpuset.New(), cpuset.New(), cpuset.New()
+	for _, g := range p.allocations.grants {
+		if g.GetContainer().GetQOSClass() != corev1.PodQOSGuaranteed {
```
askervin: nit (matter of taste):
As we have here an `if` to skip certain grants, could we similarly skip grants with irqs == nil, instead of having an `if` (line 44) that contains the whole loop?

Regarding this `if`, perhaps skipping grants with empty ExclusiveCPUs() would be more accurate than skipping other than Guaranteed containers. This would skip reserved containers and containers that opted out exclusive CPUs.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -296,6 +298,35 @@ func cpuClassPreference(ctr cache.Container) (string, bool, error) {
...
+	}
+
+	a := &IrqAffinity{}
+	if err := yaml.Unmarshal([]byte(value), a); err != nil {
+		return nil, scope == cache.ContainerScopedAnnotation,
+			fmt.Errorf("invalid IRQ affinity %q: %w", value, err)
+	}
+	if a.Mode == "" {
+		a.Mode = IrqModeDefault
+	}
```
askervin, line 327: Maybe we should handle explicitly here typos and user-invented IRQ modes, erroring out would seem to be in line with how CPU class and cold start typos are handled in newRequest().

cmd/plugins/topology-aware/policy/topology-aware-policy.go
```diff
@@ -111,6 +111,7 @@ func (p *policy) Setup(opts *policyapi.BackendOptions) error {
 	defaultPrio = cfg.DefaultCPUPriority.Value()
 
 	defer p.commitCpuClasses("setup")
+	defer p.applyIrqAffinity("setup")
```
askervin, line 117: Could we support also a situation where admin has configured node's IRQ affinities for her own, non-Kubernetes managed services and applications?

IIUC, currently topology-aware would hijack IRQ management in any case, typically allowing IRQs on every Kubernetes-allowed CPUs, and not on other CPUs.

Maybe p.applyIrqAffinity() could check a policy configuration option, like
- `manageIRQs`, if we want opt-in, or
- `preserveIRQs`, if we want opt-out.

In principle, these could be IRQ match expressions instead of booleans, for extreme flexibility...

I'd perhaps vote for requiring `manageIRQs: "*"` if an admin wants the policy to take care IRQ affinities, because I'm expecting a significant performance hit in resource allocations and releases, if we add hundreds of IRQ affinity procfs reads without a need to do anything.

In balloons, IRQ management is enabled only if policy config includes IRQ claims/modes, too.

klihub: > I'm expecting a significant performance hit in resource allocations and releases, if we add hundreds of IRQ affinity procfs reads without a need to do anything.

Yes, that's something I was also thinking about and one thing I was considering (instead of, but could be also in addition to, administrative enabling/disabling) is to keep track of containers that have IRQ affinity settings, and if we have none skip hwIRQ scanning and affnity setting. WDYT ?

cmd/plugins/topology-aware/policy/irq-affinity.go
klihub: Thanks for spotting it. Updated.

cmd/plugins/topology-aware/policy/irq-affinity.go
klihub: Added extra checks before the inner loops. Also updated checks in inner loops for shallower nesting.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: Split out `IrqAfffinity` parsing as `parseIrqAffinity()` in `irq-affinity.go`, called from `irqAffinityPreference()`. Added mode validation checks. Added e2e tests for mode validation.

cmd/plugins/topology-aware/policy/topology-aware-policy.go
klihub: @askervin I added tracking for containers' "overall IRQ affinity state", and skipping unnecessary updates if no changes have taken place since the last update.

One last thing now would be to add selective administrative lockdown/enabling of IRQ affinity handling, with the `manageIRQs` configuration option you came up with above.

klihub: @askervin I added control over which IRQs can be managed.

cmd/plugins/topology-aware/policy/irq-affinity.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		for _, m := range irqs.Mask {
+			if !hwIrq.Match(m) {
+				continue
+			}
+			if irqs.Mode == IrqClaimFirst {
+				preMask = preMask.Union(g.ExclusiveCPUs())
+				log.Debugf("irq: %s pre-masks %s", g.GetContainer().PrettyName(), hwIrq.String())
+			} else {
+				mask = mask.Union(g.IsolatedCPUs()).Union(g.ExclusiveCPUs())
```
askervin: s/
mask.Union(g.IsolatedCPUs()).Union(g.ExclusiveCPUs())
/
mask.Union(g.ExclusiveCPUs())

pkg/irq/irq.go
```diff
@@ -72,12 +123,51 @@ func Interrupts() ([]*Irq, error) {
...
+	var errs []error
+
+	data, err := os.ReadFile(filepath.Join(proc, "interrupts"))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read interrupts: %w", err)
+	}
+
+	denied := []*Irq{}
+	for line := range strings.SplitSeq(string(data), "\n") {
+		if irq, ok := parseInterruptsLine(line); ok {
```
askervin: Could we have a single /proc/interrupts reader and parser that would be used behind both Interrupts() and ValidateReferencedInterrupts()?

pkg/irq/irq.go
```diff
@@ -72,12 +123,51 @@ func Interrupts() ([]*Irq, error) {
...
+			}
 		}
 	}
 	return irqs, nil
 }
 
+// ValidateReferencedInterrupts takes a set of user interrupt glob
+// patterns and validates them against the denied set of HW IRQs.
+// If any denied IRQ is matched by a pattern and error is returned.
+// This can be used to pre-validate user-facing configuration, such
```
askervin: I'm not very fond of function comments that suggest when to use a function, or even worse, document who uses the function and for what, as those would be a maintenance horror. Even here, "can be used to pre-validate user-facing configuration" starts making assumptions about the context where this module/function is used. I'd drop that sentence.

pkg/irq/irq.go
```diff
@@ -72,12 +123,51 @@ func Interrupts() ([]*Irq, error) {
...
+				irqs = append(irqs, irq)
+			}
 		}
 	}
 	return irqs, nil
 }
 
+// ValidateReferencedInterrupts takes a set of user interrupt glob
+// patterns and validates them against the denied set of HW IRQs.
+// If any denied IRQ is matched by a pattern and error is returned.
```
askervin: s:
// If any denied IRQ is matched by a pattern and error is returned.
:
// Returns denied IRQs referenced by any pattern with denied reference details in errors.

askervin (commented):
A couple of comments.

pkg/irq/irq.go
klihub: Updated.

pkg/irq/irq.go
klihub: Updated removing suggested usage.

pkg/irq/irq.go
klihub: Rewrote all collection functions using a single split out parsing+collection helper `ForEachInterrupt(func(*Irq) error) error).

cmd/plugins/topology-aware/policy/irq-affinity.go
klihub: Updated.

askervin (approved):
LGTM

cmd/plugins/topology-aware/policy/irq-affinity.go
```diff
@@ -0,0 +1,146 @@
...
+			cpus = cpus.Difference(mask)
+		}
+
+		if cpus.Equals(current) {
+			continue
+		}
+
+		if err := hwIrq.SetAffinityCpus(cpus); err != nil {
+			log.Errorf("%s: failed to set affinity to cpus %s (for %s): %v",
+				hwIrq.String(), claim.String(), user, err)
```
kad: Should it be here `cpus.String()` ?

cmd/plugins/topology-aware/policy/cache.go
```diff
@@ -148,6 +148,7 @@ type cachedGrant struct {
 	MemType    memoryType
 	MemSize    int64
 	ColdStart  time.Duration
+	Irqs       *IrqAffinity
```
kad, line 151: if we adding new fields, in json tests, shall it be also updated?

cmd/plugins/topology-aware/policy/irq-affinity.go
klihub: Yes, it should have been. Thanks for spotting. Fixed.

kad (approved):

cmd/plugins/topology-aware/policy/cache.go
klihub: In principle yes. But we've practically disabled the T-A.cache.Unmarshal()ing code path by always clearing any policy-specific data found in the cache on startup. So not that critical now, and a bit unclear to me if we'd rather get rid of saving grants in the cache instead.

PR #748, opened 2026-08-12
topology-aware: don't try to restore cache on startup.

klihub:
Don't try to restore allocations from the cache on startup. All policy specific data is cleared from the cache by resmgr before it is handed over to the policy, so allocations will be empty. Will re-establish them during runtime synchronization.


askervin (approved):
LGTM

PR #750, opened 2026-08-13
helm/balloons: add OpenShift SCC support

kwisniewski98:
Adds optional OpenShift/OKD support to the `nri-resource-policy-balloons` Helm chart via a new `openShift.enabled` value (default `false`).

On OpenShift, pod admission is gated by SecurityContextConstraints (SCC) in addition to RBAC/PodSecurity. The balloons plugin DaemonSet mounts host paths  (`/sys`, `/proc`, `/var/run/nri`, the kubelet pod-resources socket) and runs privileged when `allowPCT` is set. No SCC is assigned to the plugin ServiceAccount by default, so on a stock OpenShift cluster the DaemonSet's pods are rejected at admission and the chart is effectively unusable there.

When `openShift.enabled=true`, the chart grants the plugin ServiceAccount an SCC matched to the mode the pod actually runs in:
  - **`allowPCT=true`** → a namespaced RoleBinding to the built-in `privileged`
    SCC (via the cluster role `system:openshift:scc:privileged`). The pod is
    privileged and mounts host `/dev` for SST/PCT access.
  - **`allowPCT=false`** → a dedicated **least-privilege** SCC that permits only
    container, no privilege escalation, all capabilities dropped, and no host
    namespaces.

Both paths use a namespaced RoleBinding and avoid any cluster-wide grant of `use`.


deployment/helm/balloons/values.yaml
```diff
@@ -176,6 +176,13 @@ allowPCT: false
 # plugin on each new node.
 podPriorityClassNodeCritical: true
 
+# OpenShift / OKD support. When true, grant the plugin ServiceAccount an SCC so
+# its pods pass admission: the built-in `privileged` SCC when allowPCT is set,
+# otherwise a dedicated least-privilege SCC (hostPath, no privilege). Leave false
+# on vanilla Kubernetes.
+openShift:
+  enabled: false
```
klihub: uber-nit: wouldn't something like `openshift-cluster: true` (if we aim for a single setting under which we might collext additional openshift-specific settings in the future) or `openshift.grant-scc: true` (if we want to gather explicitly settable additional openshift-specific settings under `openshift` in the future) be a more understandable name for this setting ?

kwisniewski98: Done

askervin (approved):
LGTM

klihub (approved):
LGTM.

PR #751, opened 2026-08-14
balloons,config,docs: implement interrupt control restriction.

klihub:
Add a `controllableInterrupts` configuration setting which can be used to restrict which interrupts balloons can try to control. If `controllableInterrupts` is set to one or more glob patterns, only interrupts with a matching description in /proc/interrupts can be controlled. If unset all interrupts can be controlled.

---


askervin (commented):
Validate() validates claimed IRQs, but resets irq.SetAllowedInterrupts() after that, which makes perfect sense, as we don't want validation to take the configuration into use yet.

Once we have validated a config and are taking it into use, we should call irq.SetAllowedInterrupts(c.ControllableInterrupts) again, to make sure disallowed interrupts will not be touched?

klihub:
> Validate() validates claimed IRQs, but resets irq.SetAllowedInterrupts() after that, which makes perfect sense, as we don't want validation to take the configuration into use yet.
> 
> Once we have validated a config and are taking it into use, we should call irq.SetAllowedInterrupts(c.ControllableInterrupts) again, to make sure disallowed interrupts will not be touched?

Thanks for spotting this. I forgot, but have added it now.

askervin (approved):
LGTM

pkg/apis/config/v1alpha1/resmgr/policy/balloons/config.go
```diff
@@ -415,12 +427,28 @@ func (c *Config) Validate() error {
 			}
 		}
 	}
+
+	old, err := irq.SetAllowedInterrupts(c.ControllableInterrupts)
+	if err != nil {
+		return fmt.Errorf("invalid controllable interrupts: %w", err)
+	}
+	defer func() {
```
kad: I think there is race here. Validate temporarily sets allowed list to some value, and if something using that list at this moment (another validate or enforcement for starting/stopping container), it would use wrong value. Lock?

klihub: > I think there is race here. Validate temporarily sets allowed list to some value, and if something using that list at this moment (another validate or enforcement for starting/stopping container), it would use wrong value. Lock?

Updated by splitting up functionality a bit differently, which now allows checking referenced allowed interrupts without overwriting then restoring the currently set allowed patterns.

kad (approved):

PR #752, opened 2026-08-14
build: make sure we copy updated CRDs to helm charts.

klihub:
Make sure we copy generated CRDs to Helms charts only after they have been regenerated.


askervin (approved):
LGTM

PR #753, opened 2026-08-17
e2e: always nuke OTEL collector after tests.

klihub:
We run the OTEL collector pod in the `kube-system` namespace, IOW on the reserved CPU set/pool. Normally this would be alone enough to exhaust reserved CPUs. So always clean up the OTEL collector pod, also on test failures, to avoid false positives in subsequent test runs against the same VM failing due to the OTEL collectors unexpected CPU request.


askervin (approved):
LGTM

PR #755, opened 2026-08-18
e2e: wait for OTEL image pull/collector to start in tests.

klihub:
Wait for OTEL collector to start before proceeding with the rest of the tests. On a pristine node this typically takes tens of seconds, since the OTEL collector image needs to be pulled first.


askervin (approved):
LGTM

PR #756, opened 2026-08-18
irq: cache procfs with a write buffer

askervin:
Add procfs read/write caching to the irq package.
- Always cache reads, data cached in parsed format.
- Write caching (buffering) is controlled with BlockWrites() / UnblockWrites(). Buffered writes are synchronously written when all writes are unblocked.
- If a write fails because IRQ affinity cannot be controlled from userspace, don't try writing that affinity again.
- DropCache() forgets everything: what was read, what was write-buffered, what had failed.

### DIFF
--- a/pkg/irq/irq-cache.go
+++ b/pkg/irq/irq-cache.go
@@ -1,0 +1,334 @@
+// Copyright The NRI Plugins Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package irq
+
+// This file implements the caching layer of the irq package. All procfs
+// reads, writes and parsing of this package happen here.
+//
+// Caching assumes that interrupts are neither added nor removed at
+// runtime, and that no entity outside this package alters interrupt
+// affinities. Therefore the interrupts file is read and parsed only
+// once, and the affinity of an interrupt is read only until its first
+// known value.
+
+import (
+	"errors"
+	"fmt"
+	"maps"
+	"os"
+	"path/filepath"
+	"slices"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+
+	"github.com/containers/nri-plugins/pkg/utils/cpuset"
+)
+
+// irqInfo is a numbered interrupt as parsed from the interrupts file,
+// without the allow pattern dependent denied flag.
+type irqInfo struct {
+	num         int
+	description string
+}
+
+// fileOps contains the replaceable file operations of the cache.
+type fileOps struct {
+	readFile  func(string) ([]byte, error)
+	writeFile func(string, []byte, os.FileMode) error
+}
+
+// irqCache caches interrupt data read from procfs and buffers
+// interrupt affinities to be written to procfs.
+type irqCache struct {
+	mu   sync.Mutex
+	proc string  // procfs mountpoint, all paths are built from this
+	ops  fileOps // file operations to use
+
+	info     map[int]irqInfo       // cached numbered interrupts, nil if not read
+	cpus     map[int]cpuset.CPUSet // affinities read from or set through the cache
+	pending  map[int]cpuset.CPUSet // affinities not yet written to procfs, empty if unblocked
+	readOnly map[int]bool          // interrupts whose affinity cannot be written
+
+	writeBlock int // writes to affinities are only buffered while positive
+}
+
+// cache is the interrupt cache of this package.
+var cache = newIrqCache()
+
+// newIrqCache returns a new interrupt cache with real procfs file
+// operations.
+func newIrqCache() *irqCache {
+	return &irqCache{
+		proc: proc,
+		ops: fileOps{
+			readFile:  os.ReadFile,
+			writeFile: os.WriteFile,
+		},
+		cpus:     map[int]cpuset.CPUSet{},
+		pending:  map[int]cpuset.CPUSet{},
+		readOnly: map[int]bool{},
+	}
+}
+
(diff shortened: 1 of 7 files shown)


pkg/irq/irq-cache.go
```diff
@@ -0,0 +1,567 @@
...
+			continue
+		}
+		if firstErr == nil {
+			firstErr = fmt.Errorf("failed to set affinity of irq %d to %q: %w", w.num, w.cpus, err)
+		}
+		switch {
+		case readOnly:
+			log.Warnf("irq %d (%s) affinity is read-only, not setting it again: %v",
+				w.num, c.description(w.num), err)
+		case failures == 1:
```
klihub: My wetware brain considers this a bit asking for trouble and would prefer a `failures > 0` test instead. Even if this happened to be always correct ATM, as it is not immediately obvious (to some human readers, for instance myself) whether this (or rather `a.failures`) always starts with set to zero in this function.

pkg/irq/irq.go
```diff
@@ -20,11 +20,9 @@ package irq
 import (
 	"errors"
 	"fmt"
-	"os"
```
klihub, line 23: There is a pre-existing incorrect comment in `irq.go` (my mistake I think) on `line 40`, which I can't mark, just comment on here as it is not close enough context in this diff:
`s/globally globally disallowed/globally disallowed/`

pkg/irq/irq.go
```diff
@@ -43,7 +41,9 @@ var (
 	ErrDeniedInterrupt = errors.New("denied interrupt")
 )
 
-// SetProcRoot sets the procfs root directory and proc mountpoint.
+// SetProcRoot sets the procfs root directory and proc mountpoint. All
+// data cached from the previous mountpoint is dropped, including
```
klihub, line 51: I think it would be better to provide a more direct way of forcing a rediscovery of interrupts. After all, we might eventually have real cases of hotplugged HW bringing in new interrupts. 

This is now already possible, but IMHO only in a bit too indirect way by re-setting the procfs root with a `irq.SetProcRoot(currentProcRoot)`, even though the cache-dropping side-effect is mentioned in the docstring comments above.

Maybe we could add something along these lines, to have a more punctual, dedicated function to the point:

```golang
// DropCachedInterrupts flushes any cached writes and drops any
// cached interrupt data. This forces a rediscovery of interrupts on
// the next access. If caching is disabled DropCachedInterrupts is a
// no-op.
func DropCachedInterrupts() {
    if enabled, _, _, _ := c.snapshot(); !enabled {
        return
    }
    if err := cache.flushWrites(); err != nil {
        log.Errorf("failed to write buffered interrupt affinities: %v", err)
    }
    c.mu.Lock)(
    c.dropLocked()
    c.mu.Unlock()
}
```

klihub (commented):
@askervin First I thought I won't mention this, but, a bit hypocratically contrary to what I myself usually do, I'll play the devil's advocate for simplicity and mention it, since it was the first thing that stood out for me here.

I wonder if the affinity write coalescing with implicit flushes is really needed and justifiable wrt. the complexity it brings in.

Our usage pattern for this is very transactional. Some event comes, we do some local processing and as part of it we might recalculate and set new affinities for some or all IRQs we control. Once we are done with processing the event, we update a bunch of things, IRQ affinities among them. Then wait for the next event.

Also, we only have a single entity/piece of logic controlling IRQs, never multiple (let alone unbeknown to each other).

This suggests that we would be perfectly fine with explicit flushes/commits only, without any implicit write coalescing and implicit, time(out) based flushes/commits. AFAICT, switching to only explicit flush/commit would also eliminate a large part of the most error prone bits of the code (eg. there would be no need for keeping around a single background write/flush loop, one could always fire off an async write loop on demand triggered by explicit flushes).

That said, I don't have very strong feelings about this and it is definitely nothing even close to a blocker. But it is something I wanted to bring up and ask your opinion about.

pkg/irq/irq-cache.go
```diff
@@ -0,0 +1,567 @@
...
+	c.mu.Unlock()
+
+	if old == enabled {
+		return old
+	}
+
+	if !enabled {
+		// Never lose already buffered writes.
+		if err := c.flushWrites(); err != nil {
+			log.Errorf("failed to write buffered interrupt affinities: %v", err)
```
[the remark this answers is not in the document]
askervin: rewritten, outdated

pkg/irq/irq-cache.go
```diff
@@ -0,0 +1,567 @@
...
+)
+
+// cache is the interrupt cache of this package.
+var cache = newIrqCache()
+
+// newIrqCache returns a new interrupt cache with caching enabled and
+// real procfs file operations.
+func newIrqCache() *irqCache {
+	return &irqCache{
+		enabled: true,
```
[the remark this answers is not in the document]
askervin: This is by design. IRQ change detection and associated cache invalidation are to be done later.

pkg/irq/irq-cache.go
askervin: Rewritten.

pkg/irq/irq.go
askervin: Explicit irq.DropCache() added to reset cache contents.

askervin:
> I wonder if the affinity write coalescing with implicit flushes is really needed and justifiable wrt. the complexity it brings in.

Excellent thing to challenge. I have now rewritten the whole thing without timers, with explicit control, synchronous writing of pending affinity changes.

Topology-aware and balloons policies now co-operate with IRQ write buffering by blocking/unblocking writes. I didn't want to bring that to resmgr level, because not all policies should care about IRQs. And I find it more clear if all IRQ handling is controlled in the same level, that is, the policy.

pkg/irq/irq.go
klihub: This earlier mistake added by me is still present.
```bash
klitkey1-mobl1 pr-756 $ git grep 'globally globally' askervin/5ih-irq-procfs-cache
askervin/5ih-irq-procfs-cache:pkg/irq/irq.go:   // globally globally disallowed interrupts.
```

klihub (changes requested):
One nit about an unfixed eralier comment mistake. Otherwise looks really good to me.

klihub (approved):
LGTM. Thank you for taking care of this !

pkg/irq/irq.go
askervin: Fixed this now, thanks for noticing it and then re-noticing it!

fmuyassarov (approved):
LGTM

PR #761, opened 2026-08-21
topology: add PCIe chain and IRQ info to device hints

kad:
Device topology hints only exposed CPU/NUMA/socket affinity. Add PCIeChain and IRQs fields to Hint so callers can see the PCIe bridge/root hierarchy above a device and which interrupts it uses.

- Add a PCIeHopType enum (PCIeHopBridge, PCIeHopRoot) and a PCIeHop struct for each ancestor in the chain.
- Walk the device's ancestor directories in sysfs to build the chain: a class 0x0604 device is a bridge, and the top-level pci<domain>:<bus> directory is the root.
- Read a device's legacy irq file and its msi_irqs/ entries to collect IRQ numbers.
- Extend Hint.String() to print the new IRQs.
- Update the test-data.tar.gz fixture with a bridge-to-endpoint PCI chain and legacy/MSI IRQ entries; add TestPcieChain and TestDeviceIRQs.


pkg/topology/topology.go
```diff
@@ -173,13 +215,95 @@ func getTopologyHint(sysFSPath string) (*Hint, error) {
...
 	return &hint, nil
 }
 
+var (
+	// pciAddressRe matches a PCI BDF address directory name, e.g.
+	// "0000:00:1c.0".
+	pciAddressRe = regexp.MustCompile(`^[0-9a-f]{4}:[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]$`)
+	// pciRootNameRe matches the sysfs root bridge directory name,
+	// e.g. "pci0000:00".
+	pciRootNameRe = regexp.MustCompile(`^pci[0-9a-f]{4}:[0-9a-f]{2}$`)
```
klihub, line 234: nit: a bit of an overkill to use regexp for so regularly limited strings, but we can live with that.

klihub (approved):

askervin (approved):
LGTM

PR #762, opened 2026-08-21
deps: bump Kubernetes to v0.34 and refresh module dependencies

kad:
Unify the Kubernetes packages on v0.34.11. api, client-go and kubelet were pinned at v0.31.2 while apimachinery and code-generator had already moved to v0.33.1, so the tree was inconsistent. v0.34 maps to Kubernetes 1.34, which is what the latest released OpenShift (4.21) ships, so this stays at that ceiling rather than going past it. controller-runtime goes to v0.22.5 to match.

Update the remaining direct dependencies to their current releases, including containerd/nri v0.12.2, grpc v1.83.1, the OpenTelemetry modules (otel v1.45.0), ginkgo v2.32.1, gomega v1.42.1 and testify v1.12.1.

The OpenTelemetry SDK v1.45 builds its default resource from semconv v1.43.0. The instrumentation code still imported v1.41.0, and merging resources with different schema URLs fails at runtime. Bump the semconv import to v1.43.0 in resource.go and tracing.go to match the SDK.


askervin (approved):
LGTM

PR #764, opened 2026-08-21
go.{mod,sum}: bump NRI deps to latest tagged v0.12.2.

klihub:


kad:
as part of overall dep bump in #762 nri deps were also updated... so this pr is practically subset from there.

klihub:
Closing in favor of #762.

PR #766, opened 2026-08-21
helm/topology-aware: add OpenShift SCC support

klihub:
This is a direct liftover of / shamelessly stolen from the corresponding balloons PR #750/commit de76f63b850d803d26abeccf85893781fb257410, applied to the topology-aware policy which now has CpuClass/PCT support. All credits go to @kwisniewski98. Any potential problems are my fault.

On OpenShift, pod admission is gated by SecurityContextConstraints. The policy plugin DaemonSet mounts host paths (/sys, /proc, /var/run/nri, the kubelet pod-resources socket) and runs privileged when allowPCT is set, but no SCC assigned to the plugin ServiceAccount by default admits that pod, so its pods are rejected on a stock OpenShift cluster.

Add an openshift.grant-scc value (default false, no effect on vanilla Kubernetes) that grants the plugin ServiceAccount an SCC matched to the mode it runs in:

  - allowPCT=true  -> bind to the built-in `privileged` SCC via the cluster role system:openshift:scc:privileged (the pod is privileged and mounts host /dev for SST access).

  - allowPCT=false -> a dedicated least-privilege SCC that permits only what the non-privileged pod needs: hostPath volumes, but no privilege, no privilege escalation, all capabilities dropped, and no host namespaces.

Both cases use a namespaced RoleBinding and avoid any cluster-wide grant of `use`.

In the non-privileged (allowPCT=false) case the pod is otherwise confined as container_t, which cannot write the plugin's hostPath data dir under SELinux. Request the spc_t SELinux type in that path (gated on openshift.grant-scc, so vanilla Kubernetes stays confined) so the pod can function while remaining non-privileged.


askervin (approved):
LGTM.

When verified to work in OpenShift, let's add this to install instructions in our documentation, too. And maybe in external guides like

https://github.com/intel/optimization-zone/blob/main/software/kubernetes/nri-resource-policies/README.md

kad (approved):

PR #767, opened 2026-08-25
topology-aware: add support for claiming device-hinted IRQs.

klihub:
Extend annotated `irq-affinity` to allow matching device IRQs to be claimed from topology-hinted devices. If a container is annotated so, resolve matching hinted device IRQs and claim them. Add new e2e test cases for interrupts claimed via device hints.

Note: This PR needs and is manually stacked on top of #761. Marked as a draft until that one gets merged.

### DIFF
--- a/cmd/plugins/topology-aware/policy/irq-affinity.go
+++ b/cmd/plugins/topology-aware/policy/irq-affinity.go
@@ -16,16 +16,19 @@ package topologyaware
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/containers/nri-plugins/pkg/irq"
+	"github.com/containers/nri-plugins/pkg/topology"
 	"github.com/containers/nri-plugins/pkg/utils/cpuset"
 	"sigs.k8s.io/yaml"
 )
 
 type IrqAffinity struct {
-	Claim []string `json:"claim,omitempty"`
-	Mask  []string `json:"mask,omitempty"`
-	Mode  IrqMode  `json:"mode,omitempty"`
+	Claim   []string `json:"claim,omitempty"`
+	Devices []string `json:"devices,omitempty"`
+	Mask    []string `json:"mask,omitempty"`
+	Mode    IrqMode  `json:"mode,omitempty"`
 }
 
 type IrqMode string
@@ -58,6 +61,49 @@ func parseIrqAffinity(raw []byte) (*IrqAffinity, error) {
 	}
 
 	return parsed, nil
+}
+
+func addIrqAffinityForHints(a *IrqAffinity, hints topology.Hints) error {
+	if len(hints) == 0 || a == nil || len(a.Devices) == 0 {
+		return nil
+	}
+
+	if err := irq.ValidateAllowedPatterns(a.Devices); err != nil {
+		return fmt.Errorf("invalid IRQ affinity devices pattern: %w", err)
+	}
+
+	for source, h := range hints {
+		for _, num := range h.IRQs {
+			irq, err := irq.Interrupt(num)
+			switch {
+			case err != nil:
+				log.Errorf("irq: skipping %s-hinted IRQ %d: %v", source, num, err)
+				continue
+			case !irq.IsAllowed():
+				log.Warnf("irq: skipping denied %s-hinted IRQ %d", source, num)
+				continue
+			}
+
+			matched := false
+			for _, p := range a.Devices {
+				if irq.Match(p) {
+					matched = true
+					break
+				}
+			}
+			if !matched {
+				log.Debugf("irq: skipping unmatched %s-hinted IRQ %d (%q)",
+					source, num, irq.Description())
+				continue
+			}
+
+			log.Infof("irq: claim matching %s-hinted IRQ %d (%q)",
+				source, num, irq.Description())
+			a.Claim = append(a.Claim, strconv.Itoa(irq.Num()))
+		}
+	}
+
+	return nil
 }
 
 func (p *policy) irqCpus(hwIrq *irq.Irq) (preMask, claim, mask cpuset.CPUSet) {
--- a/cmd/plugins/topology-aware/policy/pod-preferences.go
+++ b/cmd/plugins/topology-aware/policy/pod-preferences.go
@@ -330,6 +330,10 @@ func irqAffinityPreference(ctr cache.Container) (*IrqAffinity, bool, error) {
 
 	switch {
 	case qos == corev1.PodQOSGuaranteed:
+		err := addIrqAffinityForHints(a, ctr.GetTopologyHints())
+		if err != nil {
+			return nil, scope == cache.ContainerScopedAnnotation, err
+		}
 		return a, scope == cache.ContainerScopedAnnotation, nil
 	case scope == cache.ContainerScopedAnnotation:
 		return nil, true, fmt.Errorf("invalid IRQ affinity, QoS class %v is not Guaranteed", qos)
--- a/docs/resource-policy/policy/topology-aware.md
+++ b/docs/resource-policy/policy/topology-aware.md
@@ -739,7 +739,7 @@ Containers eligible for exclusive CPU allocation can be annotated with IRQ
 tuning to claim selected IRQs, mask selected IRQs, or do both, using the
 `irq-affinity.resource-policy.nri.io` annotation key.
 
-The annotation has 3 fields:
+The annotation has 4 fields:
 **`claim`** (list of strings)
 - Lists IRQs handled by the exclusive CPUs allocated to the container.
 - Each item refers to IRQs either by an exact number, or by a pattern
@@ -747,6 +747,9 @@ The annotation has 3 fields:
   instance "*nvme*".
 - The affinity of a claimed IRQ is set to the union of CPUs of all
   containers that claim it.
+**`devices`** (list of strings)
+- The same as `claim`, but listed IRQs match only devices assigned to
+  the container instead of all devices in the system.
 **`mask`** (list of strings)
 - Lists IRQs which should not be handled by the exclusive CPUs allocated
   to the container.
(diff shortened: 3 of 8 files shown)


cmd/plugins/topology-aware/policy/irq-affinity.go
```diff
@@ -60,6 +61,37 @@ func parseIrqAffinity(raw []byte) (*IrqAffinity, error) {
...
+				log.Errorf("irq: skipping %s-hinted IRQ %d: %w", source, num, err)
+				continue
+			case irq == nil:
+				log.Errorf("irq: skipping unresolved %s-hinted IRQ %d", source, num)
+				continue
+			case !irq.IsAllowed():
+				log.Errorf("irq: skipping denied %shinted IRQ %d", source, num)
+				continue
+			}
+			// TODO(klihub): maybe update irqCpus() to support claiming by
```
askervin: It would be already possible to append strconv.Itoa(irq.num) instead of irq.Description. The first check in irq.Match(pattern), after emptiness of the pattern, is if the pattern equals string representation of irq.num. It ain't pretty, string representation of irq.num is not even cached, but still a shortcut to matching numbers.

cmd/plugins/topology-aware/policy/pod-preferences.go
```diff
@@ -318,7 +318,7 @@ func irqAffinityPreference(ctr cache.Container) (*IrqAffinity, bool, error) {
 	value, scope, ok := ctr.QueryEffectiveAnnotation(keyIrqAffinity)
 
 	if !ok {
-		return nil, false, nil
+		return irqAffinityFromHints(ctr.GetTopologyHints())
```
askervin: I'd vote for an opt-in annotation for claiming IRQs from hinted devices, because:
- default installation (no privileges) would start producing errors in logs on IRQ affinity setting failures.
- so far IRQs have been completely unmanaged by default (in certain environments this might be a disruptive/breaking change in the default behavior)
- I'm not sure if automatic claiming of all IRQs from container's assigned devices is a good idea. Consider a node with single container requesting devices present and single CPU, that single CPU becomes the sink of all IRQs from requested devices. That could cause, not only a lot of interruptions to container's execution, but also risk delaying IRQ handling in case of threaded IRQs and container using a real-time scheduling policy.

Always balancing (not touching) IRQs would be a safe default, but if a user wants something fancy but potentially risky, she could opt-in for that.

So perhaps the topology-aware policy could support "claimFromTopologyHints" (as you suggested offline) or "claimFromAssignedDevices" or something similar, in addition to "claim" in the irq-affinity annotation. For example,

```
irq-affinity.resource-policy.nri.io/container.data-pump: |
      claimFromAssignedDevices:
        - "*eth*"
      mask:
        - "*"
```

would claim IRQ from the NIC assigned to the container, and mask IRQs of other assigned devices (nvme, for instance) and non-assigned NICs.

klihub: > I'd vote for an opt-in annotation for claiming IRQs from hinted devices, because:
> 
> * default installation (no privileges) would start producing errors in logs on IRQ affinity setting failures.
> * so far IRQs have been completely unmanaged by default (in certain environments this might be a disruptive/breaking change in the default behavior)
> * I'm not sure if automatic claiming of all IRQs from container's assigned devices is a good idea. 

@askervin Yeah, I tend to think it is a bad idea. That's why I put that comment in the description above, to convey my feelings for reviewers. Topology hints (alone) are just way too implicit, and can have surprising effects, if applied without any control from the user. 

...

> 
> ```
> irq-affinity.resource-policy.nri.io/container.data-pump: |
>       claimFromAssignedDevices:
>         - "*eth*"
>       mask:
>         - "*"
> ```

Yes, we could do something like that. Or at least have an annotation to indicate that yes, we want to set IRQ affnity from device hints. But I was also considering if we could use the already existing allow-/deny-listing annotations to decide which device hints to take IRQs from... If you have a container where you do want placement based on device hints, you typically annotate an allow or deny list to filter out the unwanted ones, if they tend to show up. Otherwise you will not get the placement you'd be happy with. So then, we could use the fact that a hint allowed/taken into account to decide which interrupts to set affinity for, if we are to for any.

Then again, this might be too limited, or actually too implicit again, as sometimes you might want to do placement only based on the hint for one device, but set up affinity for the interrupts of multiple devices to the CPUs you get placed on. 

I think the biggest shortcoming with this (whole IRQ affinity setting) is that AFAICT (resolving) i(n)t(errupts) does not work for podresourceapi-based hints and devices, although that exactly would be one of the most useful things. IOW, there is no (obvious) way (that I'd be aware of) for getting from a podresourceapi based hint to a device and from that to a set of IRQs.

cmd/plugins/topology-aware/policy/irq-affinity.go
klihub: Yes, the bits in `pkg/irq` I know are in the right shape already. I was just not sure if the additions to the topology-aware policy itself are up to it or still subpar. But it looks like they should, so I update those bits already.

cmd/plugins/topology-aware/policy/pod-preferences.go
klihub: > So perhaps the topology-aware policy could support "claimFromTopologyHints" (as you suggested offline) or "claimFromAssignedDevices" or something similar, in addition to "claim" in the irq-affinity annotation. For example,

Okay, I went ahead and added simply a `devices` ([]string) to annotated claims, which should behave semantically as `claimFromAssignedDevices` below.

> ```
> irq-affinity.resource-policy.nri.io/container.data-pump: |
>       claimFromAssignedDevices:
>         - "*eth*"
>       mask:
>         - "*"
> ```

askervin: `devices` is much nicer.

test/e2e/policies.test-suite/topology-aware/n4c16/test25-irq/code.var.sh
```diff
@@ -375,3 +375,61 @@ grep -q "denied interrupt: .* denied but matched by user pattern .*" <<< $COMMAN
...
+
+verify-irq-cpus ".*ttyS0.*" "$(ctr-cpu-ids $pod ${pod}c0)"
+
+vm-command "kubectl delete pod $pod"
+
+# Create Guaranteed pod annotated to take IRQ affinity for hinted devices,
+# and with a test topology-hint with an IRQ for a denied rtc0. Shouldn't
+# claim IRQ.
+
+pod=pod10
```
askervin, line 432: Value of $pod is not used after assigning pod10 to it.

Maybe it would make sense to use it in annotation: `container.${pod}c0`, and do the same instead of `pod9c0` above, too.

askervin (commented):
Looks already good. One nit.

docs/resource-policy/policy/topology-aware.md
```diff
@@ -747,6 +747,10 @@ The annotation has 3 fields:
   instance "*nvme*".
 - The affinity of a claimed IRQ is set to the union of CPUs of all
   containers that claim it.
+**`devices`** (list of strings)
+- List IRQs of devices from topology hints which should be handled by
```
askervin: Could we formulate `devices` documentation so that it refers even more directly to "claim"?
That would help also understanding why "claim-first" and "mask-first" are enough, and there is no need for "devices-first".

Perhaps something like:
- The same as `claim`, but listed IRQs match only devices assigned to the container instead of all devices in the system.

askervin (commented):
Some thoughts still about the documentation line... Not sure if this improves, but somehow I'd wish that the reader gets that "devices" is only a scoped version of "claim".

test/e2e/policies.test-suite/topology-aware/n4c16/test25-irq/code.var.sh
klihub: Thanks ! Fixed, and also updated/fixed all existing/other test25-irq cases to
- use defined `${pod}` to get annotation keys right
- clean up [un]used `ANN*` from the environment

docs/resource-policy/policy/topology-aware.md
klihub: Updated.

askervin (approved):
LGTM.

Thanks for productizing IRQ support v2!

docs/resource-policy/policy/topology-aware.md
kad: 4 fields, as we are adding devices.

cmd/plugins/topology-aware/policy/irq-affinity.go
```diff
@@ -60,6 +63,45 @@ func parseIrqAffinity(raw []byte) (*IrqAffinity, error) {
...
+				continue
+			}
+
+			log.Debugf("irq: claim matching %s-hinted IRQ %d (%q)",
+				source, num, irq.Description())
+			a.Claim = append(a.Claim, strconv.Itoa(irq.Num()))
+		}
+	}
+
+	return nil
```
kad, line 106: we have two returns, and both of them always return `nil`, some error conditions checking forgotten?

cmd/plugins/topology-aware/policy/irq-affinity.go
```diff
@@ -60,6 +63,45 @@ func parseIrqAffinity(raw []byte) (*IrqAffinity, error) {
...
+				log.Errorf("irq: skipping %s-hinted IRQ %d: %v", source, num, err)
+				continue
+			case !irq.IsAllowed():
+				log.Warnf("irq: skipping denied %s-hinted IRQ %d", source, num)
+				continue
+			}
+
+			matched := false
+			for _, p := range a.Devices {
+				if irq.Match(p) {
```
kad, line 89: do we have proper validation of glob in Devices here? will it just not match or should be raising error?

klihub: No, we lacked it. Added missing validation.

cmd/plugins/topology-aware/policy/irq-affinity.go
klihub: Pattern validation was missing. Added it and it has now an error returning branch.

kad (approved):
LGTM

PR #768, opened 2026-08-26
e2e: distill common test functionality to `lib/test.bash`.

klihub:
This patch set refactors existing test cases by
- looking for common or very similar functionality,
- distilling them out to a new shared shell scriptlet library `lib/test.bash`,
- reworking affected test cases to use the newly added test function

The PR is split to commits to "identify one refactorable function or functions related to the same test or same functionality to be tested, distill out common functionality, then refactor the affected tests to use the new common functions, putting each such change to a separate commit". Moreover, separate sweeps over the `n4c16`, `n4c128`, `n6-hbm-cxl`, and `s8c4k` topology test sets are split to separate commits.

Here is the summary of logical changes made.

- add lib/test.bash test helper library
- add retry-until polling helper
- add pod and namespace cleanup helpers, use in balloons tests
- use the cleanup helpers in topology-aware tests
- add expect-launch-failure helper
- add wait-pod-gone helper
- add container state and creation error helpers
- add plugin log retrieval helpers
- add plugin log assertion helpers
- add node extended resource helpers
- add CPU list arithmetic helpers
- add container cpuset lookup helpers
- add interrupt affinity test helpers
- add scheduling class verification helper
- add policy configuration status helpers
- add helm-reconfigure and patch-policy-config
- add CLOS and CPU frequency assertion helpers
- add node reboot, kernel command line and NUMA helpers
- move the node resource topology helpers to the shared library
- move the metrics helpers to the shared library
- fix pod names in test07-maxballoons error messages
- add relaunch-policy helper
- document the shared test helpers
- drop the unused cleanup function of test17-cstates-scheduling
- use the cleanup helpers in the n6-hbm-cxl tests
- use the shared NRT helpers in the n6-hbm-cxl tests
- distil the memory type verifications of the n6-hbm-cxl tests
- add CPU hot-plug helpers
- add kernel version and kubepods cpuset check helpers
- use the existing shared helpers in the s8c4k tests
- use the shared helpers in the n4c128 test
- distil the repeated loops of the n4c128 test
- stop deleting namespaces the n4c128 test never creates

 The full test set still passes after these changes.

PR #769, opened 2026-08-26
e2e: fast VM bootstrap using cached preprovisioned VM images ('boxes').

klihub:
Note: this PR depends and is manually stacked on #768. Review that first.

Use vagrant's `vagrant package` support to dump, cache, and reuse provisioned VM images. Image names contain the runtime, the kubernetes version, the helm version, CNI version, and the playbook SHA1 for uniqueness and proper avoidance of reusing images when they should not be. 

Use `e2e_vm_cache=yes` to create and reuse cached images.  Use `e2e_vm_cache=drop` for cleaning up any cached boxes.


test/e2e/lib/vm.bash
```diff
@@ -117,6 +139,207 @@ vm-load-cached-var() {
...
+    #
+    # Print a hash of everything which shapes the provisioned guest, so that
+    # editing any of it invalidates the cached boxes.
+    local e2e_dir="$nri_resource_policy_src/test/e2e"
+
+    ( cd "$e2e_dir" &&
+          find playbook -type f | sort | xargs cat &&
+          cat files/Vagrantfile.in files/env.in files/containerd-nri-enable \
+              files/crio-nri-enable files/10-bridge.conf.in
+    ) 2>/dev/null | sha256sum | cut -c1-12
```
kad: will this hide errors if filepath is broken or some broken symlinks?

test/e2e/lib/vm.bash
```diff
@@ -117,6 +139,207 @@ vm-load-cached-var() {
...
+    local vmname="$1" cri_release key
+
+    case "$k8scri" in
+        crio) cri_release="$crio_release";;
+        *)    cri_release="$containerd_release";;
+    esac
+
+    key="$vmname-k8s$k8s_release-$k8scri$cri_release"
+    key="$key-cni$cni_plugin$cni_release-helm$helm_release"
+    key="$key-$(vm-provisioning-recipe-hash)"
```
kad, line 274: what about  custom kernel source or config? will the keys collide?

klihub: If I understand your question correctly, then I think not. We don't do custom kernels like that. We do custom kernels by bootstrapping the VM first with a stock kernel, then we go in, put there the custom kernel sources, compile and install the custom kernel, and finally reboot into it. IOW, it is not part of the VM setup phase, instead it is part of the test case setup phase.

test/e2e/lib/vm.bash
klihub: Hmm, not sure. It might at this point (this is just calculating the name for the snaphotted VM box/image to save or try to bootstrap from). But some of the file paths are broken or are broken symlinks, then boostrapping will fail later when we really try to use those files.

One thing that probably would make sense is to spell out the that filename list in a common shell variable, then use that variable her for the checksumming. And then before checksumming, we could also check that each and every one of those can be read, and fail if any of them cannot be.

klihub: Updated, like above. Spelled out checksummed file list into a shell var. Added an explicit check to make sure each file is readable.

klihub:
@askervin @kad @fmuyassarov There were a couple of problems still lurking there (failure of the 2nd test ran after a VM has been bootstrapped from a cached/stored image snapshots), but those should be now fixed. I reran all e2e tests, and all tests passed on the first run (when the snapshot was saved) and all tests passed on the second run, when the VMs were recreated from the saved snapshots. So this is ready for review now from my point of view.

kad (approved):
LGTM

askervin (approved):
LGTM

PR #774, opened 2026-09-03
Add DRA support to the topology-aware policy

bart0sh:
This PR adds optional DRA integration to the topology-aware policy, letting workloads request high-priority `PCT` (Performance/Turbo Class) CPUs via a `Kubernetes` `ResourceClaim` instead of (or alongside) the existing `NRI`-based allocation path.

**What's included**
- A reusable `Kubernetes` client wrapper (`pkg/kubernetes/client`) and a generic `DRA` `kubelet-plugin` skeleton (`pkg/resmgr/dra`), independent of any specific policy.
- `cpuClass.dra.publish` config field, plus `ValidateCPUClassesForDRA` guarding against overcommitting a turbo-frequency tier when multiple classes are published.
`pct.PickHpCpus/ReleaseHpCpus/Punits`, giving the `DRA` driver a way to allocate and release high-priority CPUs from the existing `PCT/SST-TF` punit accounting.
- A `DRA` device builder (`cpuclass.dra.go`) that publishes one device per high-priority `cpuClass` × `SST-TF` punit, exposing `nri/pctPriority`, `nri/packageID`, `nri/punitID` as selectable device attributes.
- Full kubelet-plugin lifecycle (registration, `PrepareResourceClaims/UnprepareResourceClaims`, `CDI` spec writing, `claim-state` persistence and restart recovery).
- Wiring into the topology-aware policy: `Setup/Start/Stop/Reconfigure` integration, pool-supply accounting for claimed CPUs (multi-container claims via a refcount), and safe refusal of dra.enabled flips or `DRA`-attribute changes on classes with live claims.
- Helm chart additions: `config.dra.{enabled,sharedCounters}` values, `RBAC`, host mounts, and the base `DeviceClass`.
- An e2e test (`test20-dra`) covering claim -> pod -> `CLOS`-association → scheduler-accounting → cleanup, with a feature-gate probe that skips cleanly on clusters without the `KEP-5075/KEP-5517` gates.

**Requirements**
- `Kubernetes 1.34+` for [KEP-5075](https://github.com/kubernetes/enhancements/issues/5075) (`DRAConsumableCapacity`) — required.
- `Kubernetes 1.37+` for [`KEP-5517`](https://github.com/kubernetes/enhancements/issues/5517) (`DRANodeAllocatableResources`) — optional; without it `DRA` allocation still works but scheduler-side capacity accounting is skipped.

**Known v1 limitations (by design, not oversights)**
- Only high-priority `PCT` classes are published as `DRA` devices; non-HP classes and CEL selectors targeting them are rejected at Prepare time.
`config.dra.sharedCounters` is accepted by the schema but rejected at validation time — it's reserved for ([KEP-5941](https://github.com/kubernetes/enhancements/issues/5941)) shared-counter publication, which isn't implemented yet.
- A `DRA`-claimed `CPU` can still collide with a `CPU` already exclusively granted to a reserved-type container (updateSharedAllocations doesn't re-pin/evict reserved grants); tracked as a follow-up.
`dra.enabled`: true with local-file (non-Kubernetes) configuration degrades to `DRA`-disabled for the process lifetime rather than retrying, since the agent skips kube-client setup in that mode.
- Live `Reconfigure()` support for `DRA` config changes (validated `cpuClass` edits, live-claim-aware refusal, PostReconfigure-deferred republish) was implemented and then deferred to a future PR — not because it doesn't work, but to keep this PR's diff reviewable. Right now, any config change while `DRA` is (or was) enabled is refused outright with a restart-required error.

**Testing**
- Unit tests across all new/touched packages (go test ./... — 27 suites, all passing).
- `test20-dra `e2e test, run against a live `Kubernetes 1.37` cluster with both feature gates enabled.


test/e2e/playbook/provision.yaml
```diff
@@ -501,9 +507,64 @@
...
+          featureGates:
+          {% for gate in k8s_feature_gates.split(',') if gate.strip() and '=' in gate %}
+            {{ gate.split('=')[0] }}: {{ gate.split('=')[1] }}
+          {% endfor %}
+          {% endif %}
+          {% if k8s_log_verbosity != "" %}
+          logging:
+            verbosity: {{ k8s_log_verbosity }}
+          {% endif %}
+      when: k8s_feature_gates != "" or k8s_log_verbosity != ""
```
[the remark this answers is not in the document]
bart0sh: This is only valid for outdated  kubeadm API versions, which are not supported anymore by Kubernetes. Kubernetes lowest supported version is 1.34 as far as I know.

fmuyassarov (commented):
@bart0sh have you tested it out? The resourceSlice is created but empty.

bart0sh:
@fmuyassarov 
>  have you tested it out?

I have. I'm running e2e DRA test for every change. 

> The resourceSlice is created but empty.

Thanks for reporting, I'll look at it.

bart0sh:
@fmuyassarov
here is what I see in the e2e test ouput:
...
root@vm topology-aware/test20-dra> kubectl get resourceslices -o json |             jq -c '[.items[] | select(.spec.driver == "nri.topology-aware.cpu") | (.spec.devices // [])[] | {name, capacity: .capacity["nri/cpus"].value, class: .attributes["nri/cpuClass"].string, priority: .attributes["nri/pctPriority"].string, allowMultipleAllocations, nodeAllocatableResources}]'
[{"name":"hp-turbo-pkg0-punit0","capacity":"2","class":"hp-turbo","priority":"high","allowMultipleAllocations":true,"nodeAllocatableResources":{"cpu":{"mapping":{"capacityKey":"nri/cpus","capacityMultiplier":"1"}}}},{"name":"hp-turbo-pkg1-punit0","capacity":"2","class":"hp-turbo","priority":"high","allowMultipleAllocations":true,"nodeAllocatableResources":{"cpu":{"mapping":{"capacityKey":"nri/cpus","capacityMultiplier":"1"}}}}]
...

fmuyassarov:
Let me double check. I'm sure the gate was open during helm otherwise I wouldn't even see the resourceSlice creation. Maybe something I messed up. Will update soon

pkg/resmgr/cpuclass/internal/pct/pct.go
```diff
@@ -520,6 +525,164 @@ func (a *Allocator) FreeClassCapacity(className string, held cpuset.CPUSet) int
...
+	}
+	if avail.Size() < n || hpRoom < n {
+		available := avail.Size()
+		if hpRoom < available {
+			available = hpRoom
+		}
+		return cpuset.New(), fmt.Errorf("pct: PickHpCpus: punit (pkg=%d, punit=%d) has %d available HP CPUs (room=%d, free=%d), need %d",
+			pkgID, punitID, available, hpRoom, avail.Size(), n)
+	}
+	// Sort for deterministic selection; take first n.
```
klihub, line 654: If this eventually becomes the CPU set a container is pinned to and runs on, then the picking logic should be CPU-topology aware. This is easiest done today with cpuallocator.Allocator, which both the topology-aware and balloons policy use also in this way: first filter available CPUs to an allowed/bounding set based on some filtering criteria (here HP/PCT CPUs), then use the allocator, so that if we have to and can pick CPUs worth one or more CPU-topological zone (full core, LLC-cluster, explicitly advertised cluster, shared NUMA node, die, socket), then we do try to pick in multiple chunks of those.

I would not use the cpuallocator directly here in all the places where need to pick CPUs... just to better prepare for potential changes to the cpuallocator, leaking through its API, which we might soon get sincesome of its offered functionality, such as picking CPUs by SST-based CPU classification, has been obsoleted and superseded by the this/the cpuclass package. So I would wrap it in some `pickCpus(from cpuset.CPUSet) (picked cpuset.CPUSet, error)` convenience function here in a single code site, an use that instead.

This might be (but I'm not sure) a moot comment right now from purely a technical point of view today, for instance if PCT-HP CPU cores are always bound to be limited in practice well below any other topological entity size than a full physical core/two hyperthreads. Nevertheless, I still think we should bake in more general topology-aware picking already from the beginning.

pkg/resmgr/cpuclass/internal/pct/pct.go
```diff
@@ -520,6 +525,164 @@ func (a *Allocator) FreeClassCapacity(className string, held cpuset.CPUSet) int
...
+		a.hpDRAUsed = map[int]cpuset.CPUSet{}
+	}
+	a.hpDRAUsed[idx] = a.hpDRAUsed[idx].Union(picked)
+	return picked, nil
+}
+
+// ReleaseHpCpus removes cpus from hpDRAUsed[punitIdx] for the punit
+// identified by (pkgID, punitID). Silently ignores unknown punits and
+// CPUs not present in hpDRAUsed (idempotent).
+func (a *Allocator) ReleaseHpCpus(pkgID, punitID int, cpus cpuset.CPUSet) {
```
klihub, line 667: Same comment applies here as for picking CPUs. We should do it so that the remaining unreleased subset is picked in a CPU topology-aware fashion. Again, that's the easiest done today using the cpuallocator.

pkg/resmgr/cpuclass/internal/pct/pct.go
```diff
@@ -606,7 +606,7 @@ func (a *Allocator) punitNonHPCapacity(idx int) int {
 	if a.allowed.Size() > 0 {
 		cpus = cpus.Intersection(a.allowed)
 	}
-	n := cpus.Size() - pu.GuaranteedHpCpus
+	n := cpus.Size() - a.punitHPCapacity(idx)
```
klihub, line 609: Shouldn't this more naturally belong to/logically fit better to the previous commit (pct: add PickHpCpus, ReleaseHpCpus, and Punits) ? This is fixing PCT bookkeeping internals, which is necessary due to the (IIUC second external allocation channel) changes put in  by the previous commit.

pkg/resmgr/cpuclass/internal/pct/pct.go
```diff
@@ -638,7 +638,8 @@ func (a *Allocator) PickHpCpus(pkgID, punitID, n int, held cpuset.CPUSet) (cpuse
 	avail = avail.Difference(held).Difference(a.hpUsed[idx]).Difference(a.hpDRAUsed[idx])
 	// Also enforce the GuaranteedHpCpus cap: can't pick more HP CPUs than
 	// the punit guarantees, regardless of how many are physically free.
-	hpAlreadyHeld := a.hpUsed[idx].Size() + a.hpDRAUsed[idx].Size()
+	// Use Union.Size() to avoid double-counting any CPU in both sets.
+	hpAlreadyHeld := a.hpUsed[idx].Union(a.hpDRAUsed[idx]).Size()
```
klihub, line 642: ditto for most of the rest of the changes made here to cpuclass/PCT: it looks to me as they'd better fit logically together with the previous commit which introduced `{Pick,Release}HpCpus()`.

cmd/plugins/topology-aware/policy/dra.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	policyapi "github.com/containers/nri-plugins/pkg/resmgr/policy"
+
+	"github.com/containers/nri-plugins/pkg/resmgr/cpuclass"
+	"github.com/containers/nri-plugins/pkg/resmgr/dra"
+	"github.com/containers/nri-plugins/pkg/utils/cpuset"
+)
+
+// buildDRAPlugin constructs p.draPlugin from the current policy
```
klihub, line 25: This looks a bit like we'd have put essentially the whole full blown DRA driver, running almost fully contained within the policy implementation, as an completely independent second driver, the other allocation path (via the stock resmgr NRI layer) being otherwise oblivious to it than via the shared (resmgr-level) lock. That's pretty implicit awareness observable only via more variance in the NRI-triggered code paths timings (usually not visible probably), and potentially via deadlocking which is now more easy to achieve with locking responsibility to protect against the NRI code paths having been fully delegated to the policy-contained DRA driver and it's related logic/additions.

Is there a viable alternative to this, where we'd have a delegation of responsibilities with closer resemblance to how we do it on the NRI-triggered allocation/release paths ? There we basically have the (resmgr) infrastructure/machinery take care of most boilerplate, including most (or maybe all, I don't recall now by heart) of the locking/serialization. The generic resmgr/policy interface has then functions for allocation/reallocation/release and those are used to hook in the policy-specific allocation/release/CPU,memory,etc selection logic into the generic/boilerplate (NRI) driver at the resmgr level.
 
So going more in that direction would mean the resmgr/policy interface gaining extra functions specific to/for the DRA-based allocation/release code paths, which then would be used to hook the policy-specific DRA decision/allocation logic into the generic/boilerplate DRA machinery, which for instance could live at the resmgr level.

I admit that such a split is easier for the traditional allocation, resulting in a rather small and simple API footprint, which is one of the key reasons why such an abstraction suggested itself as an obvious choice. This might not be the case here, or at least not such a clear cut, due to (a) DRA (device) being more versatile in this regard via device names+attributes, etc. But I think we should give it a serious and honest consideration, including thinking about the next steps of 1) exposing something more via DRA from the T-A policy, maybe for instance isolated CPU cores, P/E-cores, etc., and 2) adding DRA support to the balloons policy. Then think about what are the consequences and see what kind of compromises we'd need to make with one and the other approach, so we could take an educated guess which one looks to be less painful in the future.

askervin: Perhaps one way to drive claude towards more policy-independent DRA logic in resmgr could be asking a structure where, in addition to the topology-aware policy, the [template policy](https://github.com/containers/nri-plugins/blob/main/cmd/plugins/template/policy/template-policy.go) would have expose all allowed CPUs in a DRA resource slice, too.

The purpose of the template policy is to provide minimal basis for anyone to implement their own policy. This would extended it to enable anyone building their own policy with DRA support, and as a side effect, would hopefully force claude to design how DRA becomes visible to multiple policies.

bart0sh: That's a good idea! I'll try to add it to the implementation. So far the plan is to cancel this PR and continue sumbitting series of smaller and hopefully easier to review PRs. The series has been already started by https://github.com/containers/nri-plugins/pull/782. Will submit a second one soon.

PR #775, opened 2026-09-04
cache: don't hand-quote cpuset.CPUSet.String()'s.

klihub:
JSON marshal cpuset.CPUSet.String()'s instead of hand-quoting to get rid of a CodeQL alert. Also, do boundary checks on the data before trying to unmarshal.

### DIFF
--- a/pkg/resmgr/cache/cache.go
+++ b/pkg/resmgr/cache/cache.go
@@ -929,7 +929,7 @@ func (cch *cache) GetPolicyEntry(key string, ptr any) bool {
 func marshalEntry(obj any) ([]byte, error) {
 	switch obj := obj.(type) {
 	case cpuset.CPUSet:
-		return []byte("\"" + obj.String() + "\""), nil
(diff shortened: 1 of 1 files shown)


askervin (commented):
Yet there are good chances we'll never see the error from cpuset.Parse(), I still think copilot's suggestion on wrapping it in cacheError is a good idea. It would also be consistent with how cpuset.Parse() error is handled below when unmarshalling into map[string]cpuset.CPUSet.

klihub:
> Yet there are good chances we'll never see the error from cpuset.Parse(), I still think copilot's suggestion on wrapping it in cacheError is a good idea. It would also be consistent with how cpuset.Parse() error is handled below when unmarshalling into map[string]cpuset.CPUSet.

@askervin True, and it is also then more consistent with any previous string unmarshalling errors. Updated accordingly.

fmuyassarov (approved):

PR #778, opened 2026-09-09
sysfs: return untyped nil for hardware which is not there.

klihub:
System.CPU, System.Node and System.Package returned a nil pointer inside a non-nil interface for an id the machine does not have. Such a value is never == nil, so callers' nil checks were dead code and the call after them panicked. Return an untyped nil instead, and stop dereferencing without checking in SingleThreadForCPUs and NodeDistance.


askervin (approved):
LGTM.

Yet "already"s in below comments in unit tests were sort of helpful for reviewing this particular PR, they'll be less so after merging this code. So as a super-nit, I'd be fine dropping them.

// Node.DistanceFrom _already_ returns for an unknown destination.
...
// an absent destination was _already_ handled, and still is

PR #779, opened 2026-09-10
balloons: don't discover the system topology twice.

klihub:
NewCpuTreeFromSystem discovered a sysfs.System of its own, presumably to ask for CPU and cache topology explicitly. But DiscoverSystem ignores its flags and discovers everything anyway, so that System held the same data as the one the policy already has in its options. Take it as an argument instead.


askervin (approved):
LGTM

Thanks @klihub, very nice finding & fix!

PR #783, opened 2026-09-14
gitignore: ignore extracted testdata directories.

klihub:
Ignore unpacked test data (typically sysfs dumps), extracted into `testdata` directories. Ignore test binaries, too.


askervin (approved):
LGTM

PR #784, opened 2026-09-15
helm, e2e: fix gating of the plugin test APIs.

klihub:
plugin-test has a hyphen in it, which is not a valid character in a template field name. The workaround with quoting turned the condition into a string constant, always true. ENABLE_TEST_APIS=1 was set in every deployment, regardless of the value, even in production. Rename the value to plugin.test.enableAPIs. Update the e2e tests to use the new name.


askervin (approved):
LGTM

askervin:
Perhaps the "plugin" section in helm values will be a good home for "enable code coverage endpoint" helm option, too...

askervin:
Oh, correcting myself, it already _is_ the home for it, as enable test APIs includes coverage metrics...

PR #786, opened 2026-09-15
e2e: collect and report coverage data for e2e tests.

klihub:
Make test/e2e/run_tests.sh collect go coverage data of the plugins the tests
exercise, and report per-plugin and total coverage at the end of a run.

- `make e2e-tests` builds the resource-manager-based plugins with `go build
  -cover` (`make COVER=1 build images`).
- With the test APIs enabled a plugin serves its coverage data over the
  instrumentation HTTP server, so a plugin which never exits gracefully is
  covered too. What is served and what a plugin dumps at exit merge as they are.
- test/e2e/report-coverage.sh merges what a run collected into a profile, an
  HTML report and summary.json. `reset_coverage=1` discards earlier data first.

Verified on a full run of both suites: 55/55 PASS, balloons 78.6%,
topology-aware 68.0%, 59% of all instrumented packages.

The first commit is the standalone vm-wait-pod-regexp fix, needed for port
forwarding to the plugin; drop it if that lands first.


test/e2e/lib/vm.bash
```diff
@@ -1177,6 +1177,153 @@ vm-stop-log-collection() {
...
+
+    vm-coverage-reset-requested || return 0
+
+    case "$dir" in
+        ""|/)
+            echo "WARNING: refusing to discard coverage data under \"$dir\""
+            return 0
+            ;;
+    esac
+    [ -d "$dir" ] || return 0
```
askervin, line 1248: Can we always expect $dir/$vm_coverage_report_dir to exist, too?

I may be paranoid, but deleting files from host makes me wish to use all the checks possible to make sure that we're operating only on a known directory structure. So I was thinking if here [ -d "$dir/$vm_coverage_report_dir" ] || return 0 would work, or would it be too strict?

test/e2e/README.md
```diff
@@ -161,6 +161,93 @@ Worth knowing:
...
+- Data collected earlier is added to, not replaced, which is what makes a total
+  over several partial runs possible. A run which reruns a test replaces the
+  data of that test, but the data of the tests it does not run stays. Start the
+  run from scratch with `reset_coverage`, and the report covers this run alone:
+
+  ```shell
+  reset_coverage=1 ./run_tests.sh policies.test-suite
+  ```
+
+  `1`, `true` and `yes` all enable it. It discards everything collected earlier
```
askervin: On/off variables in the test framework (`provision`, `vagrant_debug`, `use_host_images`, `expect_error`,...) are usually 1=enabled, empty or 0=disabled. There's one nasty exception, though, with `skip_long_tests` being yes/no, and it is in the same level (run_tests.sh) as this one. So there is a tempatation to support yes/no here, too. But still...

I'd prefer limiting the range of accepted values in on/off variables rather than starting to allow new values to some of them. So maybe we could support only reset_coverage=1 or empty/0, and perhaps later switch the only exception (skip_long_lists) to support 1/0, too.

test/e2e/run.sh
```diff
@@ -538,6 +541,10 @@ helm-launch() { # script API
 
     vm-start-log-collection -n kube-system ds/$ds_name -c $ctr_name
     vm-port-forward-enable
+
+    # Drop whatever the plugin covered while starting up, so that the coverage
+    # data we collect is about what the test does.
+    vm-coverage-clear
```
askervin: As tests quite often that call helm-launch with a test-specific configuration, it would be reasonable to count a plugin applying its initial configuration as something that is covered by the test. So I'm wondering if it would be ok to drop vm-coverage-clear from here?

And on the other hand, many tests do helm-terminate and helm-launch many times with different configurations. Does the current SIGTERM behavior ensure that coverage is collectable. In some earlier experiments SIGTERM was racy and a custom handler failed to do the finishing it was supposed to. Do you think that helm-terminate could be extended with coverage collection to ensure that what was covered from first plugin launches would not be lost?

test/e2e/run_tests.sh
```diff
@@ -187,6 +192,11 @@ trap cleanup TERM EXIT QUIT
 summary_file="$summary_dir/summary.txt"
 echo -n "" > "$summary_file"
 
+# Start the run from scratch if asked to. This is done once here, for all the
+# tests, and never per test, as each test only ever overwrites the data of its
+# own earlier run.
+vm-coverage-discard-collected "${2:-$(pwd)}"
```
askervin: Could we give a name to $2, like we have
```
TESTS_DIR="$1"
```

test/e2e/README.md
klihub: I am consistently inconsistent and I think that is the primary reason why I usually aim for a liberal `yes/1/true` and `no/0/false` on FP2M and M2FP interfaces. But I understand your point and am fine with being strict instead and allowing only 1|0. I'll update the PR accordingly. We can file separate corrective PRs for the existing divergence(s).

test/e2e/run.sh
klihub: Yes, we could extend it. We have 2 methods of collecting coverage data: 1) let the runtime golang coverage infra dump collected data, and 2) retrieve the same data from an dedicated HTTP server endpoint running inside the plugin, started only if test APIs are enabled in the plugin. We could use the latter one here to always dump coverage data when we helm-terminate, so we wouldn't need to rely on the golang runtime getting around and successfully dumping everything. The only thing this requires is a working port redirection, which I think we anyway always have and already need in some cases.

test/e2e/lib/vm.bash
klihub: Narrowed it instead, matching both the filename and the directory holding it. In the discard itself and in `report-coverage.sh`'s find. So stray `covmeta.*` elsewhere under `${2:-$(pwd)}` (normally `test/e2e`) is now left alone.
                                                                                
Your paranoid extra test wouldn't have helped. That check passes on any repeat run, because `coverage-report` is there by then. So it would have deleted just as much. It only stops the first run, where there's nothing to delete anyway. And it would have made an interrupted run silent, as data left with no report would have caused `reset_coverage=1` to quietly do nothing in those cases.

test/e2e/run.sh
klihub: Dropped the clear from helm-launch too, as you suggested. helm install starts a fresh plugin, so its counters are zero anyway, and cross-test isolation comes from wiping the VM's coverage directory before
each test.

And we now dump coverage data when we helm-terminate a plugin.

test/e2e/run_tests.sh
klihub: Good point. Named it: `OUTPUT_DIR_ARG="${2:-}"` and `OUTPUT_ROOT="${OUTPUT_DIR_ARG:-$(pwd)}"` next to `TESTS_DIR="$1"`. So no bare `$2` is left. Naming it also made it visible that `$2` meant two things, which now should be fixed. It was also missing from the usage message, and that is now fixed, too.

klihub:
@askervin I pushed the branch with fixes/updates addressing the review comments for convenience, but it is now from a stacked PR set and not the bottom one (so it has a few extra bits below it). So let's move the review for these bits over to #792.

klihub:
Closing in favor of #792.

PR #787, opened 2026-09-15
e2e: resolve a single pod when waiting for one.

klihub:
In vm-wait-pod-regexp for a pod, make sure we only print running pods, not terminating ones. Otherwise our port forwarding command becomes racy when we terminate one pod and start a new one. This occasionally showed up in test logs as

    error: TYPE/NAME and list of ports are required for port-forward

This commit should fix that race.


klihub:

PR #788, opened 2026-09-15
DRA driver skeleton

bart0sh:
## Add a DRA driver skeleton to the resource manager

This PR adds DRA driver - registration, `ResourceSlice` publishing and the two claim entry points — and wires it into the resource manager behind a `dra.enabled` configuration option that defaults to off. No policy publishes devices yet, so with this PR alone an enabled driver registers and publishes an empty `ResourceSlice`. Policies get to say what they provide in the PRs that follow.

### Why the driver belongs to the resource manager

The driver is node-scoped infrastructure: one plugin registration, one socket, one `ResourceSlice` publisher per node, all of which outlive any particular allocation decision. Device *semantics* — what a device is, which ones exist,
what allocating one means — belong to the active policy. So `pkg/resmgr/dra` knows how to be a DRA driver and nothing about CPUs, memory or cache; it never inspects a device's attributes. The package comment states that boundary, and the tests assert it.

### What is here

- `pkg/resmgr/dra` — the plugin: `New`, `Start`, an idempotent `Stop`, and   `PrepareResourceClaims`/`UnprepareResourceClaims`. Built on `k8s.io/dynamic-resource-allocation/kubeletplugin`.
- `dra.enabled` in the common resource manager configuration, exposed by all three policy CRDs. "Common" here means the resource manager reads the switch from one place because it owns the driver — each policy CRD still carries its own `spec.dra.enabled`, so DRA is enabled per policy (and per node group, via the usual config CR layering), not cluster-wide. 
- resource manager wiring: `setupDRA`/`startDRA` from `start()`, `Stop()` from shutdown.
- Helm: the `resource.k8s.io/resourceslices` RBAC and the kubelet plugin and plugin-registry host mounts, in all three charts, gated on `config.dra.enabled` so a disabled deployment renders exactly as before.

### Details worth a reviewer's attention

**The driver is named after the active policy** — `topology-aware.nri.io`. The published devices are the policy's, so a node running a different policy publishes devices of a different kind, and a `DeviceClass` selecting on `device.driver` stays specific without further qualification. The cost is that switching a node's policy orphans the old driver's `ResourceSlices` until the node object goes away.

**The plugin takes the resource manager's lock directly.** `dra.Options` has a `sync.Locker`, and the resource manager passes itself: it embeds `sync.RWMutex` already. Every kubelet request takes that lock, which serializes claim handling against NRI request handling.

**`Stop()` stops the plugin before taking the lock.** `Helper.Stop()` drains the gRPC handlers in flight, and a handler arriving mid-drain blocks on the lock, so stopping the plugin under the lock deadlocks. There is a test for the ordering,
not just a comment: it starts a real plugin against a fake clientset, holds the lock to stand in for a request being served, and waits for the plugin's socket to disappear.

**Turning DRA on or off at runtime is refused**, with a message asking for a restart. Both directions have to happen with the lock released, and reconfiguration holds it.

**Missing cluster access degrades rather than fails.** A plugin configured from a local file has no kubernetes client and no node name; enabling DRA there logs a warning and leaves DRA off instead of refusing to start.

### Testing

`go test ./...` passes. Unit tests cover the plugin (registration, double start, idempotent stop, both claim entry points, and the boundary property that device attributes are never read), the configuration option across all three policy types, and the resource manager wiring including the shutdown ordering. `helm template` and `helm lint` verified on all three charts with DRA both enabled and disabled.

An end-to-end test arrives with the policy-side implementation, and will skip cleanly on clusters without the required feature gates.


pkg/apis/config/v1alpha1/dra.go
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+// DRAConfig provides configuration data for the DRA (Dynamic Resource
+// Allocation) driver. The driver is owned by the resource manager, not by a
+// policy, so this is common configuration for all policies.
+type DRAConfig struct {
```
klihub, line 20: nit: I wonder if this should go under `pkg/apis/config/v1alpha1/resmgr/dra` (mirroring where the driver skeleton itself lives which IIUC this is the configuration for), for consistency with our other stuff. I think this might have been modelled after `AgentConfig`, which I suspect is one of the existing inconsistencies. Not entirely sure though ATM, and not insisting on this at all. Just something that caught my eye so I thought I should mention it.

klihub: Okay, sorry. I think I was clearly wrong. That's the DRA skeleton, but this configures stuff in `pkg/resmgr/dra.go`. This is in the right place then I think.

klihub (commented):
@bart0sh I did a first skim through and I haven't spotted anything off. I'll do one more with more fresh eyes tomorrow.

pkg/resmgr/resource-manager.go
```diff
@@ -223,6 +233,10 @@ func (m *resmgr) start(cfg cfgapi.ResmgrConfig) error {
 func (m *resmgr) Stop() {
```
bart0sh, line 235: @klihub This function is not called anywhere as far as I can see. It's not related to this PR, but we may want to either remove it if it's not needed or call it when resource manager is stopped.

klihub: @bart0sh Looking at how fairly plain it is, let's just get rid of it, if it is unused.

askervin: @bart0sh, I ran into that, too, when looking for a spot for clearing published extended resources.

Discussion about such clean-ups finally produced this open issue:

https://github.com/containers/nri-plugins/issues/709

but based on @marquiz's advice, I finally gave up whole idea of trying to clear published resources on highly unreliable plugin exit handlers. Instead of following plugin pod/container life cycle, extended resource (un)publishing follows reconciliation model: if the desired state expressed in most recent configuration is "don't publish", and if the plugin notices resources being published anyway, then it clears them.

bart0sh: @askervin Having some cleanup capabilities is not mandatory, but it's not a bad thing either. Can we simply catch SIGTERM or/and other signals and call resmgr.Stop in the signal handler? Did you try this approach?
Anyway, this is not part of DRA, so I'll leave this code here. If we decide to remove it, it would be better to do it in another PR.

askervin: My first implementation of extended resources included only the "best effort" clean-up, triggered by SIGTERM. I don't think I would have wired it up through resmgr.Stop(), though.

One nasty thing is that a policy may have quite a lot of quite slow things to clean up, so it's very realistic to assume that the policy gets killed before finishing everything (frequencies, governors, SST configurations, IRQ affinities, CPU affinities of hundreds of containers, and of course published resources).

A partly or unnecessarily done clean-up can be more disruptive than no clean-up. Markus argued that, for instance in case of nri-plugins update, we should avoid deleting and then re-creating node resources. For these reasons, I gave up trying to clean up resources at all on exit, and only reconcile published resources (and all other above mentioned system configurations) to match effective configuration.

Of course what a policy should or should not do, is another thing that what the resmgr should let any policy know. In this perspective, connecting resmgr.Stop with SIGTERM makes sense, and definitely in its own PR.

bart0sh: > Stop with SIGTERM makes sense, and definitely in its own PR.

Submitted https://github.com/containers/nri-plugins/pull/799, PTAL.

bart0sh:
@klihub Rebased on top of fresh main. Updated HandleError to properly process unrecoverable driver errors. PTAL.

klihub (commented):
@bart0sh Still looking good to me. Let me read this once more tomorrow morning, and I'd like to give @askervin also the opportunity to do so, if he has some spare cycles. We can then chat about it if we spot anything in our sync call.

klihub (approved):
LGTM.

pkg/apis/config/v1alpha1/dra.go
```diff
@@ -0,0 +1,26 @@
...
+	// Enabled registers a DRA driver for this node and publishes the devices
+	// the active policy provides. A policy which provides no devices leaves a
+	// registered driver with no devices, which claims cannot be made against.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
askervin: @bart0sh, @klihub, what do you think, would it enable being more user-friendly in the next DRA steps, if DRAConfig.Enabled would be *bool instead of bool?

I'm considering if we could avoid a small piece of user-facing configuration complexity (as dependencies between configuration options), and pay the price in the complexity of policies. That is, if a user has given a policy configuration that clearly states "I want to publish _____ as a DRA resource", then the policy could enable DRA, unless DRAConfig.Enabled has been explicitly set to false. With *bool, `DRAConfig.Enabled == nil` could be interpreted as "false unless needed by some other option". I don't know yet about the topology-aware policy, but I'm absolutely sure that at least the balloons policy will have per-balloonType DRA parameters...

On one hand, this may sound and be a little thing in the configuration, but on the other hand, I see that they are piling up. For instance, if a balloons user configures "preferCloseToDevices: podresourceapi: vendor.com/gpu", it is obvious that the user wants her container near a GPU. However, balloons requires that user configures the main switch for kubelet pod resources API  (`agent.podResourceAPI: true`), too, otherwise her preferCloseToDevices will be just wishful thinking.

That said, I'm fine merging this PR as is, or with simple bool -> *bool change with no other changes, because building a way for a policy to either modify common configuration, or express wishes/new defaults on it, is far from the purpose of this and follow-up PRs. But I'm hoping we'd be able to keep user configuration complexity minimal, as (especially with balloons) it's already pretty complex.

askervin (approved):
LGTM

pkg/apis/config/v1alpha1/dra.go
bart0sh: @askervin Good point, thanks! `Enabled` is now a `*bool` (see the latest commit in this PR).

klihub: > @bart0sh, @klihub, what do you think, would it enable being more user-friendly in the next DRA steps, if DRAConfig.Enabled would be *bool instead of bool?

Yes. I think unless there is a pressing reason, it is better to idiomatically always go for a pointer instead of a direct value/scalar in these cases. That allows us to change default behavior later, without using kubebuilder tags.

PR #789, opened 2026-09-15
e2e: boot the node back to a NUMA kernel whatever happens.

klihub:
The two test30-numa-disabled tests boot the node with a kernel which has NUMA support compiled out, and boot it back at the end by calling enable-numa as their last statement. A failure anywhere in between never gets there: error exits the test, so the node keeps running the kernel without NUMA. If we are running multiple tests, every test after the failure sees a machine with a single node... and will fail.

Restore the kernel from an EXIT trap instead, the way the tests which have state to put back already do it, and let enable-numa return without doing anything if NUMA is enabled, so that calling it when there is nothing to restore is free.

Looking into a failure of these tests is easier on a node which still has NUMA disabled, so take keep_numa_disabled=1 to leave it that way:

    keep_numa_disabled=1 ./run_tests.sh \
        policies.test-suite/balloons/n4c16/test30-numa-disabled


klihub:

PR #790, opened 2026-09-15
e2e-runner: improve test result collection and reporting.

klihub:
Note: This PR is manually stacked on top of (IOW includes) #786. Please review that one first.

Improve our e2e nightly runner and reports.

- Publish a report per run: the failures first, each with the reason it failed
  and links to the logs, the commands and the test source at the tested
  revision, then the coverage of each plugin, then every test case. Plus an
  index of all runs, and a status.txt and a latest symlink for the cronjob.
- Keep results for RETENTION_DAYS (100 by default) and pack up the artifacts of
  the tests which passed: a full run publishes 17M instead of 93M.
- New options: --runtime, --full-build|--minimal-build, --retention-days,
  --retention-keep, --keep-coverage-data, --keep-artifacts.
- Fixes: the results of a topology not matching n[0-9]*-c* were left out of the
  summary, so a run whose only failures were there was published as passing;
  the first run on a host died of a git internal error.
- A caddy configuration for serving the results.

Exit status semantics are unchanged: non-zero only if the runner itself failed,
never because the tested system failed.


klihub:

PR #791, opened 2026-09-16
e2e: fix pod resolution, NUMA kernel restore and swap detection

klihub:
Three fixes to the test harness which the rest of this stack builds on.

- vm-wait-pod-regexp resolved more than one pod, and could resolve one which was
  terminating. Port forwarding to the plugin then got two names and kubectl
  refused with "TYPE/NAME and list of ports are required for port-forward".
  Print a single pod, and never a terminating one.
- The test30-numa-disabled tests boot the node with a kernel which has NUMA
  compiled out and boot it back by calling enable-numa as their last statement,
  so a failure anywhere in between left the node without NUMA and every later
  test saw a single node and failed too. Restore the kernel from an EXIT trap
  instead, the way the other tests with state to put back already do, and let
  enable-numa return without doing anything when NUMA is already enabled.
  keep_numa_disabled=1 leaves the node NUMA-less for looking into a failure.
- Provisioning decided whether to switch swap off from the top-level
  ansible_swaptotal_mb fact, which is not there with injected top-level facts
  turned off. Take it from ansible_facts instead.


askervin (approved):
LGTM

askervin (commented):
We have some other tests, too, that change system configuration in such a way that it may lead failures in next tests (modify kernel command line, for instance).

Therefore I wonder if we should actually have some more generic options to control:
1. whether or not we want a test to try recovering the node after failed test execution, and/or
2. whether or not we want execution of multiple tests stop on first error

In that perspective, some sort of global debug_first_fail=1 could perhaps imply keep_numa_disabled=1 and possibly other "don't restore on failure" variables.

But this doesn't need to be part of this PR. I'm just thinking how we could manage keep_numa_disabled-like variables...

klihub:
> We have some other tests, too, that change system configuration in such a way that it may lead failures in next tests (modify kernel command line, for instance).
> 
> Therefore I wonder if we should actually have some more generic options to control:
> 
> 1. whether or not we want a test to try recovering the node after failed test execution, and/or
> 2. whether or not we want execution of multiple tests stop on first error
> 
> In that perspective, some sort of global debug_first_fail=1 could perhaps imply keep_numa_disabled=1 and possibly other "don't restore on failure" variables.
> 
> But this doesn't need to be part of this PR. I'm just thinking how we could manage keep_numa_disabled-like variables...

@askervin That's a good point. Let's take a closer look at that and then extend/clean up the limited concept this PR introduced to be more generally usable, but let's do it in a separate dedicated PR.

PR #792, opened 2026-09-16
e2e: collect coverage data from e2e test runs

klihub:
Make test/e2e/run_tests.sh collect go coverage data of the plugins the tests
exercise, and report per-plugin and total coverage at the end of a run.

- `make e2e-tests` builds the resource-manager-based plugins with `go build
  -cover` (`make COVER=1 build images`).
- With the test APIs enabled a plugin serves its coverage data over the
  instrumentation HTTP server, so a plugin which never exits gracefully is
  covered too. What is served and what a plugin dumps at exit merge as they are.
- test/e2e/report-coverage.sh merges what a run collected into a profile, an
  HTML report and summary.json. `reset_coverage=1` discards earlier data first.

Verified on a full run of both suites: 55/55 PASS, balloons 78.6%,
topology-aware 68.0%, 59% of all instrumented packages.

Port forwarding to the plugin needs the vm-wait-pod-regexp fix from the PR below.


askervin (approved):
LGTM

PR #794, opened 2026-09-16
e2e: generate the e2e reports in go

klihub:
Generate the e2e reports in go.

Replace the 750 lines of python of scripts/testing/e2e-report and the awk of
report-coverage.sh with test/e2e/cmd/e2e-report, golang and standard library only.

- Subcommands `run`, `index` and `coverage`; html/template instead of building
  HTML by hand, and one implementation of the coverage arithmetic.
- Run with `go run ./test/e2e/cmd/e2e-report`, so it always matches the
  revision under test; nothing is added to any image.
- Byte-identical output: results.json, index.html and status.txt of the runs we
  have, and of the index over them, are the same files the python wrote. Only
  the spelling of escaped quotes in HTML text differs.
- The coverage totals match `go tool cover -func` and the per-plugin numbers
  match the awk it replaces.


askervin (approved):
LGTM

PR #795, opened 2026-09-16
e2e: publish the results of a run packed, and serve them

klihub:
Publish the results of a run packed, and serve them.

A run of the whole suite collects some 120M, most of it the command transcripts
and the plugin logs of each test, and most of that the same lines over and over.
In one solid archive it takes 2.3M, so nothing has to be thrown away to keep the
results of months of nightly runs.

- `e2e-report pack` writes what a run collected to `results.tar.zst` and removes
  what it packed. The report, `results.json`, `status.txt` and `summary.txt` stay
  where they are, so how a run went is readable without unpacking anything.
- `e2e-report serve` serves a result root over HTTP, the packed runs as if their
  archives had been extracted where they are: a report links to the same paths
  whether its run is packed or not, and directory browsing works either way.
- `e2e-runner --pack-results` publishes a run this way, and then keeps the
  artifacts of every test and the coverage data of each as well, since with the
  results packed there is nothing to gain by pruning them.
- A tarball is served both ways: downloaded without a trailing slash, browsed
  into with one, so the artifacts a test packed up can be read a file at a time
  instead of being downloaded. That works for the archive of a run, for the
  artifacts of a test, and for the latter inside the former. The report links to
  the artifacts of a test both ways, the tarball first, as a plain file server
  can only ever hand it over.

Packing is off by default: a packed run needs the server to browse. `tar --zstd
-xf` gets the results out without one.

`e2e-runner` and the caddy configuration move to `scripts/testing/nightly`, which
also gets a README of how to set the nightly up from cron and how to serve what
it publishes, and a systemd unit for the server.

Measured on a published nightly run: 17M as published today (pruned, with the
artifacts of each test in a tar.xz) packs to 2.8M, and the same run with nothing
pruned at all, 117M and 4693 files, packs to 2.3M. Reading a file from an
archive takes 20-300ms, decompressing it up to that file; only the listing of
each archive is kept.

Standard library plus `klauspost/compress` for zstd, which was already in the
module graph. Reading a `.tar.xz` shells out to `xz`, rather than taking a
dependency for the sake of the tarballs of tests published so far; without it
such a tarball is still served, it just cannot be served into.

Tests cover packing, listing, serving packed and unpacked runs, browsing into
`.tar.xz`, `.tar.gz` and `.tar` both on their own and nested in a run archive,
content types, and that neither a path leading out of the result root nor a
doctored archive gets anything else served.


askervin (approved):
LGTM

PR #799, opened 2026-09-17
Stop the resource manager on shutdown

bart0sh:
## The problem

`resmgr.Stop()` has never been called in production. `Main.Run()` ends with

```go
err := m.mgr.Start()
return err
```

and `Start()` blocks in the config agent's event loop, so nothing follows it. On
SIGTERM we stop the agent, the loop returns, `Run()` returns — and the process
disappears from under its own subsystems. Nobody calls `resmgr.Stop()`, so the
NRI stub is never stopped and the runtime is left to notice a closed socket on
its own. `git log -S "mgr.Stop()"` confirms the call was never there.

## The fix

This PR calls it, on the way out of `Run()`:

```go
defer m.mgr.Stop()

return m.mgr.Start()
```

Two things had to be fixed for that call to be safe:

- `nriPlugin.stop()` dereferenced `p.stub`, which only `start()` ever assigns.
  `Stop()` is now reachable on paths where `start()` never ran, so it returns
  early on a nil stub.
- `Agent.Stop()` waited on `doneC`, a channel which is declared but never
  created or closed, so it never returned. Nothing noticed, since its only
  caller is a goroutine and the event loop was already on its way out. The wait
  is gone; closing `stopC` is what makes `Start()` return. `stopC` is no longer
  cleared either, so the two goroutines selecting on it don't race with
  `Stop()`.

Also trimmed a comment in the signal handler which claimed we clean up
published extended resources on the way out. We don't, and per the earlier
discussion we shouldn't — that state belongs to reconciliation, not to an exit
handler.

## What the shutdown path does — and should keep doing

This makes `Stop()` reachable, which makes it an inviting place to hang
resource cleanup off. It deliberately does none, and shouldn't grow any. In
full, a signalled plugin now does:

- `Agent.Stop()`: one status PATCH clearing this node's entry in the config
  resource's `.status.nodes` (pre-existing, 10s timeout), then `close(stopC)`,
  which makes the config watches and the extended-resource reconciliation loop
  return.
- `resmgr.Stop()`: `nri.stop()`, i.e. `stub.Stop()`, closing the NRI
  connection.

Nothing else. No policy teardown — there is no `policy.Stop()` to call — and
nothing that walks containers. Frequencies, governors, SST configuration, IRQ
and CPU affinities are left as they are; published extended resources stay
published; NRT custom resources are not deleted. The work is O(1) and does not
scale with container count, so there is no realistic way for the kubelet's
SIGKILL to catch us half-way through, and no partly-done state that could be
more disruptive than no cleanup at all. Reconcile-to-effective-configuration on
start remains the only mechanism that converges system state, and the cache the
next instance reconciles from is untouched — which is exactly the conclusion of
the earlier review discussion about clean-up on exit.

## Out of scope

No node-state cleanup on exit (extended resources, policy hooks) — see above
for why. No context plumbing either. A lifetime `context` would be a reasonable
cleanup on its own — today's contexts are all leaves, `context.Background()`
handed to `stub.Start` and to the config watches — but here it would add a
mechanism without replacing one: cancelling a context does not stop the NRI
stub, the controllers or the HTTP server, which have `Stop()` methods instead.

## Testing

New e2e test,
`policies.test-suite/topology-aware/n4c16/test41-graceful-shutdown`. It signals
the plugin with SIGTERM and asserts, from the log of the terminated instance,
that handling the signal took it through stopping the resource manager and
closing the NRI connection, and that the process exited with 0. It signals the
process rather than deleting the pod, because that restarts the container in
place and keeps the previous instance's log readable; reading it needs the new
`--previous` option in `plugin-log`.

The test was verified to detect the regression it guards, against a live
cluster: it passes with these commits, fails with `error: the resource manager
was not stopped on shutdown` when the fix is reverted and the image rebuilt,
and passes again once it is restored. The exit-code assertion alone would not
have caught it — unfixed, the plugin still exits 0, it just skips stopping its
subsystems.

`test00-basic-placement`, `test09-container-exit` and `test12-config-status`
also pass, the last one covering both a plugin terminated mid-run and one which
fails on its initial configuration. Unit tests, `go vet` and `gofmt` are clean.

Log on SIGTERM, where previously it stopped after the agent:

```
received signal terminated, shutting down gracefully
[resource-manager] shutting down...
[nri-plugin] stopping plugin...
msg="Stopping plugin 90-resource-manager..."
[http] stopping HTTP server...
```


klihub (dismissed):
@bart0sh Thank you ! This looks really good, with even an e2e thrown in. Plus now at least I learned about the `logs --previous` flag. I did not know it exists.

bart0sh:
@klihub 

> Thank you ! This looks really good, with even an e2e thrown in. Plus now at least I learned about the `logs --previous` flag. I did not know it exists.

Great! If you merge it, I'll rebase https://github.com/containers/nri-plugins/pull/788 on top of it and make sure DRA driver's shutdown is wired into the resmgr shutdown.

bart0sh:
> I'm happy with it as such. But it's still a draft.

I created it as a draft to see that all CI pieces are green. Now they're and you can merge it.

klihub (approved):

PR #803, opened 2026-09-21
DRA policy boundary

bart0sh:
## Let the policy decide what the DRA driver publishes

Second PR of the DRA series, on top of #788. It adds a three-method DRA
interface to `policy.Backend`/`policy.Policy`, gives all three backends
identical no-op implementations of it, and makes the driver publish the devices
the active policy describes instead of a fixed empty slice.

On a cluster this is still inert: every backend answers that it publishes no
devices, so an enabled driver registers and publishes an empty `ResourceSlice`,
exactly as after #788. What changes is who decides that.

### The interface

```go
DRADevices() ([]resourceapi.Device, error)
AllocateClaim(
    claim *resourceapi.ResourceClaim,
    results []resourceapi.DeviceRequestAllocationResult,
) ([]specs.ContainerEdits, error)
ReleaseClaim(uid types.UID) error
```

That is the whole extension point for DRA, and it is deliberately narrow:

**Upstream types only.** No `Claim`, `ClaimAllocation` or `CDIEdit` type of our
own. `resourceapi.ResourceClaim`, `resourceapi.DeviceRequestAllocationResult`
and `specs.ContainerEdits` already carry everything the boundary needs,
including `ShareID` and `ConsumedCapacity` for devices with consumable
capacity, so the interface adds no vocabulary a reader has to learn. The only
new dependency is `tags.cncf.io/container-device-interface/specs-go`.

**`ReleaseClaim` takes a UID and nothing else**, because a UID is all the
unprepare path has: `kubeletplugin` hands `UnprepareResourceClaims` only
namespace, name and UID, with no allocation results. So the policy has to keep
whatever it needs to undo an allocation, keyed by claim UID — which it wants
anyway, since it also has to restore that state across a restart, the same way
it restores its container allocations.

**The returned edits correspond to `results` positionally**: `edits[i]` belongs
to `results[i]`. That is what lets the driver assemble a CDI spec from them
without knowing what any single edit means.

**`DRADevices()` is also where a policy validates its DRA configuration.** The
driver asks before registering: a policy that cannot say what it offers must
not become a driver kubelet sends claims to.

**`nil, nil` from `DRADevices()` means "this policy publishes nothing"** — it
does not switch the driver off. `dra.enabled` alone decides whether the driver
runs, and a policy with no devices still gets an empty slice published. That
keeps "this policy publishes nothing" and "this policy has nothing right now"
the same observable state, which is what a reader of `kubectl get
resourceslices` should see, and it means a policy that stops publishing does
not leave a stale slice behind.

### Adding a device type later touches no driver code

Say topology-aware wants to publish isolated cores alongside high-priority
ones. It emits more devices from `DRADevices()`, with whatever attributes it
wants them selected by, and teaches `AllocateClaim` what to do with a request
for them. `pkg/resmgr/dra` is untouched: it never reads a device attribute, and
there is a test asserting it passes attribute-less devices through unchanged. A
`DeviceClass` selects on `device.driver` plus whatever attributes the policy
chose. Balloons gains DRA support the same way, by replacing three stubs in its
own package.

### Why two of the three methods merge without a caller

The claim dispatch path that calls `AllocateClaim`/`ReleaseClaim` is the next
PR. Declaring only `DRADevices()` here would mean touching `policy.go` and all
three backends twice for one interface, and would split the two methods
carrying the subtle parts of the contract — idempotency across repeated
prepares, the positional edits, being the last line of defense against handing
the same resources to two claims — away from the interface they belong to. The
interface is what is under review here, so it is reviewed as one piece.

### Deliberately left for later

- **claim dispatch**: prepare/unprepare routed to `AllocateClaim`/
  `ReleaseClaim`, plus the CDI spec store that turns the returned edits into
  something kubelet can inject. Next PR.
- **republishing**: nothing in this PR can change what `DRADevices()` returns,
  so slices are published once at start. Keeping them fresh is a separate
  mechanism, and it arrives with allocation, together with the trigger that
  gives it a cause.
- **real implementations**: the template policy first, as a worked example for
  policy authors and as a second implementation of this interface written
  before the first one is finished, then topology-aware.
- **end-to-end coverage**: arrives with the first policy that actually
  allocates.

### Worth a reviewer's attention

- **The three signatures and the contract comments on them.** Everything that
  follows is built on them, and this is the cheapest point at which to move
  them.
- **`pkg/resmgr/policy` gained only `resourceapi`, `specs` and `types`
  imports** — no dependency on `pkg/resmgr/dra`. The dependency runs the other
  way: the driver holds a `policy.Policy` and asks it.
- **The driver publishes devices exactly as the policy described them**, in one
  pool named after the node, and it fails `Start` if `DRADevices()` errors —
  before registering, so a misconfigured policy never becomes a driver kubelet
  talks to.
- **All three stubs are identical**: `nil, nil` from `DRADevices()`, a shared
  `policy.ErrNoDRAClaims` from the other two. Not "not implemented yet"
  placeholders — each of these policies genuinely does not support DRA claims,
  and `errors.Is` works on that answer.

### Testing

`go test ./...` passes. New tests cover the forwarders on the `policy` struct
(success and error propagation for each of the three), publishing the active
policy's devices, an empty slice when the policy publishes nothing, a
`DRADevices()` error failing `Start` before registration, and the boundary
property that the driver passes devices through without inspecting their
attributes.


klihub (commented):
I skimmed through this once and it looks good to me. I'll give it another read tomorrow.

klihub:
@bart0sh There is a duplicated (test) function flagged by linting...

klihub (approved):
LGTM.

pkg/resmgr/policy/policy.go
```diff
@@ -132,6 +135,43 @@ type Backend interface {
 	// Returning nil or an empty map means the policy owns and
 	// publishes nothing.
 	GetExtendedResources() map[string]*resource.Quantity
+	// DRADevices returns the DRA devices this policy wants published, or nil
+	// if it publishes none. It also validates the policy's DRA-related
+	// configuration, so that a misconfigured policy is caught before any of
+	// its devices can be claimed.
+	DRADevices() ([]resourceapi.Device, error)
+	// AllocateClaim allocates resources for a claim being prepared. results
```
askervin: s/results/Results

pkg/resmgr/dra/plugin.go
```diff
@@ -230,16 +250,40 @@ func (p *Plugin) HandleError(_ context.Context, err error, msg string) {
...
 }
 
-// driverResources returns the resources to publish for this driver. Publishing
-// an empty slice, instead of no slice at all, states that the driver is alive
-// and has no devices, which is what this driver has until a policy provides
-// some.
-func (p *Plugin) driverResources() resourceslice.DriverResources {
+// driverResources returns the resources to publish for this driver: the devices
+// of the active policy, in a single pool named after our node. The devices are
+// passed on as the policy described them, attributes and all — what they mean
```
askervin: nit: let's stick with pure ASCII.

askervin (commented):
Just a nit on a non-ASCII character, otherwise LGTM.

Looking forward to currently out-of-scope resource updates from policies...

pkg/resmgr/dra/plugin.go
bart0sh: @askervin Thank you for the review! Fixed.

PR #809, opened 2026-09-22
resmgr: retain pending updates on nri failure

bart0sh:
Pending container updates that go out through `UpdateContainers()` are currently taken from their containers *before* the push: `updateContainers()` starts with `getPendingUpdates()`, which clears each update as it returns it. If the push then fails, the update is gone - and it was the only record of what the container should have looked like, so the container keeps its old resources and there is nothing left to retry it with.

The updates are now taken from their containers only once the runtime has them. `updateContainers()` peeks at the pending updates, pushes them, and clears the ones that were delivered; a failed request leaves all of them pending.

The runtime can also accept the request and report back the updates it could not apply. That answer used to be discarded. Those containers were not updated either, so their updates now stay pending just the same, and a warning names them.

The updates carried by an NRI response are unchanged: they are still taken as they are collected, because nothing  acknowledges them - of the NRI messages only `UpdateContainersResponse` has a `failed` field — so there is no later point to clear them at, and keeping them would repeat them in every response from then on.

`cache.Container` accordingly trades its destructive `GetPendingUpdate()` for `PeekPendingUpdate()` and  ClearPendingUpdate()`: reading a pending update and discarding it are now separate steps, which is what the retry above needs.

This path is how any out-of-band allocation change reaches the runtime, so the existing user is policy reconfiguration; DRA claim allocation (#806) is what made the lost updates easy to hit.

Tests: `pkg/resmgr/nri_test.go` covers the three outcomes — retained when the request fails, cleared when the runtime has them, retained for the containers the runtime rejected. The first and third fail against the previous behaviour.


klihub (approved):
I think this should be fine. I have a few nits about comment wording, which sound a bit like LLMisms.

pkg/resmgr/nri.go
```diff
@@ -696,25 +696,76 @@ func (p *nriPlugin) RemoveContainer(ctx context.Context, pod *api.PodSandbox, co
 	return nil
 }
 
+// updateContainers pushes the pending container updates to the runtime. Unlike
+// the updates carried by the response to an NRI request, these are only taken
```
klihub: nit: @bart0sh Wouldn't it be more humanly understandable to say 'these are only cleared from the containers' instead of 'taken from the containers' ?

pkg/resmgr/nri.go
```diff
@@ -696,25 +696,76 @@ func (p *nriPlugin) RemoveContainer(ctx context.Context, pod *api.PodSandbox, co
...
+		if u := c.PeekPendingUpdate(); u != nil {
+			p.setDefaultClasses(c, u)
+			updates = append(updates, u)
+		}
+	}
+
+	return updates
+}
+
+// clearPendingUpdates takes the delivered updates from their containers. The
```
klihub: nit @bart0sh ditto here. Wouldn't it be more humanly understandable to say 'clears the delivered updates from their containers' ?

pkg/resmgr/nri.go
```diff
@@ -729,24 +780,14 @@ func (p *nriPlugin) getPendingAdjustment(container *api.Container) *api.Containe
 	return nil
 }
 
+// getPendingUpdates collects the updates to carry in the response to an NRI
+// request, and takes them from their containers as it does so. Nothing tells us
```
klihub: nit: @bart0sh  ditto here. Wouldn't it be more humanly understandable to say 'and clears them from their containers' ?

bart0sh:
@klihub Thank you for the review. All review comments have been resolved.

